PR #35, opened 2014-12-15
Sidesync

Redsandro:
For controlling Samsung tablets and smartphones on your desktop.


ferventcoder:
Git fetch && git rebase

gep13:
Am I not right in saying that creating a local branch would be better?

My routine is normally the following...

```
git fetch upstream
git merge upstream/master
git push
git checkout -b AddingSideSync
git add .
git commit -m "message here"
```

Then issue a pull request once pushed.

Working in the local branch gives you complete isolation.

Redsandro:
I'm not sure why not to merge. When merging master to local branch, head and master on that branch are the same. Merging a new change will be at the top when you push that back.

Anyway, this thread is deprecated. SideSync is merged.
I just need @gep13 or someone with access to add it to the ketarin machine.

https://github.com/chocolatey/chocolatey-coreteampackages/issues/38

gep13:
So, I am confused...

The only Pull Request that was created to bring in SideSync was closed. How did the SideSync files get into the master branch?

gep13:
> I'm not sure why not to merge. When merging master to local branch, head and master on that branch are the same. Merging a new change will be at the top when you push that back.

This is dangerous if the merge cannot be fast-forwarded no?  Better to work in isolation on a separate branch, rebasing on master as and when required, after fetching.  It also makes it easier to back out the changes as a result of a pull request if it is issued from a branch, rather than directly from remote master.

This is the approach that I would definitely recommend going forward, especially for new package additions.  For make quick break fix changes, sure commit directly to master, but I think we have to mandate that new package addition are done through a Pull Request.

ferventcoder:
> This is dangerous if the merge cannot be fast-forwarded no? Better to work in isolation on a separate branch, rebasing on master as and when required, after fetching. It also makes it easier to back out the changes as a result of a pull request if it is issued from a branch, rather than directly from remote master.

This is exactly why the `git fetch upstream` then `git rebase upstream/master` - should always do a fast forward, but in the case where it doesn't you are covered.

ferventcoder:
if it doesn't you have to git reset and more fun stuff, easily avoidable. I run into this because I push to the upstream sometimes after a change has been made. I get the "whoops, something has changed upstream. Please go get that and pull it down and fix up your stuff before your next push" - if I did a git pull (git fetch and merge), it would apply the merge commit in there and things would look really nasty. 

Easily avoidable with always using `git fetch remote` / `git rebase remote`.

gep13:
> This is exactly why the git fetch upstream then git rebase upstream/master - should always do a fast forward, but in the case where it doesn't you are covered.

But I am right in saying that working directly on master is just bad practice, right?

Redsandro:
> How did the SideSync files get into the master branch?

I merged the new SideSync branch into master and removed it.

Redsandro:
About all the git-foo, and apart from @ferventcoder's valid tips, I'm not sure I understand the reservations. But I do think most reservations are based on older git experiences, because I often find my more seasoned teammates (not this team, I mean at my work) providing friendly advice on how to use git in a more complicated way, and funnily, they usually disagree among themselves on how exactly.

We've been using git (at work) for the last year or so and I always checkout and fetch the target branch, and apply my own branch onto it. That's all, I never had a problem, and I've been ignoring all the advise because they all contradict each other.

They are all clever people and they don't make stuff up, so I am thinking that their "way" originated from a time where the universe would collapse more easily if you used git to manage it. I think these days, git is a bit more clever and knows what you want.

But I am not sure.

gep13:
> Please keep it as simple as possible. If there is a team preference (this team), write it down and add some smileys.

Yes, I think we need to do this, and documenting the core team packages in general, i.e. what it is, what it is and isnt' for, is on my radar to complete.

PR #43, opened 2015-01-08
Updated Pale Moon for use with FossHub

Redsandro:
This does require manually changing the __ketarin__ url's passed to `chocopkgup` both for `url` and `url64`

`url` `http://www.fosshub.com/genLink/Pale-Moon/palemoon-{version}.win32.installer.exe`
`url64` `http://www.fosshub.com/genLink/Pale-Moon/palemoon-{version}.win64.installer.exe`

I cannot adequately edit ketarin.xml because for some reason it does not contain the 32 bit url?

### DIFF
--- a/automatic/palemoon/tools/chocolateyInstall.ps1
+++ b/automatic/palemoon/tools/chocolateyInstall.ps1
@@ -4,5 +4,16 @@ $url = '{{DownloadUrl}}'
 $url64 = '{{DownloadUrlx64}}'
 $silentArgs = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
 $validExitCodes = @(0)
+$pwd = "$(split-path -parent $MyInvocation.MyCommand.Definition)"
+
+
+
+# Combatibility - This function has not been merged
+if (!(Get-Command Get-UrlFromFosshub -errorAction SilentlyContinue)) {
+	Import-Module "$($pwd)\Get-UrlFromFosshub.ps1"
+}
+
+$url = Get-UrlFromFosshub $url
+$url64 = Get-UrlFromFosshub $url64
 
 Install-ChocolateyPackage "$packageName" "$installerType" "$silentArgs" "$url" "$url64" -validExitCodes $validExitCodes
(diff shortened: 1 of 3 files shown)


automatic/palemoon/tools/chocolateyInstall.ps1
```diff
@@ -4,5 +4,16 @@ $url = '{{DownloadUrl}}'
 $url64 = '{{DownloadUrlx64}}'
 $silentArgs = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
 $validExitCodes = @(0)
+$pwd = "$(split-path -parent $MyInvocation.MyCommand.Definition)"
+
```
gep13: What is the reason for multiple line returns here?

gep13:
Looks good to me, aside from one comment/question.

automatic/palemoon/tools/chocolateyInstall.ps1
Redsandro: Short: None || readability.

Long: Preference. I always do that in my own work. 3 lines between declaring boring stuff and the actual work. It's one of the things why my code is easier to read than others. But of course that's a matter of preference, and an automatism. Just my handwriting I guess.

gep13: Fair enough.  My particular preference goes the other way :-)  I am used to StyleCop telling me in my C# code if I have multiple lines between code segments.  This is not a major thing here though, just wanted to make sure I wasn't missing anything.

gep13:
@Redsandro I have just loaded the new ketarin file etc onto the Choco Server.  Does this need to be forced onto chocolatey.org with a break fix notation, or you happy to wait?

automatic/palemoon/tools/chocolateyInstall.ps1
Redsandro: StyleCop :lollipop:

Redsandro:
No it's okay. I have made a mistake too. I thought there was an update, and that the relmirror was broken. Turns out there was no mirror of this release because 25.1.1 is android only. The 'old' package still works and is vCurrent for Windows.

Anyway, fosshub works fine too so waiting for an actual update is fine. :)

gep13:
coolio :-)

Issue #32, opened 2014-12-04
Bring in tonigellida's packages


ferventcoder:
https://github.com/adgellida/chocolateyautomaticpackages

Redsandro:
@ferventcoder we can start by circumventing having to click and add a maintainer 203 times, and accepting to be maintainer 203 times, by entering the chocolatey maintainer directly into the database.

I don't know what database we use and how the maintainers are stored, but here is some pseudo-code that we can use:

```
$push: { 'maintainer' : $each: 'chocolatey'}
[...]

PR #44, opened 2015-01-09
Added Tribler AutoIt autopackage

Redsandro:


gep13:
@Redsandro can you fix up this pull request as per your others?  This one seems to have all the same problems, i.e. duplicate executeCommand, etc.

Redsandro:
Please update [wiki instructions](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#automatic-packages)that doesn't actually say anything about that. Most things are trivial. Perhaps we can come up with a conversion script that only takes out the relevant settings? I think only the variables are relevant, and the rest is static and can only be wrong.

gep13:
I can certainly take a task to update the wiki instructions to show how the chocolatey ketarin instance that we are running is set up, so that others who want to contribute to this repo can match it.  As for creating a conversion script, I don't really see the value in that.

> Or change the chocolatey system to use the %temp% folder in stead of something non-standard?

We have over 200 packages set up this way, so I would much rather have new packages match the way that we are already set up.

Redsandro:
> I don't really see the value in that.

I don't know how many people have autopackages set up on their own machine, but they are all gonna have the same problem, no?

I thought it might come in handy also if/when we address this issue: https://github.com/chocolatey/chocolatey-coreteampackages/issues/32

gep13:
> I don't know how many people have autopackages set up on their own machine, but they are all gonna have the same problem, no?

Not if we document how we have set up the ketarin instance that we are using.  With that in place, there shouldn't be that many differences, no?  Although, now I think about it, a small script to load the xml, and regex through the known offenders, wouldn't take that much to achieve.

Redsandro:
> @gep13 said:
Although, now I think about it, a small script to load the xml, and regex through the known offenders, wouldn't take that much to achieve.

That's what I'm talking about :)

Except we might as well use `xsl` (Extensible Stylesheet Language) templates, it kinda does exactly what we need.

> @Redsandro commented in https://github.com/chocolatey/chocolatey-coreteampackages/pull/46#issuecomment-69514985
>
> For the `xml` part, if the relevant parts and unsettable or static parts are documented, we can automate this using `xslt`. A bit of a puzzle but pretty powerful. [Here's how it split the database into separate packages](https://github.com/chocolatey/chocolatey-coreteampackages/commit/a4899f37d6872244ed18280ae44f446f1acbec32).

Redsandro:
(Removed the waiting label since I updated the branch)

gep13:
> Except we might as well use xsl (Extensible Stylesheet Language) templates, it kinda does exactly what we need.

Been a while since I had the pleasure of creating any XSL, so would be very rusty at this.  Are you an XSL guru?  Happy to look at something you can come up with.

Redsandro:
I had to use it often in the past. Little rusty too but it's perfect for the job. I'll have a shot once we know what needs to happen.

Redsandro:
If all is fine, you can merge this and be done with it.

> Happy to look at something you can come up with.

Continue here: https://github.com/chocolatey/chocolatey-coreteampackages/pull/47

PR #49, opened 2015-01-24
cyg-get: removed mirror url

Ferk:
It's not required to provide a mirror url for cywinsetup.exe, specially if cygwin was already installed and the settings from the installation are kept (which is a requirement for cyg-get already, anyway).

Simply removing it would work.

Alternativelly, you could get a mirror at random (or based on location if possible) from: https://cygwin.com/mirrors.lst

But please don't pick the US California mirror like this making it the forced default. Right now it's quite slow from where I'm in.


ferventcoder:
Can you rebase this?

Ferk:
Rebased

ferventcoder:
Thanks!

gep13:
@ferventcoder did you want to push a new version of this package?

ferventcoder:
Yes please.

PR #55, opened 2015-02-04
fixed gimp; install+uninstall were not silent

dtgm:
silent args were not stringent enough to install or uninstall silently.  updated to suppress progress boxes.

also, filled in empty nuspec tags and corrected poor info

### DIFF
--- a/automatic/gimp/gimp.nuspec
+++ b/automatic/gimp/gimp.nuspec
@@ -4,11 +4,12 @@
     <id>{{PackageName}}</id>
     <version>{{PackageVersion}}</version>
     <title>GIMP</title>
-    <authors>The GIMP Team</authors>
-    <licenseUrl>http://www.gnu.org/licenses/gpl.html</licenseUrl>
+    <authors>GIMP Team</authors>
+    <licenseUrl>https://git.gnome.org/browse/gimp/plain/LICENSE</licenseUrl>
     <projectUrl>http://gimp.org</projectUrl>
     <iconUrl>http://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/50fd97744110dcbce1acde889c0870599c9d5584/icons/gimp.svg</iconUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
+		<summary>GNU Image Manipulation Program</summary>
     <description>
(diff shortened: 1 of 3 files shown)


automatic/gimp/gimp.nuspec
```diff
@@ -16,9 +17,8 @@ GIMP has many capabilities. It can be used as a simple paint program, an expert
 
 GIMP is expandable and extensible. It is designed to be augmented with plug-ins and extensions to do just about anything. The advanced scripting interface allows everything from the simplest task to the most complex image manipulation procedures to be easily scripted.
     </description>
-    <summary>GIMP</summary>
-    <releaseNotes></releaseNotes>
-    <copyright>GPL</copyright>
+    <releaseNotes>http://www.gimp.org/release-notes/</releaseNotes>
```
gep13: @dtgm The release notes section is capable of doing markdown.  Could you make this URL a markdown URL so that it is clickable?

gep13:
@dtgm thanks for your contribution!  Apart from one comment, this looks good to me.  Can you fix it up, and then I will get it pulled in.

Thanks!

automatic/gimp/gimp.nuspec
dtgm: @gep13 Hi!  Thanks for the comment!  It is not necessary to break this out into a newline for a clicky link.  Any fully valid URI is automatically converted into a clickable link by markdownsharp, regardless of it's position.  I try to just  leave a static link for releaseNotes tag because there isn't a good way for now to automatically update the content in this section with chocopkgup.  Of course some people ::cough:: @ferventcoder ::cough:: think it better to leave the most epic releaseNotes ever: https://github.com/chocolatey/choco/blob/master/nuget/chocolatey/chocolatey.nuspec#L40-L142

gep13: >  Any fully valid URI is automatically converted into a clickable link by markdownsharp

I did not know this!  Good to know!

>  try to just leave a static link for releaseNotes tag because there isn't a good way for now to automatically update the content in this section with chocopkgup

Yes, I do the same thing, but I write them as a fully declared markdown URL, i.e with [] and ()'s.  Will know that this is not required going forward.

gep13:
@dtgm These changes will appear in the next official release of GIMP, or do you think it is justified to release a version with package fix notation?

PR #60, opened 2015-03-06
added dependency: vcredist2013

matbech:
the .msi no longer contains the vcredist 2013 merge module. It is now a prerequisite.


TomOne:
Thanks

PR #62, opened 2015-03-07
cyg-get support for multiple packages to be installed in a single cygwinsetup session

Ferk:
The help text in cyg-get seemed to imply that support for installing multiple packages was not available.

This is however not true, it has been supported in cygwin setup.exe from the start, however the powershell script was parsing the arguments and removing the commas that separate the packages.

It works correctly with this patch.

It would be nice if  "choco install package1 package2 -source cygwin" was able to make use of this instead of launching a separate cygwinsetup every time... which is very slow and requires redownloading the package indexes every time.

### DIFF
--- a/packages/cyg-get/tools/cyg-get.ps1
+++ b/packages/cyg-get/tools/cyg-get.ps1
@@ -6,16 +6,17 @@ if ($package -eq "" -or $package -eq $null) {
   Write-Warning 'Please specify a package or list of packages. Run -help or /? for more information.'
 }
 elseif ($package -eq '-help' -or $package -eq '/?') {
-  Write-Host "To run please specify `'cyg-get packageName`'. At the current time it doesn't appear that passing a list of packages works, but it would be done like this: `'cyg-get package1,package2,packageN`'. Note the commas and no spaces."
+  Write-Host "To run please specify `'cyg-get packageName`'. You can also specify a list of packages like this: `'cyg-get package1,package2,packageN`'. Note the commas and no spaces."
 }
 else {
   try {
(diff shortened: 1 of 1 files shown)


packages/cyg-get/tools/cyg-get.ps1
```diff
@@ -5,17 +5,18 @@ param(
...
+elseif ($package -eq '-help' -or $package -eq '--help' -or $package -eq '-h' -or $package -eq '/?') {
+  Write-Host "To run please specify `'cyg-get packageName`'. You can also specify a list of packages like this: `'cyg-get package1,package2,packageN`'. Note the commas and no spaces."
 }
 else {
   try {
     $cygRoot = (Get-ItemProperty HKLM:\SOFTWARE\Cygwin\setup -Name rootdir).rootdir
     $cygwinsetup = Get-Command $cygRoot"\cygwinsetup.exe"
-    $cygPackages = join-path $cygRoot packages
+    $cygLocalPackagesDir = join-path $cygRoot packages
+    $cygInstallPackageList = $package  -replace(" ",",")
```
ferventcoder: How does this react to spaces at the end?

Ferk: The reason why spaces exist is the way powershell parses the argument with the params being used in line 1, it seems to be transforming the comma separated list into a string of elements separated by space. It doesn't look like it adds spaces to the end, even if you call it as cyg-get "a,b,c " ... however, if you add a comma at the end it will complain: 

    $ powershell cyg-get.ps1 a,b,c,
    Missing expression after ',' in pipeline element.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : MissingExpression

This is however not caused by the -replace, but by the params parsing, it would be happening before too

PR #66, opened 2015-04-03
Added keepass-classic. New version is out. Please run this.

Redsandro:
1.29 is out. I cannot run vbs here.

Remember this is from sourceforge, so the `url` variable is just a mirror. The command for the package I used is:

`chocopkgup /p {packageId} --pf {packageRoot} /v {version} /u "{downloadUrl}"`


TomOne:
Thanks @Redsandro 

This needed to be fixed before the merge:

* For files hosted on SF, the mirror-specific URL should not be used. Always use the general URL, which will redirect to the best mirror for the user.
* Some information in the package description was incorrect. KeePass Professional (= KeePass 2.x) is the cross-platform app written in .NET (Mono), not KeePass Classic. KeePass Classic is Windows-only, but its file format is compatible with KeePassX.
* Use the package ID as the Ketarin job’s application name, see https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#automatic-packages.

Please keep these things in mind for future packages.

TomOne:
The package is released: https://chocolatey.org/packages/keepass-classic/1.29

Redsandro:
:+1:

PR #76, opened 2015-05-14
New package: RenameMaster (autopackage untested)

Redsandro:
Rename Master 3.12 is out. I tried to push a package, but my VirtualMachine setup is gone. I tried for 2 hours to get a new setup, mapping directories, to my original ketarin db and stuff, but I cannot get it to work and I am quite simply out of time.

I just used git from linux to push this to a new branch, and I am hoping that someone can check this out.


Redsandro:
**Offtopic:** Actually 3 hours went by. We could really use a vagrant box that does all this stuff, including setting up a share to /tmp, and cloning the repo there, so git is already using the setup from the host.

On the other hand, how many of you are making this complicated like me by not using Windows?

PR #81, opened 2015-06-29
New package: Resource Hacker

Redsandro:
There have been many many many updates, but the package never broke because the installer file remains the same. Now I needed a fix because it changed, but it would be nice if this could be run with the autopackages every now and then so the version number doesn't stay out of date anymore.

So I pushed one more update, and created this autopackage.


Redsandro:
Can anyone verify this and put it in the mamebapma?

PR #83, opened 2015-07-18
New Package: AOMEI partition manager Standard and Light

Redsandro:
These autopackages work fine except that the download target needs some regex processing. Hence Ketarin will fail because it doesn't have the direct download link.

Someone willing to fix this?


Redsandro:
This package needs sources. I cannot find a download location for these files that doesn't work with hashes and stuff.

Redsandro:
Original author hosts files for crawling sites to update, but they are 1990 slow.
For now, there is no other choice.

PR #92, opened 2015-11-24
updated smartftp icon

matbech:


gep13:
@matbech is there a reason for making the SmartFtp Icon smaller?

matbech:
Nope this wasn't my intention. I have incorrectly assumed that the size will be set in the img element on the package page. But this is not the case. I have changed the default size to 128x128 now (same as current size).

gep13:
Perfect, thank you! :+1:

matbech:
Thank you.

PR #93, opened 2015-11-25
SmartFTP: updated iconUrl

matbech:


gep13:
I think the URL that you are after is this one:

https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/97d1e867fd5feaee1b89b24ffe6d40c22c819dc3/icons/smartftp.svg

matbech:
Wouldn't it be better to use the url pointing to the master branch so that we don't have to update the iconUrl again when the icon is updated?

gep13:
Unfortunately not, no.  There is an issue with package icons getting cached against the original URL (part of the way the CDN works), so when you change the icon, it won't get served up if you use the same URL.  You can find the details here:

https://github.com/chocolatey/choco/wiki/CreatePackages#package-icon-guidelines

matbech:
Ah I understand. Thanks. I have updated the IconUrl accordingly.

gep13:
Perfect! :+1:

PR #97, opened 2015-11-26
SmartFTP: Fix 32-bit msi url

matbech:
url for the 32-bit msi file was pointing to the wrong (64-bit) msi.


gep13:
Looks good to me!  Good work! :+1:

PR #102, opened 2015-12-23
Fix firefox install on 32 bit only systems

jroweboy:
The firefox package would successfully install on the system, but it would not report that it installed successfully because it would call `Get-ChildItem ` on the 64 bit uninstall directory which doesn't exist on 32 bit. This will avoid that check when on a 32 bit platform and will also only try to install the 32 bit version on a 32 bit platform.


TomOne:
Good catch, thanks. The fixed package is released: https://chocolatey.org/packages/firefox/43.0.2.20151214

PR #114, opened 2016-01-16
vcredist2015: removed unnecessary try/catch

matbech:


gep13:
@matbech when I first looked at this, I thought the try/catch was still warranted, but having thought about it, I think you are right.

Thanks!

PR #120, opened 2016-01-19
qBittorrent fixes

Link-Satonaka:
Sorry for messing up the diffs, couldn't figure out another way to capitalize the B

- Added details to the nuspec
- Reordered some of the items in nuspec to match updated template
- Followed wiki advice for automatic SF packages, don't use {{DownloadUrl}} (also removes the {{DownloadUrlx64}} workaround)
- Adjusted ketarin variables now that we aren't using the workaround
- Added checksum to install script
- Updated uninstall script to newest template, consolidated into one script


Link-Satonaka:
hmm looks like the website isn't showing the capital B regardless :/

TomOne:
Thanks for this PR. Keep in mind that package IDs should be lowercase, see https://github.com/chocolatey/choco/wiki/CreatePackages#naming-your-package. Can you update the PR according to that rule? Thanks

Link-Satonaka:
Done.

Out of curiosity, why would one opt to use {{PackageName}} over hardcoding the id to the nuspec? It's not supposed to change, so making it a substituted variable seems unwise.

gep13:
@Link-Satonaka thanks for your contribution!

PR #121, opened 2016-01-20
Add Flash Player PPAPI plugin for Opera and Chromium browsers

leonghui:
- uses a different URL and source for $version, may work for the NPAPI and ActiveX packages
- Adobe did not release MSI installer for the PPAPI version (see https://forums.adobe.com/message/8108499#8108499), but the EXE installer supports silent setup. 
- reuses existing Flash icon
- tested installation and uninstallation on Win7 x64 with Chromium 49.0.2598.0 x64

My first PR to this project, please review

### DIFF
--- a/automatic/flashplayerppapi/flashplayerppapi.ketarin.xml
+++ b/automatic/flashplayerppapi/flashplayerppapi.ketarin.xml
@@ -1,0 +1,140 @@
+<?xml version="1.0" encoding="UTF-8"?>
+  <Jobs>
+   <ApplicationJob xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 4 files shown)


gep13:
@leonghui thank you for your submission! On the whole, this looks good, but I do have one question...

In the chocolateyInstall.ps1 file, the download URL is hard-coded.  In order for this to be considered an automatic package, the download URL, and anything else that is specific to the package version, should be replaced via the ketarin script.

If you can make that up to make it the case, we can get this package pulled in.

gep13:
@leonghui apologies, didn't mean to close teh Pull Request.

leonghui:
@gep13 
Thanks for the review, I have replaced the hardcoded values with the tokens.

gep13:
@leonghui that looks better.  Let me try out the package just now.  Thanks!

PR #125, opened 2016-02-01
(cyg-get) Fix the path of cygwinsetup.exe

pallxk:
`cygwinsetup.exe` is sure to be found in `${Env:ChocolateyInstall}\lib\Cygwin\tools\cygwin`, as cyg-get depends on Cygwin.

However, it's not guaranteed to be found in the installation rootdir of Cygwin under the following circumstances:
1. Cygwin is installed manually to another path, and then installed by chocolatey;
2. When rootdir of Cygwin can be customized when installing via Chocolatey in the future.

This PR is intended to fix installation described in the following image:

cygwinnotfound

After patch:

afterpatch

### DIFF
--- a/packages/cyg-get/tools/cyg-get.ps1
+++ b/packages/cyg-get/tools/cyg-get.ps1
@@ -57,8 +57,9 @@ if ($help -or $packageNames -join '|' -eq '/?') {
       }
     }
 
-    $cygwinsetup = "$cygRoot\cygwinsetup.exe"
-    $cygLocalPackagesDir = join-path $cygRoot packages
+    $cygwinsetupDir = "${Env:ChocolateyInstall}\lib\Cygwin\tools\cygwin"
+    $cygwinsetup = "$cygwinsetupDir\cygwinsetup.exe"
+    $cygLocalPackagesDir = join-path $cygwinsetupDir packages
     $cygInstallPackageList = $packageNames -join ','
 
     $cygArgs = "--root $cygRoot --local-package-dir $cygLocalPackagesDir"


fatso83:
I can verify that this fixes #176 . I changed the contents of `C:\ProgramData\chocolatey\lib\cyg-get\tools\cyg-get.ps1` using the patched file and afterwards it works fine!

pyhedgehog:
It still passes `--root $cygRoot` that will be translated to something like `--root C:\tools\cygwin64 C:\tools\cygwin`. Root should be selected depending on `-arch` option with default to auto-detect installed (if only) or native (64).

fatso83:
@pallxk Are you able to integrate what @pyhedgehog said into your fix? Really waiting for this ...

pallxk:
@fatso83, what @pyhedgehog said can be fixed by testing `$arch` first when setting `$cygRoot`, and I'm willing to help if some one is interested to merge `cyg-get` related PR, but I'm not sure...

What's more, currently there is #200 which blocks the functionality of `cyg-get` since it seems that the new `cygwin` install script no longer saves the cygwin setup file to Chocolatey lib directory `%ProgramData%\Chocolatey\lib\Cygwin\tools\cygwin\cygwinsetup.exe`, but leaves it in a temp directory like `%TEMP%\chocolatey\Cygwin\2.5.1\CygwinInstall.exe`. Though this only affects new cygwin installation...

majkinetor:
This is fixed. Will push it tomorrow.

majkinetor:
2f66dea

PR #172, opened 2016-02-24
Get cygwin into Bin-Root and a few other corrections

Link-Satonaka:
While looking through the script I noticed a few logical errors that needed correcting. If github isn't showing you my comments inline with the code on the PR, it does so here: https://github.com/Link-Satonaka/chocolatey-coreteampackages/commit/13b5a3a4ba54b125c027d0ae7a24bce16b773a88


ferventcoder:
@Link-Satonaka it's an easier conversation if you comment on the files section of the PR instead of individual commits.

PR #173, opened 2016-02-24
Added chocolateyUninstall.ps1 file for Firefox

sparkvilla:
I made the chocolateyUninstall.ps1 for Firefox choco package, please review. I tested the package with the new file using my local folder/repository and the uninstalling pocedure is now silent. I also tested through Ansible (that needs Firefox to be silently uninstalled otherwise it does not work) and it works too.

### DIFF
--- a/automatic/Firefox/tools/chocolateyUninstall.ps1
+++ b/automatic/Firefox/tools/chocolateyUninstall.ps1
@@ -1,0 +1,9 @@
+$unfile = "Mozilla Firefox\uninstall\helper.exe"
+
+if (Test-Path "${Env:ProgramFiles(x86)}\$unfile"){
+   $unpath = "${Env:ProgramFiles(x86)}\$unfile"
+}
+else {
+   $unpath = "${Env:ProgramFiles}\$unfile"
+}
+Uninstall-ChocolateyPackage 'Mozilla Firefox' 'exe' '/S' "$unpath"


automatic/Firefox/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,9 @@
+$unfile = "Mozilla Firefox\uninstall\helper.exe"
+
+if (Test-Path "${Env:ProgramFiles(x86)}\$unfile"){
+   $unpath = "${Env:ProgramFiles(x86)}\$unfile"
+}
+else {
+   $unpath = "${Env:ProgramFiles}\$unfile"
+}
```
ferventcoder: Our preference for packages is to use the registry to get the uninstall location.

sparkvilla:
In the registry I could find only the path to the firefox.exe file. How would I get to the helper.txt path through the registry?

ferventcoder:
Weird, I am going to call no dice on this one.

image

ferventcoder:
The only reason Chocolatey can't automatically uninstall it is that it is not able to detect the installer type (yet).

ferventcoder:
So this is interesting...

* FirefoxInstall.exe: installer: 7-Zip(1.0); overlay: 7-zip Installer data(-)

Unzipped there is a setup file:

* /FirefoxInstall/setup.exe: installer: Nullsoft Scriptable Install System

sparkvilla:
Yep, I found the location of the Uninstall string in the registry. Sorry but I just started with powershell, so.. how would I change my script to get to uninstall location from the registry?

ferventcoder:
@sparkvilla first `choco upgrade chocolatey -pre` and then run `choco new firefox` and take a look at the uninstall script that is generated to adjust this one.

sparkvilla:
@ferventcoder I have used the template you suggested to make up a second version of chocolateyuninstall.ps1 file. It works but Firefox is not uninstalled silently (things that was achieved with the script of this pull request). Something wrong. How can I upload the new script so that you can have a look at it? or, Shell I open a new pull request?

ferventcoder:
The script generated for uninstall is a template result which still requires some editing. The idea is to look at that to get an idea how you can adjust this PR to use the registry instead. You still need to make the adjustments to it for the silent arguments.

ferventcoder:
You can always paste script text into these messages as another alternative way of looking at what you have questions about.

sparkvilla:
` # If this is an exe, change installerType and silentArgs

$ErrorActionPreference = 'Stop'; # stop on all errors

$packageName = 'firefox'
$softwareName = 'Mozilla Firefox' #part or all of the Display Name as you see it in Programs and Features. It should be enough to be unique
 $installerType = 'EXE' 

#$silentArgs = '/qn /norestart'
# https://msdn.microsoft.com/en-us/library/aa376931(v=vs.85).aspx
$validExitCodes = @(0, 3010, 1605, 1614, 1641)
if ($installerType -ne 'MSI') {
  # The below is somewhat naive and built for EXE installers
  # Uncomment matching EXE type (sorted by most to least common)
  $silentArgs = '/S'           # NSIS
  #$silentArgs = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-' # Inno Setup
  #$silentArgs = '/s'           # InstallShield
  #$silentArgs = '/s /v"/qn"' # InstallShield with MSI
  #$silentArgs = '/s'           # Wise InstallMaster
  #$silentArgs = '-s'           # Squirrel
  #$silentArgs = '-q'           # Install4j
  #$silentArgs = '-s -u'        # Ghost
  # Note that some installers, in addition to the silentArgs above, may also need assistance of AHK to achieve silence.
  $silentArgs = 'none'             # none; make silent with input macro script like AutoHotKey (AHK)
                                #       https://chocolatey.org/packages/autohotkey.portable
  $validExitCodes = @(0)
}

$uninstalled = $false
$local_key     = 'HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*'
$machine_key   = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\*'
$machine_key6432 = 'HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\*'

$key = Get-ItemProperty -Path @($machine_key6432,$machine_key, $local_key) `
                        -ErrorAction SilentlyContinue `
         | ? { $_.DisplayName -like "$softwareName" }

if ($key.Count -eq 1) {
  $key | % { 
    $file = "$($_.UninstallString)"

    if ($installerType -eq 'MSI') {
      # The Product Code GUID is all that should be passed for MSI, and very 
      # FIRST, because it comes directly after /x, which is already set in the 
      # Uninstall-ChocolateyPackage msiargs (facepalm).
      $silentArgs = "$($_.PSChildName) $silentArgs"

      # Don't pass anything for file, it is ignored for msi (facepalm number 2) 
      # Alternatively if you need to pass a path to an msi, determine that and 
      # use it instead of the above in silentArgs, still very first
      $file = ''
    }

    Uninstall-ChocolateyPackage -PackageName $packageName `
                                -FileType $installerType `
                                -SilentArgs "$silentArgs" `
                                -ValidExitCodes $validExitCodes `
                                -File "$file"
  }
} elseif ($key.Count -eq 0) {
  Write-Warning "$packageName has already been uninstalled by other means."
} elseif ($key.Count -gt 1) {
  Write-Warning "$key.Count matches found!"
  Write-Warning "To prevent accidental data loss, no programs will be uninstalled."
  Write-Warning "Please alert package maintainer the following keys were matched:"
  $key | % {Write-Warning "- $_.DisplayName"}
} `

gep13:
@sparkvilla @ferventcoder I just wanted to touch base on this...

Is there still work required here?

gep13:
@sparkvilla pinging again about this...

Is there still work required?

RedBaron2:
@gep13 @ferventcoder  @sparkvilla  Taking a look at this now will be back with a PR soon as work permits.

RedBaron2:
I have found a solution for @sparkvilla problem; This borrows some code from the original install file. Created a new PR #262

gep13:
Closing this in favour of new PR: https://github.com/chocolatey/chocolatey-coreteampackages/pull/262

PR #187, opened 2016-03-07
added scripts directory to Python2 install path

### LINKED ISSUE #185: Python needs another directory in $PATH
For Python packages,

At the end, where it sets the path to enable python to be executed anywhere, I would suggest adding the .\scripts subdirectory as well. The Scripts directory contains the pip command, which is almost as useful as the python executable.

$env:Path = "$($env:Path);$installPath;$installPath/Scripts"

vchan2002:
fixes #185


gep13:
Thanks for the contribution.  This fix should appear in the package when the next release of python is available.

PR #189, opened 2016-03-12
(wixtoolset) Updating to 3.10.2.2516

corux:


gep13:
@corux thank you very much for your help with this!

gep13:
@corux I have just pushed this package, and assuming it goes through moderation ok, it should be available soon.  Thanks again!

PR #190, opened 2016-03-18
Check full version

BuffaloTrace:
Script only checked for major version and ignored minor version.  Now it checks both.


gep13:
@BuffaloTrace  Thanks for the contribution! This should appear when the next release of the flashplayerplugin is available.

PR #219, opened 2016-05-27
Initial PSPP autopackage draft

Redsandro:
@gep13 I have received a polite request to update this package from 2014. I'm now on Crete working on another project, so I've attempted to create an autopackage after hours without actually testing this on Windows. They obfuscate their paths a bit, but I think I've got the regexes worked out.

Could you be so kind as to see if this builds? If it does, it would be awesome if you could merge this and add it to the ketarin machine.

### DIFF
--- a/automatic/pspp/pspp.ketarin.xml
+++ b/automatic/pspp/pspp.ketarin.xml
@@ -1,0 +1,173 @@
+﻿<?xml version='1.0' encoding='utf-8'?>
+<Jobs>
+  <ApplicationJob xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" Guid="48990104-8c8d-401f-a9c2-60653fc81c65">
+    <SourceTemplate><![CDATA[<Jobs>
+  <ApplicationJob xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema" Guid="0fb30714-8ed0-4611-8f1b-cb8fec9dae91">
+    <WebsiteUrl />
+    <UserAgent>Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/535.19 (KHTML, like Gecko) Chrome/18.0.1025.168 Safari/535.19</UserAgent>
+    <UserNotes />
+    <LastFileSize>384846</LastFileSize>
+    <LastFileDate>2012-05-23T02:09:37.7748325</LastFileDate>
+    <IgnoreFileInformation>false</IgnoreFileInformation>
+    <DownloadBeta>Default</DownloadBeta>
+    <DownloadDate xsi:nil="true" />
+    <CheckForUpdatesOnly>false</CheckForUpdatesOnly>
+    <VariableChangeIndicator />
+    <CanBeShared>true</CanBeShared>
+    <ShareApplication>false</ShareApplication>
+    <ExclusiveDownload>false</ExclusiveDownload>
+    <HttpReferer />
+    <SetupInstructions />
+    <Variables>
+      <item>
+        <key>
+          <string>version</string>
+        </key>
+        <value>
+          <UrlVariable>
+            <RegexRightToLeft>false</RegexRightToLeft>
+            <VariableType>StartEnd</VariableType>
+            <Regex />
+            <Url><placeholder name="Url with version information" value="http://sourceforge.net/projects/calibre/files/" /></Url>
+            <StartText>&lt;TABLE cellspacing ="1" cellpadding ="6" border = "0"&gt;
+  &lt;TR&gt;
+    &lt;TH class="Title" align="center" width=90&gt;7-Zip </StartText>
+            <EndText>&lt;BR&gt;</EndText>
+            <TextualContent />
+            <Name>version</Name>
+          </UrlVariable>
+        </value>
+      </item>
+      <item>
+        <key>
+          <string>url64</string>
+        </key>
+        <value>
+          <UrlVariable>
+            <RegexRightToLeft>false</RegexRightToLeft>
+            <VariableType>Textual</VariableType>
+            <Regex />
+            <TextualContent>""</TextualContent>
+            <Name>url64</Name>
+          </UrlVariable>
+        </value>
+      </item>
+    </Variables>
+    <ExecuteCommand />
+    <ExecutePreCommand />
+    <ExecuteCommandType>Batch</ExecuteCommandType>
+    <ExecutePreCommandType>Batch</ExecutePreCommandType>
+    <Category />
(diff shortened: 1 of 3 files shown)


gep13:
@Redsandro just tried to build this on the choco server, and I was met with this:

```
pspp	The downloaded file is not a binary file type (text/html; charset=utf-8). Possibly there is an error page. Status code: 200 (OK) (https://sourceforge.net/projects/pspp4windows/files/2016-04-01/pspp-0101%2B20160401-snapshot-32bits-setup.exe/download)
```

Are you in a position to take a look?

Thanks!

Redsandro:
@gep13 I'm not sure if this is related, but sourceforge is down for me:

```
This site can’t be reached

sourceforge.net’s server DNS address could not be found.
DNS_PROBE_FINISHED_NXDOMAIN
```

If waiting doesn't automatically fix this, at least it will fix sourceforge and I can look into it.

gep13:
@Redsandro Hmm, when I tried the URL in my browser just now, it did seem to work, but perhaps there are geographical problems.

gep13:
@Redsandro just tried again, and now I get a different error! :cry:

```
pspp	The remote server returned an error: (500) Internal Server Error. (https://sourceforge.net/projects/pspp4windows/files/2016-05-28/)
```

gep13:
@Redsandro and that URL doesn't work for me locally.  So perhaps the initial error was a general failure with sourceforge, but this one seems more package specific.

Redsandro:
@gep13 thank you for trying this. I still cannot reach sourceforge at all. a 500 could very well be sourceforge related too. That page should give 200 or 404, not 500.

Either way, could you post the `-debug` output? I'll try to recreate the problem with curl without having to use a VM and chocolatey once sourceforge is back up for me too. The debug log would probably answer all my questions trying to debug this.

Redsandro:
@gep13 The first [url that didn't work](https://sourceforge.net/projects/pspp4windows/files/2016-04-01/pspp-0101%2B20160401-snapshot-32bits-setup.exe/download) now works fine today, so the ketarin works for that part (`url32` I guess). This is the url passed to Chocolatey because it has the sourceforge installer built-in.

The second URL indeed does not exist. It is not even part of the first url. I wonder why it was requested. Bad server hiccup? If it's a script error, I don't see it just yet. Could you give me the debug log so I can see at what point this occurs?

Redsandro:
@gep13 Added different refs (and fixed date for previously unused vars) - try this.

gep13:
@Redsandro can you confirm what debug logs you are referring to?  I am getting the error message that I gave you when trying to do `Update` from within ketarin, once i load in the XML file.

gep13:
@Redsandro I get the same issue with the most recent commit that you have made :cry:

Redsandro:
@gep13 aww gawd.. I was referring to the extra verbose output when running chocolatey with `-debug`, but am I understanding correctly that you are getting errors from just the ketarin, without even getting to the chocolatey part?

(PS - you need to pull master, not just copy the code from the commit linked above - because the regex failed to commit to the message but did commit to the master branch.)

Redsandro:
[This right here](https://github.com/chocolatey/chocolatey-coreteampackages/blob/f02c04eb99514fa5e541b13a0853f9e1bf04f01b/automatic/pspp/pspp.ketarin.xml#L95) should make that date you're getting in an url impossible.

gep13:
@Redsandro yes, sorry, this is even before getting to Chocolatey, this is within ketarin.  And yes, I have pulled the latest, this is actually part of the ketarin batch file that we have, it also pulls the latest before doing it's run.

gep13:
@Redsandro this is what I am doing...

pspp

And this is the error message:

```
pspp	The downloaded file is not a binary file type (text/html; charset=utf-8). Possibly there is an error page. Status code: 200 (OK) (https://sourceforge.net/projects/pspp4windows/files/2016-04-01/pspp-0101%2B20160401-snapshot-32bits-setup.exe/download)
```

Redsandro:
The regex that works for _bulk crap uninstaller_ with the `mirror` variable just responds with the html page here.

Perhaps @dtgm can shed a light?

dtgm:
As noted: https://gitter.im/chocolatey/choco?at=57572b2517856dc5179b2fe5

Issue 1 is ketarin isn't getting the direct download URL because sf appears to serve the file requested based on some options set for the repo. Some repos will not redirect any EXEs (like in pspp repo) and others will properly redirect some files, and not others. It's an odd situation and I never looked in depth why this happens to some repos and not others. I suspect that the admin of the repo has options configured to trigger this, but not sure.

Issue 2 is ketarin won't run commands unless a file is downloaded. This can be any file from any available host. There are options in ketarin to not download the file, but then only ketarin's db gets run, not the commands.

Two possible solutions

1. I use another file from the repo that redirects me properly to valid mirror to use. (not valid in this case)
2. Use any arbitrary file, like a single pixel, from a dependable host. (the downside is if that host goes down or URL changes, pkg won't get updated) for example: 
  * `https://www.google-analytics.com/__utm.gif`
  * On advanced page, you would set to use `version` variable to check for updates.

I don't use `#2` because I rely on direct URL to calc SHA sums.

Redsandro:
@dtgm thank you for explaining. It's indeed exactly what was going on. The file from the mirror was offered in every example I inspected except in this case, so I was sure it was something I did wrong. But apparently it wasn't. :smile: 

@gep13 I've added solution 2 as described by @dtgm. Ketarin should work fine now. _But the new problem_ is that - if I recall correctly - the script this repo uses is copied from @ferventcoder's example using the ketarin url (which is now the google pixel) in stead of the `url` variable:

```
/u "{preupdate-url}" /u64 "{url64}"
```

I have the variables `url` and `url64` set up for x86 and x64 respectively.

It should be like this for this package only:

```
/u "{url}" /u64 "{url64}"
```

gep13:
@Redsandro sorry, I am not following your last statement....

Is there something that i need to do here?  If so, what?

Redsandro:
@gep13 well, the Ketarin update works with this configuration, I'm just not sure if the chocolatey automatic package builder script works with the settings used by the ketarin machine. So I just made a couple of assumptions:

1. The _"update url"_ setting in Ketarin now points to the url of a google pixel.
2. The [example script](https://github.com/chocolatey/choco/wiki/AutomaticPackages#setup) shows this command:  
`chocopkgup /p {appname} /v {version} /u "{preupdate-url}" /u64 "{url64}" /pp "{file}"`  
3.  
  * I don't know what `{preupdate-url}` points to
  * I'm _assuming_ it's the _"update url"_ setting in Ketarin, which is now wrong. Therefor:
  * I _think_ it should now point to the `{url}` in stead of `{preupdate-url}` at the `/u` switch
  * Since we use a single global script for all packages, you would need to override that script in the settings of this package specifically. Everything the same, except `url` in stead of `preupdate-url`.

PR #228, opened 2016-06-08
KeePassX autopackage

Redsandro:
@gep13 This is also an old static package turned autopackage because updates were requested.
This one should be straightforward because the download is direct (no mirrors) and the installer is just a .zip.


gep13:
When trying to load the ketarin xml file, I am getting this:

image

Can you correct?  I can't see anything obviously wrong, but perhaps I am looking in the wrong place.

gep13:
@Redsandro ^^

Redsandro:
@gep13 sorry, I can't see what I did wrong. So I took a different (working) xml and merged the `keepassx` settings. Can you verify if it imports now?

gep13:
@Redsandro yes, that seems to work :+1:

PR #245, opened 2016-07-08
Update chromium.ketarin.xml

RedBaron2:
Updated regex for getting revision numbers; Added start/end text for revision


RedBaron2:
@gep13 Could You take a look at this PR. I think it might help in the future. Thanks

Issue #256, opened 2016-07-12
eclipse


030:

majkinetor:
This ticket is useless so I'll close it.

gep13:
@majkinetor this issue is related to the PR that was created, and which is linked in this issue.  I can understand closing it once the PR is either merged or closed, but not before.

majkinetor:
There is no value in it, PR itself is a form of an issue.

majkinetor:
Furthermore, it will stay linked from PR, it just doesn't have to stay open and get me 10s lost each time I browse the issues since there is no action to be done here.

gep13:
In the case of this issue, I think I would agree with you, but in general, no I don't agree.  

Creating an issue first to discuss a topic is far better than discussing in a PR.  Too many times I have been in a situation where work has been done in a PR that hasn't been able to be merged (which means wasted effort on the part of the contributor) that could have been saved by first discussing the intended approach and implementation in a related issue first.

>@majkinetor said...
it just doesn't have to stay open and get me 10s lost each time I browse the issues since there is no action to be done here.

That would be where usage of Labels would come into play, similar to what @ferventcoder does extensively on the Chocolatey Repository.

majkinetor:
If I ever want something **generally**, I will always explicitelly say so. Otherwise, everything is contextual.

PR #257, opened 2016-07-12
(new package) eclipse

030:


gep13:
Relates to #256

gep13:
@030 I thought your packages were automatic packages?  This is not a problem if not, I just thought that the main problem that you had was that you didn't have a windows server to run ketarin in order to continue automatically pushing the packages.

Also, with regard to your common functions, I think that it might be better to include them as an extension package, and then take a dependency on that package, rather than dot sourcing them.  That way, if anything changes in one of those files, you can push out a new version of the extension package, without touching the actual package itself.  You can see examples of these in the extensions folder in this repository.

gep13:
@030 did you happen to see my comment above?  Thanks!

majkinetor:
I think we should close this as there is no feedback.

gep13:
Agreed

PR #259, opened 2016-07-15
Create chocolateyUninstall.ps1

RedBaron2:
@gep13 This is a Uninstall file for the Chromium Package. This will find if a registry key exists, and then find the uninstall string from the registry. This may not be silent


gep13:
Ok, let's give this a try :smile:

Issue #111, opened 2016-01-15
Add the \Scripts directory to the PATH when installing Python


ryanewtaylor:
When installing Python via Chocolatey I noticed that I was unable to use PIP from the command line as the python \Scripts directory was not added to the environment path. PIP is included in newer versions of Python and so it would be helpful to have this available on the PATH as well. 

To confirm the behavior of the MSI I ran it in GUI mode and via msiexec. The default behavior of both is to not add the python.exe and scripts dir to PATH but there is an option to do so in the MSI itself. You can do this through msiexec by adding `ADDLOCAL=ALL`. as an argument.

If this is done, perhaps Chocolatey would not need to add it to the path itself?

ferventcoder:
I agree this should be done.

ryanewtaylor:
If the maintainers do not have time for this addition, Is this something that I could potentially do via a pull request? I think the solution would be to:

1, Sign the CLA
2. Fork this repository
3. Create a branch such as `add-pythons-scripts-dir-to-path`
4. Add `ADDLOCAL=ALL` to `$partialInstallArgs = '/qn /norestart ALLUSERS=1 TARGETDIR='` as an argument in the chocolateyInstall.ps1 file
5. Remove `Install-ChocolateyPath $installPath 'Machine'` from the chocolateyInstall.ps1 file
6. Repeat for each python2 package (I would need to check what the behavior for the python3 installers are and see if they have similar requirements)
7. Submit a PR back to this repository

I read the contributors page. However I would want to be 100% sure of proper procedure so that I do not disrupt or offend any contributors and users. Please add or correct anything else you think is important and/or required for someone to make this sort of change.

ferventcoder:
I do not believe there is a CLA for the coreteam packages.

gep13:
@ferventcoder correct, there is no CLA for this repo.

@ryanewtaylor would be happy to accept a PR to get this change included.

obilodeau:
For Python3 this is done using `PrependPath=1` installer option: https://docs.python.org/3/using/windows.html

obilodeau:
Python 2's problem was likely fixed with: ce24b5c3bc and 5e214d3dd9 which introduced `\Scripts` injection into $env:PATH.

I'm doing a pull request for python 3. I tested it and it works (adding pip installed scripts to path). I can also add `ADDLOCAL=ALL` to Python 2 while I'm at it. Just let me know if interested.

[...]

PR #261, opened 2016-07-15
Added Python2 and Python3 Scripts dir to PATH using installer option (#111)

obilodeau:
This allows pip installed programs that provide scripts to be accessible from the `cmd.exe`. This is how this works on Linux.

Example without this option:
```
pip3 install git+https://github.com/GoSecure/malboxes.git@pip-packaging#egg=malboxes
C:\...> malboxes -V
'malboxes' is not recognized as an internal or external command,
operable program or batch file.
```

Example with this option:
```
pip3 install git+https://github.com/GoSecure/malboxes.git@pip-packaging#egg=malboxes
C:\...> malboxes -V
malboxes-script.py 0.2.0dev
```


obilodeau:
Added Python2 after discussion in #111.

gep13:
@obilodeau looks good to me.  Thanks for getting this done, and apologies for taking so long to get back to you about this.

PR #269, opened 2016-08-05
Chromium Clarification of Package

RedBaron2:
@gep13 Clarification of the install directory for those that may install this package in the future.


gep13:
@RedBaron2 looks good to me, thanks!

PR #275, opened 2016-08-20
Update chocolateyInstall.ps1

RedBaron2:
I noticed a glitch that even power shell didn't catch as being wrong. The  ":\" should only be ":" or the $chromium_string shouldn't start with a '\'
I also added the extra Argument of '--do-not-launch-chrome'. This was removed with all the confusion about the passing of variables to chocolatey.
@gep13 Tested and works


gep13:
LGTM! :+1:

PR #279, opened 2016-08-31
Update chocolateyInstall.ps1

RedBaron2:
Power Shell needs to read the $silentArgs value before it reads it from the $packageArgs

@gep13 I haven't found a solution other than to change the order. This should make the package follow the install that is listed on the website now :+1:


gep13:
LGTM! :+1:

PR #280, opened 2016-08-31
Update Clover.ketarin.xml

simonsdeng:
Fixes Clover download URL.

PR #283, opened 2016-09-01
Fix Clover install script

simonsdeng:
Fixes chocolateyInstall.ps1 from #280 (I think)


gep13:
Doh! Good catch.  Sorry, I should have spotted that :cry:

PR #286, opened 2016-09-11
Added checksum for autohotkey.portable

Elderry:


ferventcoder:
This is an auto package. It will need to be fixed in the ketarin job unless you are fixing the last version in the _output folder. HTH

Elderry:
Is there any tutorial for ketarin xml jobs? I went to ketarin.org and found almost everything is about the software itself.

ferventcoder:
@Elderry we have it all here https://github.com/chocolatey/chocolatey-packages-template and https://chocolatey.org/docs/automatic-packages

Elderry:
Now I have a problem: If one package has both 32-bit and 64-bit version, how can I use one Ketarin job to download them all and then use ChocoPkgUp to get the checksum of these files? Is AU a better approach according to this condition?

ferventcoder:
@Elderry the template already handles both. It's a pretty normal thing to see a package with both 32 and 64bit  urls.

Elderry:
Sent a [new PR](https://github.com/chocolatey/chocolatey-coreteampackages/pull/299) to replace this one.

PR #310, opened 2016-10-05
Au autohotkey

majkinetor:
Test for all updaters:

```
16-10-05 22:24:07  C:\Work\chocolatey-coreteampackages\automatic
> updateall autohotkey* -Options ( [ordered]@{Force = $true } )

Updating 3 automatic packages at 2016-10-05 22:21:31 (forced)
Push is disabled
FORCE IS ENABLED. All packages will be updated
   autohotkey is updated to 1.1.24.01
   autohotkey.install is updated to 1.1.24.01
   autohotkey.portable is updated to 1.1.24.01

Finished 3 packages after 0,58 minutes.
3 updated and 0 pushed.
0 errors - 0 update, 0 push.
```
Each is tested with `Test-Package`

### DIFF
--- a/automatic/autohotkey.install/tools/chocolateyInstall.ps1
+++ b/automatic/autohotkey.install/tools/chocolateyInstall.ps1
@@ -1,9 +1,31 @@
-﻿$name   = 'autohotkey.install'
-$url = '{{DownloadUrl}}'
-$url64 = $url
-$silent = '/S'
+﻿$ErrorActionPreference = 'Stop'
 
-$is64bit = (Get-WmiObject Win32_Processor).AddressWidth -eq 64
-if($is64bit) { $silent += ' /x64' }
+$packageName = 'autohotkey.install'
+$url32       = 'https://autohotkey.com/download/1.1/AutoHotkey112401_Install.exe'
+$url64       = $url32
+$checksum32  = '78ea671c3305465933a5d00f39da7e86c116a8a5a1849835196a27799ac29f8f'
+$checksum64  = $checksum32
 
-Install-ChocolateyPackage $name 'exe' $silent $url $url64
+$packageArgs = @{
+  packageName            = $packageName
+  fileType               = 'EXE'
+  url                    = $url32
+  url64bit               = $url64
+  checksum               = $checksum32
+  checksum64             = $checksum64
+  checksumType           = 'sha256'
+  checksumType64         = 'sha256'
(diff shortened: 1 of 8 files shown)


automatic/autohotkey.install/tools/chocolateyInstall.ps1
```diff
@@ -1,9 +1,34 @@
...
+  validExitCodes         = @(0)
+  registryUninstallerKey = 'autohotkey'
+}
+Install-ChocolateyPackage @packageArgs
 
-$is64bit = (Get-WmiObject Win32_Processor).AddressWidth -eq 64
-if($is64bit) { $silent += ' /x64' }
+$installLocation = Get-AppInstallLocation $packageArgs.registryUninstallerKey
+if ($installLocation)  {
+    Write-Host "$packageName installed to '$installLocation'"
```
ferventcoder: Chocolatey already does this - it feels like double reporting.

ferventcoder: I'd prefer only doing this for when Chocolatey isn't able to figure it out (and possibly migrating more of your additional checks into Chocolatey core).

gep13: @majkinetor are you in a position to update this PR to remove the additional Write-Host?

majkinetor: I wouldn't do that. This code works almost always, while internal chocolatey stuff works maybe 50% of the time or less.

Also, the function lets u actually use install location while the choco reported one is unusable to the maintainer. People may use this in future to improve on updater, for example to associate some files to ahk (historically ahkl extension was manually added while AHK_L was considered alternative)

majkinetor: Double reporting IF it happens, is not a big deal either. Better to report double, then none at all.

ferventcoder: It definitely feels like this logic should be in core Chocolatey. It already has some of this, it would not be a stretch to add the search in Porgram Files like you are doing.

majkinetor: Its not just search in program files. Its searching everywhere. I had at least 1 package that uses each of those paths. 

Yeah, it should be in choco I agree.

PR #311, opened 2016-10-05
Updated Fosshub extension to latest requirements and test with audacity

gep13:
@ferventcoder can you take a look at the following?  Thoughts?

### DIFF
--- a/extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
+++ b/extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
@@ -1,28 +1,48 @@
-# Get the resolved URL form a FossHub download link.
+# Get the resolved URL from a FossHub download link.
 #
-# Takes a FossHub URL in the “genLink” format and returns the generated
+# Takes a FossHub URL and returns the generated
 # expiring download link for the file.
 #
 # Usage: Get-UrlFromFosshub genLink_url
 # Example:
-# Get-UrlFromFosshub http://www.fosshub.com/genLink/Data-Crow/FILE.zip
+# Get-UrlFromFosshub https://www.fosshub.com/Audacity.html/audacity-win-2.1.2.exe
 
-Function Get-UrlFromFosshub($genLinkUrl) {
+Function Get-UrlFromFosshub($linkUrl) {
 
-  $fosshubAppName = $genLinkUrl -match 'genLink/(.+?)/'
+  $fosshubAppName = $linkUrl -match 'fosshub.com/(.*)/(.*)'
   # If there’s no match, it means that it’s not a FossHub genLink URL.
   # Then this function simply returns the input URL.
   if (!$Matches) {
-    return $genLinkUrl
+    return $linkUrl
   }
+
   # Get the actual FossHub appname from the matches array
+  # From the example above, this would be Audacity.html
   $fosshubAppName = $Matches[1]
 
-  $referer = "http://www.fosshub.com/${fosshubAppName}.html"
+  # Get the file name from the URL
+  # From the example above, this would be audacity-win-2.1.2.exe
+  $fosshubFileName = $Matches[2]
+
+  # Create the contents of the href that is being searched for
+  # From the example above, this would be /Audacity.html/audacity-win-2.1.2.exe
+  $hrefText = "/$fosshubAppName/$fosshubFileName"
+
+  # Construct the Regular Expression pattern.  An example of what is being searched for is this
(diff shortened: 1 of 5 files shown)


gep13:
Note, with the above changes, it installs correctly:

image

extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
```diff
@@ -1,28 +1,48 @@
...
+  # From the example above, this would be /Audacity.html/audacity-win-2.1.2.exe
+  $hrefText = "/$fosshubAppName/$fosshubFileName"
+
+  # Construct the Regular Expression pattern.  An example of what is being searched for is this
+  # <a href="/Audacity.html/audacity-win-2.1.2.exe" rel="nofollow" class="dwl-link xlink" file="audacity-win-2.1.2.exe" data="https://download.fosshub.com/Protected/expiretime=1475696791;badurl=aHR0cDovL3d3dy5mb3NzaHViLmNvbS9BdWRhY2l0eS5odG1s/606878a89d0b1e2bf998e0482bf081a451e8d98479d8b246a83d33db1a597538/Audacity/audacity-win-2.1.2.exe">
+  $regexPattern = "<a href=`"$hrefText`".*data=`"(.*)`""
+
+  $referer = "https://www.fosshub.com/${fosshubAppName}"
   $webClient = New-Object System.Net.WebClient
   $webClient.Headers.Add("Referer", $referer)
```
ferventcoder: It's possible you should look to make this more adapted to the proxy stuff that Chocolatey does - https://github.com/chocolatey/choco/blob/078cd3219c75328802c862cde9ee5e884e0478ca/src/chocolatey.resources/helpers/functions/Get-WebFileName.ps1#L108-L147

ferventcoder: Or a little more geared to WebClient - https://github.com/chocolatey/chocolatey.org/blob/8fe2b8d9ea6018c96e8eec6e6a6f244ca5e67a79/chocolatey/Website/Install.ps1#L87-L127

majkinetor: So, what is the real choco helper function we should use instead of `WebClient` ? `Get-WebFile` ? It looks like current helpers are not DRY - why do we have proxy stuff in both Get-WebFileName and Get-WebFile etc.. ?

gep13: @majkinetor I don't think we can use a Choco Helper.  In this instance, we have to alter the Referrer of the request, which is not possible in the current methods.  I was planning on recreating the GetDownloader function, and placing it into the FossHub Extension as a helper method.

majkinetor: I didn't have to alter referer with smplayer, that was true before....

Anyway, this is big issue IMO. People use WebClient all the time because helpers are not complete. Implementing better helper should be done so that people can stop using other methods.

ferventcoder: Incorrect - the helpers have a parameter called options where you can pass referer - this became available in 0.9.10.

extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
```diff
@@ -1,28 +1,48 @@
...
 # Usage: Get-UrlFromFosshub genLink_url
 # Example:
-# Get-UrlFromFosshub http://www.fosshub.com/genLink/Data-Crow/FILE.zip
+# Get-UrlFromFosshub https://www.fosshub.com/Audacity.html/audacity-win-2.1.2.exe
 
-Function Get-UrlFromFosshub($genLinkUrl) {
+Function Get-UrlFromFosshub($linkUrl) {
 
-  $fosshubAppName = $genLinkUrl -match 'genLink/(.+?)/'
+  $fosshubAppName = $linkUrl -match 'www.fosshub.com/(.*)/(.*)'
```
gep13: @ferventcoder done.

extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
gep13: @ferventcoder ah, ok, so I should be able to re-use the helper method, rather than copy pasta, which is what I was literally just doing 😄

gep13:
@ferventcoder @majkinetor can you both take a look at this?  I will get an update.ps1 file added to this later tonight, and assuming all goes well, I will get this merged in. Thanks!

PR #313, opened 2016-10-06
(lightalloy) Convert to using AU

gep13:


gep13:
@majkinetor can I get you to have a look at this please? Thanks!

automatic/lightalloy/tools/chocolateyInstall.ps1
```diff
@@ -1,24 +1,28 @@
...
-# Installer if the old silent mode would still work was easy
-#Install-ChocolateyPackage "$packageName" "$installerType" "$silentArgs" "$url" -validExitCodes $validExitCodes
+$pwd			= "$(split-path -parent $MyInvocation.MyCommand.Definition)"
+$au3			= Join-Path $pwd 'lightalloy.au3'
 
 # Need some AutoIt3 wizardry because installer is not silent
 $chocTempDir	= Join-Path $env:TEMP "chocolatey"
-tempDir		= Join-Path $chocTempDir "$packageName"
+$tempDir		= Join-Path $chocTempDir "$packageName"
 if (![System.IO.Directory]::Exists($tempDir)) {[System.IO.Directory]::CreateDirectory($tempDir)}
```
majkinetor: you mean `mkdir $tempDir -force` ? :)

gep13:
@majkinetor so, in order to see the automatic pushing and commiting working... Can I do the following?

The latest version of this package is 4.8.8.2, but the latest pushed to Chocolatey.org is 4.8.8.  If I change the nuspec version to say 4.7.7, check it in, and then let the AppVeyor build run, it should detect the new version, update the files, and push to chocolatey.org, right?

automatic/lightalloy/update.ps1
```diff
@@ -0,0 +1,32 @@
+import-module au
+import-module .\..\..\extensions\chocolatey-fosshub.extension\extensions\Get-UrlFromFosshub.psm1
```
majkinetor: I am not sure if this is necessary if its installed in the system but it wont harm either.

automatic/lightalloy/tools/chocolateyInstall.ps1
majkinetor: you mean `mkdir $tempDir -force ?` :)

majkinetor:
You have to put the version to something higher then already in community feed,  lets say 4.8.8.1 (or use `NoCheckChocoVersion`).

Then it should do all the stuff.

gep13:
@majkinetor why would it need to be higher? I thought the whole point was that if it found a version, which is higher than the one currently on chocolatey.org, it would update, and push the package out.  What am I missing?!?

majkinetor:
Oh yeah, you are right :) I am totally tripping.

gep13:
>@majkinetor said...
Oh yeah, you are right :) I am totally tripping.

ok, cool, we are on the same page then.  I will fix up this PR tomorrow, and set the nuspec version to be an earlier version, and then check to see if everything works correctly from AppVeyor execution.

I am expecting the following...

* Update nuspec and chocolateyInstall.ps1 with new values
* Create nupkg with higher version number
* Commit the changes to the file and push to GitHub
* Push the latest package to Chocolatey.org
* Update gist with what happened

Does that sound right?

majkinetor:
Thats the deal. And may send mail too :P

majkinetor:
Just make sure token is set right for github and with right permissions.

gep13:
>@majkinetor said...
Just make sure token is set right for github and with right permissions.

At the minute, it is using my credentials, but I may change that to be the chocobot account, but that will be a test for another day.

gep13:
>@majkinetor said...
Thats the deal. And may send mail too :P

I haven't set up the email configuration yet.  I am just using the built in appveyor mails for when a build fails.  I suspect you are sending emails at other times as well, so I think I will need to get this set up as well.

gep13:
Ok, I will get this tested out tomorrow, and I will let you know if I run into anything.  Thanks for your help!

majkinetor:
You already use token for the gist, you can use the same one for the git, just set it to public repo scope.

Mails are sent when update throws exceptions. Build will fail only if you make it so in the `update_all.ps1`

PR #314, opened 2016-10-07
Update chromium.ketarin.xml

RedBaron2:
Updated the api connections to be https


gep13:
@RedBaron2 thanks! 😄

gep13:
On a side note, you might have noticed that we are moving to using AU, rather than Ketarin for packages.  If you are in a position to, it would be great if you could convert the Chromium Package that use that.  If you have any questions, let me or @majkinetor know.

Thanks

PR #319, opened 2016-10-08
Replacing usage of System.Net.WebClient with choco internals

### LINKED ISSUE #305: Using System.Net.WebClient in some packages
Some packages use `System.Net.WebClient` to download pages for parsing, for example [Firefox](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/Firefox/tools/chocolateyInstall.ps1#L32).
I'm behind corporate proxy with NTLM-auth and this thing doesn't work for me. I got this error:
```
ERROR: Exception calling "DownloadString" with "1" argument(s): "The remote server returned an error: (407) Proxy Authentication Required."
```
When I configure webclient like this:
```powershell
$wc = New-Object System.Net.WebClient
$wc.Proxy.Credentials = [System.Net.CredentialCache]::DefaultCredentials
```
`DownloadString` does work.

ebugusey:
Fix #305


gep13:
@ebugusey thank you for taking the time to do this!

In principal, I like what you have done in this PR, with one exception... The duplication of the Get-WebContent.ps1 file seems like something we should take care of.  We already have examples on how to do this via the concept of a chocolatey extension.  In fact, we have two examples of these in this repository:

https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/extensions

It feels like the functions that you have added would be a good addition to a `chocolatey-install.extension`.  That way, each package could take a dependency on this extension, and therefore avoid the duplication of the code.  In addition, if anything changes in those functions, we only have to change it in one place, and it will work.

What do you think?  Would you be in a position to fix this up as described here?  Thanks!

ebugusey:
Did you meant `chocolatey-uninstall.extension`? Or should I create new one?

gep13:
@ebugusey I was thinking to create a new one.  Do you think that these functions would/could be used for both an uninstall and an install?

If so, perhaps we need to create an extension called `chocolatey-common.extension`.  Thoughts?

ebugusey:
I don't think they could be used for uninstall.
I will create new extension then.

gep13:
@ebugusey perfect! Thank you!

majkinetor:
That is exactly what I was planning to add to my extension. I think `chocolatey-common.extension` should be created for those stuff that miss in chocolatey. I could move functions in [mm-choco.extension](https://github.com/majkinetor/au-packages/tree/master/mm-choco.extension) to this one too.

One another thing that could be added there is `Unistall-ChocolateyPath` which is currently missing. Thats why `chocolatey-install` is not such a good name.

gep13:
@majkinetor @ebugusey I am happy either way.

gep13:
@majkinetor @ebugusey actually, anything that is specifically for use when installing a package should go in one called uninstall, but anything that is common should be in one called common.

Oh, and one thing while I remember...

The Get-WebFile provides the ability to pass in the referrer in the header, I would need that same ability to be exposed in the Get-WebContent function, if it isn't there already.

majkinetor:
> @majkinetor @ebugusey actually, anything that is specifically for use when installing a package should go in one called uninstall, but anything that is common should be in one called common.

I disagree. This is maintenance problem. The stuff is already bundled in helpers module in default chocolatey. 

This way I will need multiple dependencies specified instead of one (will also need to take care of multiple versions when one gets updated). 

I think that just one should be created that extends chocolatey defaults. In that case maybe `commons` is not best choice for name, more like `core-extensions` or something like that.

gep13:
Hmm, going to have to think about this one.

@ferventcoder thoughts?

ebugusey:
Am I working in the right direction?

gep13:
@ebugusey had a very quick glance, and yes, I would say you are.

I would still like to be able to pass in the same object array to the Get-WebContent function so that this can be passed through to the get-WebFile function.

extensions/chocolatey-core.extension/ReadMe.md
```diff
@@ -0,0 +1,9 @@
+This package provides helper functions installed as a Chocolatey extension.
+These functions may be used in Chocolatey install/uninstall scripts by declaring this package a dependency in your package's nuspec.
+
+```xml
+<dependencies>
+    <dependency id="chocolatey-common.extension" />
```
gep13: @ebugusey  I think this should be `chocolatey-core.extension`, right?

extensions/chocolatey-core.extension/ReadMe.md
```diff
@@ -0,0 +1,9 @@
...
+</dependencies>
+```
+## Functions
```
gep13: @ebugusey can you include an example of calling the function?

extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
```diff
@@ -0,0 +1,38 @@
...
+    <copyright>© 2016 Chocolatey Core Team Package Contributors</copyright>
+    <licenseUrl>https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/LICENSE.md</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <projectSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages</projectSourceUrl>
+    <!--<docsUrl>At what url are the software docs located?</docsUrl>-->
+    <!--<mailingListUrl></mailingListUrl>-->
+    <bugTrackerUrl>https://github.com/chocolatey/chocolatey-coreteampackages/issues</bugTrackerUrl>
+    <tags>chocolatey core extension admin</tags>
+    <summary>Helper functions extending core choco functionality</summary>
+    <description><![CDATA[__REPLACE__MarkDown_Okay ]]></description>
```
gep13: @ebugusey I think we could use a description here.

extensions/chocolatey-core.extension/extensions/Get-WebContent.psm1
```diff
@@ -0,0 +1,34 @@
...
+  $fileName = [System.IO.Path]::GetRandomFileName()
+
+  $tempFile = Join-Path $tempDir $fileName
+
+  $tempFile
+}
+
+function Get-WebContent ([string]$url) {
+  $filePath = GetTempFileName
+  Get-WebFile -url $url -fileName $filePath
```
gep13: @ebugusey have a look at this example:

https://github.com/majkinetor/au-packages/blob/master/nirlauncher/tools/chocolateyInstall.ps1#L9-L10

Sometimes, it is necessary to alter the options that are passed into Get-WebFile, and it would be good if we could continue to set these when calling Get-WebContent.

You are see how these parameters are available here:

https://github.com/chocolatey/choco/blob/c798ae3a1b79101814599143a0ed416a69dc13ff/src/chocolatey.resources/helpers/functions/Get-WebFile.ps1#L85-L91

gep13:
@ebugusey are you in a position to fix this up, or would you like me to take care of it? Thanks!

ebugusey:
@gep13 I can do that. Just didn't had free time in last 2 days.

gep13:
Nice one! Thank you!

gep13:
@ebugusey I have rebased and merged this here:

https://github.com/chocolatey/chocolatey-coreteampackages/commit/d4dec4c400b8c0f545c6e28eeeec3b4c0d2d3c6a

As a result, I can't close this PR in the normal way, however, your contribution is correctly attributed, and very much welcomed!  Thanks!

ebugusey:
Sorry it took so long.

gep13:
>@ebugusey said...
Sorry it took so long.

You have nothing to apologize for! Thank you for helping out in the first place!

PR #321, opened 2016-10-09
Notepadplusplus

majkinetor:
This is from @ferventcoder.

I updated all 3 packages, but can't push as I am not maintainer.  I will use in the future version 0.0 do denote that it is intented to run the updater.


majkinetor:
I had to use `NoCheckUrl` although I don't want to becasue my current implementation of this AU feature timeouts quite frequently. See https://github.com/majkinetor/au/issues/35

@gep13, can you reproduce this behavior or is it location specific ?

I can't figure out how to do this so far, everything I tried either downloads the entire file or doesn't work, so any help is highly appreciated. This check is very important in order not to push spoiled packages (it protects from vendor site changes). For some sites its OK to use `NoCheckUrl` (for example FossHub).

automatic/notepadplusplus.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,30 @@
...
+  url                    = $url32
+  url64bit               = $url64
+  checksum               = $checksum32
+  checksum64             = $checksum64
+  checksumType           = 'sha256'
+  checksumType64         = 'sha256'
+  silentArgs             = '/S'
+  validExitCodes         = @(0)
+  registryUninstallerKey = 'notepad\+\+'
+}
```
ferventcoder: There is almost zero need to repeat the urls and checksums, it would make more sense to push them all into the packageArgs instead of above and then just linked.

I'm assuming the regex could handle finding that properly.

gep13:
@majkinetor yes, I am seeing the same issue with the URL Checking as you mentioned.

automatic/notepadplusplus.install/tools/chocolateyInstall.ps1
majkinetor: Almost 0 isnt 0.

It must be se this way when u want to set url32 equal to url64 and the same for checksums.

Here it isnt mandatory but i dont like to change the template and regex just for that.

Plus,  the intene is clear here .... above lines are changed by the updater, bellow lines are not.

majkinetor:
@gep13

Thx for verifying. So this would definitelly need to get inproved.... i have no idea why that call timesout

PR #322, opened 2016-10-09
(opera) Updating to use AU

gep13:


gep13:
*NOTE:* Updated to use checksums, but will wait for next release before pushing, as Opera uses 4 part version number, and don't want to muddy the water with package fix notation.

PR #324, opened 2016-10-10
FossHub genLink URLs don't work and can be easily fixed without touching affected packages.

opello:
I don't know if this is something you want to do but it would probably save some user frustration with package maintainers that haven't updated download URLs to move away from ones that use the genLink web service.  I tested with the existing IrfanView (still uses genLink) and Audacity (does not use genLink) packages.

### DIFF
--- a/extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
+++ b/extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1
@@ -3,14 +3,14 @@
 # Takes a FossHub URL and returns the generated
 # expiring download link for the file.
 #
-# Usage: Get-UrlFromFosshub genLink_url
+# Usage: Get-UrlFromFosshub url
 # Example:
 # Get-UrlFromFosshub https://www.fosshub.com/Audacity.html/audacity-win-2.1.2.exe
 
 Function Get-UrlFromFosshub($linkUrl) {
 
   $fosshubAppName = $linkUrl -match 'fosshub.com/(.*)/(.*)'
-  # If there’s no match, it means that it’s not a FossHub genLink URL.
+  # If there’s no match, it means that it’s not a FossHub URL.
   # Then this function simply returns the input URL.
   if (!$Matches) {
     return $linkUrl


gep13:
@opello can you help me understand the need for this PR?  I recently updated the FossHub extension to work on all the packages in this repository that use FossHub, including Audacity, and all of them seem to work.

opello:
@gep13 Sorry, I thought it was clear in my initial comment.  The [IrfanView package][1] still uses a fosshub.com/genLink/... URL:

```
$packageName = 'irfanview'
$installerType = 'exe'
$urlArray = @('http://www.fosshub.com/genLink/IrfanView/iview442_setup.exe','http://www.fosshub.com/genLink/IrfanView/iview442_x64_setup.exe')
$url = $urlArray[0]
```

I didn't (out of ignorance about how to easily do so) search through all packages for other instances of the genLink component of FossHub URLs.  If IrfanView is the last one then maybe this is superfluous to updating that one package.  Since the translation was so simple it seemed very reasonable.

[1]: https://chocolatey.org/packages/IrfanView

gep13:
@opello ah, I see.  That package is not one of the ones that I look after, therefore it wasn't on my radar.  Rather than risk the existing packages not working, and the fact that genLink URL's don't resolve anymore on FossHub, I think the best plan would be to get the IrfanView package updated to use the latest FossHub Extension.

opello:
@gep13 I don't disagree with that perspective, it just seemed so easy to avoid user inconvenience.  Is there a simple way to search the package index for packages that need updating to alert the maintainers?

The comment change from baa6dcf48badee708e0289f9f9a911506f662a8d should still probably be addressed.

gep13:
Agreed, if you could update this PR to have only that commit, I will gladly pull it in.

@ferventcoder might be able to help in that regard, i.e. querying the packages for mention of Fosshub.

opello:
@gep13 Done.

Is the hang up about merging the migration code that you don't trust the reliability?  While I don't know much about FossHub their FAQ is pretty clear about URL structure and that it's designed to be structured in the way that the existing code already assumes as far as I can tell.  I'd appreciate any insight as to what problems this could introduce.  Thanks.

gep13:
@opello it's mainly that the existing genLinks, say in the irfanview package, simply doesn't work anymore, so why support them in the extension?

opello:
@gep13 Sure, I guess I've articulated my rationale, which is to avoid user inconvenience.

opello:
Thanks!  I pushed a copy of the original branch to [opello:fix/fosshub-migrate.orig](https://github.com/opello/chocolatey-coreteampackages/tree/fix/fosshub-migrate.orig) in case the migrating is ever reconsidered.

PR #326, opened 2016-10-11
Added Update-IconUrl script

AdmiringWorm:
Added a script to easily update or add new icon url to nuspec files
Works for a single package, or passing without a name tries to update all packages which have an icon matching the package name.

### DIFF
--- a/setup/Update-IconUrl.ps1
+++ b/setup/Update-IconUrl.ps1
@@ -35,6 +35,47 @@
 .NOTES
   Currently supports icons with the following extensions
   (png, svg, jpg, ico)
+
+.EXAMPLE
+  ps> .\Update-IconUrl.ps1
+  Updates all nuspec files with matching icons
+-    <iconUrl>https://cdn.rawgit.com/AdmiringWorm/chocolatey-packages/e4a49519947c3cff55c17a0b08266c56b0613e64/icons/thunderbird.png</iconUrl>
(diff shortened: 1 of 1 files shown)


majkinetor:
Can you please add regular Powershell documentation ? 
Is it idempotent ?

Can it handle situation when package is moved over repository. If I, for example move my packages here, icons will point to my repository and {user}{repo} should be changed.

AdmiringWorm:
> Can it handle situation when package is moved over repository. If I, for example move my packages here, icons will point to my repository and {user}{repo} should be changed.

@majkinetor with this script, it would require you to move the icons over as well.
then run `Update-IconUrl.ps1 "Insert Package Name"` which then searches for the
matching icon with the same name as the package, and updates the commit hash (committing the icon if it hasn't already) and updates the rawgit url with that commit has and the specified `GithubRepository` parameter (defaults to chocolatey/chocolatey-coreteampackages)

> Can you please add regular Powershell documentation ?

I'll look into trying to get some powershell documentation for this

majkinetor:
Could you please add some examples, those are more telling then all the docs combined.

AdmiringWorm:
@majkinetor something like that perhaps?

majkinetor:
Thanks for the work. I will try it as soon as I can, if somebody bits me to it let us know here.

AdmiringWorm:
ok, let me know if there is any problems, I'm still in the process of learning powershell so there is bound to be some hiccups

gep13:
@AdmiringWorm going to go ahead and pull this in, so that we can take it for a spin, we can PR again if additional changes are required.

AdmiringWorm:
sure thing

PR #328, opened 2016-10-11
(youtube-dl)Migrated to AU

AdmiringWorm:
### Other changes
* update.ps1 checks for url against github
* Updated license url
* Added project source url
* add bug tracker url
* Add docs url (it's the url which is on their main page)
* Updated release notes
* Updated copyright
* Added vcredist2010 as a dependency (msvcrt100.dll is needed)


automatic/youtube-dl/update.ps1
```diff
@@ -0,0 +1,41 @@
...
+
+  return @{
+    Version = $version
+    URL32 = $url
+    Checksum32 = $checksum
+    ChecksumType32 = "sha512"
+  };
+}
+
+update -ChecksumFor none
```
majkinetor: AU can actually verify this checksum before it finishes on public repo. Just don't use `none`. Since you gave AU the checksum it will switch to verification mode.

AdmiringWorm: I'm guessing though, that I must pass the checksum type?
Or do AU auto detect it?

automatic/youtube-dl/update.ps1
```diff
@@ -0,0 +1,41 @@
...
+      "(^[$]url\s*=\s*)('.*')"      = "`$1'$($Latest.URL32)'"
+      "(^[$]checksum\s*=\s*)('.*')" = "`$1'$($Latest.Checksum32)'"
+      "(^[$]checksumType\s*=\s*)('.*')" = "`$1'$($Latest.ChecksumType32)'"
+        }
+    }
+}
+
+function global:au_GetLatest {
+  $headers = @{}
+  if (Test-Path Env:\github_api_key) {
```
majkinetor: What actually happens here if no env var is present ?

I wanted to use Github API before but in my case releases page actually had more releases then what API returns (pre releases AFAIR).

automatic/youtube-dl/update.ps1
majkinetor: From: https://github.com/majkinetor/au#automatic-checksums
>If the ChecksumXX hash key is present, the AU will change to checksum verification mode - it will download the installer and verify that its checksum matches the one provided. If the key is not present, the AU will calculate hash with the given ChecksumTypeXX algorithm.

Default is `sha512`, otherwise you speify algorythm.

automatic/youtube-dl/update.ps1
AdmiringWorm: I haven't really hit any place that anyone had more releases then what the API returns.
However, the first item should always be the latest release in my experience.

automatic/youtube-dl/update.ps1
AdmiringWorm: according to your documentation it is sha256 by default

> If the checksum is actually obtained from the vendor's site, you can provide it along with its type (SHA256 by default) by returning corresponding ChecksumXX hash keys in the au_GetLatest function:

Or am I missing something big here

majkinetor: The documentation is always correct. I talk gibberish most of the time when I don't doublecheck:)

automatic/youtube-dl/update.ps1
majkinetor: Isn't just simple to not use API but web page ? API is uselful IMO when you have to pinpoint specific release, etc... with latest stuff, why bother ?

gep13:
@AdmiringWorm this is working fine locally, so happy to move this in once you have made your final changes.  Just let me know when ready.

AdmiringWorm:
@gep13 struggling a little with changing the call from api to the web page.
But once that is done I believe it ready

automatic/youtube-dl/update.ps1
majkinetor: I don't have a problem with your code and it is indeed the right thing to do. However, given the limitations which don't exist with web page and aditional setup I think its way more maintanable to not use API in general case.

AdmiringWorm:
Unless I missed something, I believe it is ready.
It ran until completion in *my test environment*

gep13:
>@AdmiringWorm said...
It ran until completion in my test environment

Oh well, I better run it on _mine_ as well then 😛

gep13:
@AdmiringWorm ok, this is strange...

What version of AU are you running locally?  I am trying to run the update script, using force for this package, and it hangs after this line:

`youtube-dl - checking updates using au version 2016.10.4.83756`

Did you see anything like that on yours?

AdmiringWorm:
locally I'm running 2016.9.25

gep13:
@majkinetor how would I have got `2016.10.4.83756` The latest pushed to PowerShell Gallery is `2016.9.25`

gep13:
@majkinetor @AdmiringWorm hmmm, even with `2016.9.25` installed, it still seems to hang on the update command.

Any ideas?  Is there a verbose log output for the update command?

AdmiringWorm:
@gep13 you don't have to force the package. the latest version with the new url is newer than what is hosted on chocolatey,
AU also hangs for me sometimes on 2016.9.25, but not on that line. Usually on the Automatic checksum line, usually fixes itself when I close powershell and reopen it.

AdmiringWorm:
In the update.ps1 script, change the line
`update -ChecksumFor 32`
to
`update -ChecksumFor 32 -Verbose -Debug`
to enable both verbose and debug

gep13:
Fair point, but even with no force enabled, it hangs...

gep13:
@AdmiringWorm @majkinetor verbose log...

```
VERBOSE: Running inside the script
youtube-dl - checking updates using au version 2016.9.25
```

PR #332, opened 2016-10-12
(youtube-dl) Updated to use au_afterUpdate

AdmiringWorm:


majkinetor:
You will need to push from your machine new package.

AdmiringWorm:
I can't, I'm not the maintainer of the package. that would be

majkinetor:
I think all the packages in core should have obligatory maintainer `chocolatey` besides any other. That way we can share chocolatey api key to selected few.

What do you think @gep13 ?

AdmiringWorm:
@majkinetor this package already have a `chocolatey` as a maintainer, I didn't mention it since I thought it already was obligatory.

gep13:
@majkinetor @AdmiringWorm yes, all packages in the Core Team Repo require that the chocolatey user be added as a maintainer.

We tend to leave the original package maintainer as a co-maintainer as well, that way they can administer it, if/when required.

In terms of sharing the Chocolatey User API Key, I am still trying to decide the best way to do that.

gep13:
For now, if a package, outside of the AppVeyor run, needs to be pushed, let me know and I can take care of it.

majkinetor:
> In terms of sharing the Chocolatey User API Key, I am still trying to decide the best way to do that.

What is wrong with sending it via mail directly to the member ? We can test that out and revoke it if needed. 

Alternativelly, more secure way would be to use Gpg and encrypt it via users public key which he must share here or in some public keyring. We could `cinst kleopatra` for this and for mail there is [mailvelope](https://www.mailvelope.com).

gep13:
That isn't really what I meant.

The chocolatey user is currently the same user that can also push to the official chocolatey package.  With the greatest of respect, neither Rob or I know you well enough to hand over the keys to that package, so I need to talk with @ferventcoder to decide how best to approach this. I think the best idea would be to create a user on chocolatey.org specifically for the core team repo, but again, haven't finished working out the details.

majkinetor:
> With the greatest of respect, neither Rob or I know you well enough to hand over the keys to that package, so I need to talk with @ferventcoder to decide how best to approach this.

I don't want the keys personally, current method is inefficient is all.

I have better idea.

How about I update `appveyor.yml` so that it handles special messages. Since everything is set there already it makes sense that AV pushes. This makes it possible for nobody to have a key except those that own the AV project.

So, when you merge and push is required, you can just set commit message to `[AU pkg1 pkg2 ... pkgN]` and it will force push them. We can also explicitelly set the version something like `[AU: pkg1|v1 pkg2|v2 ... ]`. So this can be done only by the chocolatey core members who need to merge PR anyway.

majkinetor:
Just to clarify, explicit version may be required for situations like this, where version is not changed but you need to repush because of verification failure.

majkinetor:
Maybe explicit version is bad idea... I could add new flag to AU, something like `NoVersionChange` as you typically want to keep existing version (force will use choco fix notation). So that would then become `[AU: pkg1! pkg2!]` where `!` marks this option.

PR #336, opened 2016-10-12
Change Chromium to use AU

RedBaron2:
@gep13 
This should work to update Chromium, but I have reservations about not having a double check of the checksum.
This file that AU finds the sha256 checksum of should also be checked. AU should check the file first against the package provided checksum ( { if provided } md5 in this case ) against the file downloaded before finding the sha256  checksum. This would be a double check to the integrity of the file being downloaded.

FYI: I know md5 is not the latest, but it is done at the time the file is first created.

### DIFF
--- a/automatic/chromium/tools/chocolateyInstall.ps1
+++ b/automatic/chromium/tools/chocolateyInstall.ps1
@@ -1,5 +1,5 @@
 
-  $version = '{{PackageVersion}}'
+  $version = '56.0.2890.0'
 
 	$chromium_string = "\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Chromium"
 	$hive = "hkcu"
@@ -12,7 +12,7 @@
   }
   
   $packageArgs = @{
-  packageName   = 'Chromium'
+  packageName   = 'chromium'
   fileType      = 'exe'
   url           = 'https://storage.googleapis.com/chromium-browser-snapshots/Win/425094/mini_installer.exe'
   url64bit      = 'https://storage.googleapis.com/chromium-browser-snapshots/Win_x64/425083/mini_installer.exe'
--- a/automatic/chromium/tools/chocolateyInstall.ps1
+++ b/automatic/chromium/tools/chocolateyInstall.ps1
@@ -12,16 +12,16 @@
   }
   
     $packageArgs = @{
-  packageName   = '{{PackageName}}'
+  packageName   = 'Chromium'
   fileType      = 'exe'
-  url           = '{{DownloadUrl}}'
-  url64bit      = '{{DownloadUrlx64}}'
+  url           = 'https://storage.googleapis.com/chromium-browser-snapshots/Win/425094/mini_installer.exe'
+  url64bit      = 'https://storage.googleapis.com/chromium-browser-snapshots/Win_x64/425083/mini_installer.exe'
   silentArgs    = $silentArgs
   validExitCodes= @(0)
   softwareName  = 'Chromium'
-  checksum      = '{{Checksum}}'
+  checksum      = '97da1f6f835b4b6ca7156c723e897083'
   checksumType  = 'md5'
-  checksum64    = '{{Checksumx64}}'
+  checksum64    = '0ffc77b8638e7f5c0d5bb8c6cf02aeff'
   checksumType64= 'md5'
 }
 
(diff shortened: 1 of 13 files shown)


gep13:
@RedBaron2 I will give this a try at lunch time to see how it looks.

gep13:
@majkinetor based on the issue that we were having with the youtube-dl package, are we saying that there is an issue with providing AU the checksum for validation?

The Chromium package is one that collects the checksum from a website, and it would be great for AU to verify this checksum against the file once downloaded.  I noticed that this was commented out:

https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/youtube-dl/update.ps1#L35

gep13:
@RedBaron2 can you also update this file:

https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/chromium/chromium.nuspec

To remove the tempalted variables?  You can also go ahead and delete the ketarin file for chromium.

majkinetor:
Yes for time being checksum verification mode works on its own but seems not to work inside a thread.

I suggest living everything in place in updater as it is just dont pass parsed checksum back until I see what is going on. 

I will have more time next week, but anybody can try to fix this and PR.

majkinetor:
https://github.com/majkinetor/au/issues/36

majkinetor:
There are multiple errors in Chromium package that need to be fixed + formating is really bad (mixed tabs and spaces identation etc.)

- Checksum in `chocolateyInstall.ps1` should be `sha256`
- Template string for package name isn't removed in Uninstall.
- `return $file` before call to `Uninstall-ChocolateyPackage` ?

Update.ps1 seems to be OK.

Furthermore, I wrongly merged 3dc87cc2 as it should be part of this one.

Please fix those errors and commit to this same branch.

RedBaron2:
I will finish this up soon

automatic/chromium/tools/chocolateyInstall.ps1
```diff
@@ -12,16 +12,16 @@
...
-  url           = '{{DownloadUrl}}'
-  url64bit      = '{{DownloadUrlx64}}'
+  url           = 'https://storage.googleapis.com/chromium-browser-snapshots/Win/425094/mini_installer.exe'
+  url64bit      = 'https://storage.googleapis.com/chromium-browser-snapshots/Win_x64/425083/mini_installer.exe'
   silentArgs    = $silentArgs
   validExitCodes= @(0)
   softwareName  = 'Chromium'
-  checksum      = '{{Checksum}}'
+  checksum      = '97da1f6f835b4b6ca7156c723e897083'
   checksumType  = 'md5'
```
majkinetor: Update.ps1 doesn't have set checksumType to md5 so it will use sha256. You need to change this.

automatic/chromium/tools/chocolateyInstall.ps1
```diff
@@ -12,16 +12,16 @@
   }
   
     $packageArgs = @{
```
majkinetor: Formating.

majkinetor:
> @majkinetor the Checksum is md5
> AU will change this when it downloads the new file, right?

No, you need to either return `ChecksumType32` and `ChecksumType64` in $Latest as `md5` in `update.ps1` or better, use `sha256` which AU uses by default in `chocolateyInstall.ps1`.

majkinetor:
> The use of the Chocolatey helper file Get-CheckSumValid.ps1 would allow for easy comparison of (if package supplied) checksum & AU's generated checksum of the file. This is also what Chocolatey uses at installation.

No, package supplied checksum is already handled by AU correctly, that is the whole point of this part of the framework - not to think about checksums at all. Checksum verification mode currently doesn't work right in threads and your update.ps1 is currently set not to use it. 

See https://github.com/majkinetor/au#automatic-checksums.

automatic/chromium/tools/chocolateyInstall.ps1
RedBaron2: Didn't know that AU had that ability. I do know that "sha256" is the more secure form, and it will be worked on in the future. I will add the check for the current provided type.

RedBaron2:
@gep13 @majkinetor 
That should be all the needed changes.  If anything else needed just comment :-)

majkinetor:
LGTM now. I hate to be to pedantic with you now, but please fix the ident in chocolateyInstall & Uninstall.
With that finished this is ready to be tested out.

Thanks for patience @RedBaron2.
@gep13, could you test it up in vagrant ? I currently don't have it setup as I am on the new machine.

gep13:
@majkinetor tested and merged.  I added some extra commits to fix up the indetantion.

PR #342, opened 2016-10-14
(suggestion) Updated editorconfig and gitattributes

AdmiringWorm:
Updated .editorconfig to get rid of the duplicated rules, as well as apply to all opened text files.

Updated .gitattributes file to normalize text files to crlf line endings, except for nuspec should normalize to lf line endings.
Added some basic (really basic) whitespace error checking when running `git diff`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,8 @@
 root = true
 
-[*.nuspec]
+[*]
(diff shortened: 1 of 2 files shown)


gep13:
@AdmiringWorm thanks for looking into this!

PR #343, opened 2016-10-14
Set "admin=prefers" for 64bit too

ConcurrentHashMap:
Got into trouble when trying to install this package with a non-privileged user account.
It makes no sense to me that for 64bit admin is required, while on 32bit it isn't.

Please correct me if I'm wrong.


gep13:
@RedBaron2 do you have any input on this change?  Thanks!

RedBaron2:
This all looks fine to me, but the ketarin file usage will be changed to AU soon.
@ConcurrentHashMap  Thanks for the PR
@gep13 LGTM

gep13:
>@RedBaron2 said...
This all looks fine to me, but the ketarin file usage will be changed to AU soon.

Agreed. 😄

PR #345, opened 2016-10-14
Chromium Fix & Update for future

RedBaron2:
The Uninstall file is fixed to correct its issue of not Uninstalling, and Update is updated to take advantage of the changes that will be taking place in the API for Chromium


RedBaron2:
@gep13 
Almost forgot the second half of the changes for the update file

RedBaron2:
@gep13 
Finishing touches done to the file. If you could test it and see if it all works 👍

RedBaron2:
@gep13 
I found a better way to get the information for the Uninstall of Chromium. This also uses less code.
I just have the problem of it not removing the lib\chromium folder. Any suggestions?

PR #348, opened 2016-10-23
Proposal of package kb2999226

RedBaron2:
@gep13 
A member on the gitter channel still really needs this package as a dependency for their package. This package was not in the chocolatey community feed. It has been tested and works. It will install the KB if not already installed, and skips install if already installed. Since it is a Microsoft KB no uninstaller in needed. The package should not need updating because it is static. Thank you for considering this package for the community feed.

### DIFF
--- a/automatic/kb2999226/kb2999226.nuspec
+++ b/automatic/kb2999226/kb2999226.nuspec
@@ -1,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8" standalone="yes"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    <id>KB2999226</id>
+    <version>1.0.20160915</version>
+    <title>Update for Universal C Runtime in Windows (KB2999226)</title>
+    <authors>Microsoft</authors>
+    <owners>RB</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <licenseUrl>https://support.microsoft.com/en-us/kb/2999226</licenseUrl>
+    <projectUrl>https://support.microsoft.com/en-us/kb/2999226</projectUrl>
+    <iconUrl>https://assets.onestore.ms/cdnfiles/onestorerolling-1509-17011/shell/v1_2/images/logo/microsoft.png</iconUrl>
+    <description>The Windows 10 Universal CRT is a Windows operating system component that enables CRT functionality on the Windows operating system. This update allows Windows desktop applications that depend on the Windows 10 Universal CRT release to run on earlier Windows operating systems.
+	Microsoft Visual Studio 2015 creates a dependency on the Universal CRT when applications are built by using the Windows 10 Software Development Kit (SDK). You can install this update on earlier Windows operating systems to enable these applications to run correctly.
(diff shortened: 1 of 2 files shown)


majkinetor:
One note: Get-HotFix is Powershell 3+ functions, and packages should work with Powershell2. As it is, this package wont work on vanilla Windows 7.

AdmiringWorm:
another note is, this update requires [kb2919355](https://chocolatey.org/packages/KB2919355) which is available on chocolatey.

gep13:
>@majkinetor said...
One note: Get-HotFix is Powershell 3+ functions, and packages should work with Powershell2. As it is, this package wont work on vanilla Windows 7.

Did you test this?

I have just ran this:

```
powershell -Version 2.0 Get-HotFix -id "kb2999226" -ErrorAction SilentlyContinue
```

and it output suggested that it works:

```
Source        Description      HotFixID      InstalledBy          InstalledOn
------        -----------      --------      -----------          -----------
MINTLAW       Update           KB2999226     NT AUTHORITY\SYSTEM  10/07/2015...
```

I don't have a vanilla Windows 7 machine to run on, but assuming that specifying the PowerShell Version Number works as I expect it does, then the Get-HotFix cmdlet works on Version 2.0.

majkinetor:
No I didn't. I just checked online. And you are right.

gep13:
>@majkinetor said...
No I didn't. I just checked online. And you are right.

Cool. Thanks for checking 😄 

So, assuming that we add a dependency on kb2919355, this one should be good to go, right?  If so, I can push it just now, as I am away to work on a few things.

gep13:
@AdmiringWorm while I see you online...

You ok with the above?

AdmiringWorm:
@gep13 with the exception of missing a dependecy for KB2919355, I'm all ok with it.
But I don't really know how to handle MSU updates on chocolatey, so I may not be the
correct person to ask though.

RedBaron2:
Just a note the kb is only for windows 7 with Service Pack 1 not vanilla.

gep13:
@RedBaron2 I have made some adjustments to the PR that you sent through, and have merged it into master.  The package is now in the moderation queue: https://chocolatey.org/packages/kb2999226

Thanks again!

ferventcoder:
@gep13 did you make adjustments so this package does not error when a platform is not supported?

jberezanski:
@majkinetor 

> One note: Get-HotFix is Powershell 3+ functions, and packages should work with Powershell2.

You were probably following the [TechNet documentation for Get-HotFix](https://technet.microsoft.com/en-us/library/hh849836(v=wps.620).aspx). The catch is that Microsoft has removed the docs for PowerShell versions older than 3.0, so the lack of "PowerShell 2.0" in the version selection combo box is not conclusive.

Archive.org to the rescue: https://web.archive.org/web/20090830212640/http://technet.microsoft.com/en-us/library/dd315358.aspx

gep13:
>@jberezanski said...
The catch is that Microsoft has removed the docs for PowerShell versions older than 3.0

WTF! Why would they do that?!?

jberezanski:
Bill^H^H^H^HSatya knows... Perhaps as a subtle message "you should stop using this" ;-). They have also taken down several old, but useful KB articles and downloads (even some still referenced by their other docs). A royal PITA, for sure.

Fortunately, .NET docs still go as far back as 1.1. We should count our blessings, I suppose.

gep13:
>@jberezanski said...
We should count our blessings, I suppose.

Ha ha, fair point 😄

majkinetor:
@jberezanski , you are totally right, thats what I did. Thanks for clarification.

ferventcoder:
They took down 2.0 a LONG time ago. Usually the v3 tells you whether it was introduced in v3 or not.

PR #354, opened 2016-10-24
(core extension) PowerShell v2 support

### LINKED ISSUE #352: (firefox) 49.0.2.20161023 install fails on PowerShell < 3
I get this error when upgrading firefox:

```
ERROR: A parameter cannot be found that matches parameter name 'Raw'.
The upgrade of firefox was NOT successful.
```

From a quick search, it could be powershell version related. See [this SO post](http://stackoverflow.com/questions/39580203/get-content-raw-parameter-error):

> The parameter -Raw was introduced with PowerShell v3.

My PS version is 2:

```
PS C:\Users\jlafrechoux> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
2      0      -1     -1
```

ebugusey:
Fix #352


gep13:
@ebugusey thanks for making this change!

PR #357, opened 2016-10-25
(gimp) Migrated to AU

AdmiringWorm:
* Added update.ps1 script
* Updated install script
* Updated icon url
* updated package source url

**NOTE:** Requires force push


automatic/gimp/tools/chocolateyInstall.ps1
```diff
@@ -1,9 +1,21 @@
...
+  packageName    = 'gimp'
+  fileType       = 'exe'
+  softwareName   = 'GIMP'
+
+  checksum       = '41e98ef75c2992acc9b481147bf3db35652a4c7e1da70bb06a2d2ec9a0853a2d'
+  checksum64     = '41e98ef75c2992acc9b481147bf3db35652a4c7e1da70bb06a2d2ec9a0853a2d'
+  checksumType   = 'sha256'
+  checksumType64 = 'sha256'
+  url            = 'https://download.gimp.org/mirror/pub/gimp/v2.8/windows/gimp-2.8.18-setup.exe'
+  url64bit       = 'https://download.gimp.org/mirror/pub/gimp/v2.8/windows/gimp-2.8.18-setup.exe'
```
AdmiringWorm: I'm a little unsure about this, but if I understood the docs <https://chocolatey.org/docs/helpers-install-chocolatey-package#url64bit-string> this is legit, when the installer contains both 32bit and 64bit executables (which it selects based on the OS architecture)

gep13: @AdmiringWorm if they are actually the same URL, and the download the same exe, then you shouldn't need the url64bit variable.

>Please note that the 32 bit url will be used in the absence of this.

gep13: @AdmiringWorm so, is that the case?  Does the URL in this case contain an exe that is both 32 and 64 bit enabled?  Or is it just a single 32 bit installer?

AdmiringWorm: Quote from the gimp site `The installer contains both 32-bit and 64-bit versions of GIMP, and will automatically use the appropriate one.`

AdmiringWorm: sure I'll add that

majkinetor: I think that code is correct. Otherwise it will always show 32x ver is installing.

AdmiringWorm:
Updated with the added comment

gep13:
@AdmiringWorm thanks for doing this one! Have just pushed a new package fix notation version to Chocolatey.org.

AdmiringWorm:
@gep13 No problem, always happy to help out.

AdmiringWorm:
@gep13 btw, any reason you just didn't add [AU: gimp] to the commit message when you merged this pr?
Wouldn't that have forced pushing a new version?

gep13:
It would have, but I pushed the big green button in the GitHub UI 😄

AdmiringWorm:
Ah, I thought you could change the merge message when you merge through github.
Maybe that's just when you rebase or squash then.

gep13:
@AdmiringWorm if you can do that, I would love to know about it, but I don't think you can 😢

AdmiringWorm:
This is what I usually gets after I press the merge/rebase or squash button on pull requests in my repositories.
image

With the exception of `Use your administrator privileges to merge this pull request` when all required
tests has passed.

Maybe this has something to do with me using protected branches, and requiring an accepted review
before merging them

AdmiringWorm:
Just found out, it doesn't work when rebasing. Only ordinary merge and squash

gep13:
Interesting, I have never seen that UI.  Would love to know how to enable that, as it would be really useful for the situation you mentioned above.

AdmiringWorm:
if it is dependent on protected branches, you can go into repository settings -> branches
and add the master branch to protected branches

majkinetor:
I have that UI too.

gep13:
@AdmiringWorm @majkinetor so you both get that UI when you click on the Merge Pull Request button?

image

AdmiringWorm:
@gep13 on my own repositories I do yes, except if I do a rebase instead of a merge.

gep13:
@AdmiringWorm so is this all you have done:

image

Or other stuff also?

Would you still get that UI if there was already a completed review?

AdmiringWorm:
I have a few other checks that needs to pass before I merge, but that should suffice.

It all depends on your requirements for this repository

AdmiringWorm:
Note though, that an administrator of this repository can still merge without an approved review.
He just gets an checkbox he needs to tick before merging.

AdmiringWorm:
sorry, missed the last question. I do always get that ui, with the exception of the `administration privileges`
checkbox if all checks has passed and it has one approved review

AdmiringWorm:
@gep13 could you remove the requirement for requiring an approved review?
it prevents majkinetor from pushing directly to the master branch.
it also seems that the reason you're missing the ui, must be related to something else.

gep13:
@AdmiringWorm @majkinetor removed the review required.

PR #360, opened 2016-10-26
SOFTWARE DEPRECATED: becyicongrabber

AdmiringWorm:
Since becyicongrabber hasn't gotten any updates since 2007, I've moved it to the
manual directory.
Before I did that though I did the following changes:
* added update.ps1 (just in case development starts anew)
* added BeCyIconGrabber.exe.gui
* update icon url
* updated install script

**NOTE:** It would need one last force push to add the checksum though


gep13:
Nice one! Thanks for doing this!

majkinetor:
> added update.ps1 (just in case development starts anew)

IMO this is waste of time.

AdmiringWorm:
@majkinetor, I had already created it before I decided to move it to manual

majkinetor:
BTW, I opted to delete ketarin files when I move package. First of all, its in the repo history if ever needed and it probably will never be needed.

PR #363, opened 2016-10-29
(libreoffice) AU; reduced chocolateyInstall; ketarin removed; nuspec updated

majkinetor:
I reduced install because I don't think version checking mechanism ever executes as the version in registry is different then the one reported remotelly (5.2.2. vs 5.2.2.2) and that is probably the same for other versions.

So please take a look and compare with old version.
I am not sure about comment on the first line either:

```
Important: The contents of this file – the `libreoffice` and       
`libreoffice-oldstable` install scripts – **must** be identical.
```


majkinetor:
@ferventcoder , @gep13 , PING

PR #364, opened 2016-10-30
[Firefox] The 64bits detection is overridden in all cases

wget:
The variable `$uninstallPaths` gets overridden whatever the 64bits detection could be. Prevent that.

### DIFF
--- a/automatic/Firefox/tools/chocolateyUninstall.ps1
+++ b/automatic/Firefox/tools/chocolateyUninstall.ps1
@@ -6,13 +6,13 @@ $silentArgs = '-ms'
   $regUninstallDir = 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\'
   $regUninstallDirWow64 = 'HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\'
   
+  $uninstallPaths = $(Get-ChildItem $regUninstallDir).Name
+  
   if ( Get-ProcessorBits 64 ) {
       if (Test-Path $regUninstallDirWow64) {
         $uninstallPaths += $(Get-ChildItem $regUninstallDirWow64).Name
       }
   }
-  
-  $uninstallPaths = $(Get-ChildItem $regUninstallDir).Name
 
   $uninstallPath = $uninstallPaths -match
     "Mozilla Firefox [\d\.]+ \([^\s]+ [a-zA-Z\-]+\)" | Select -First 1


ferventcoder:
LGTM

wget:
@ferventcoder Obviously, it depends what the maintainer wanted to do here with the first match: either take the first match from the standard location (x86 location on a 32 bits system, 64 bits location on a 64bits system) or take the match from the 32 bits location on a 64 bits system (the SysWoW64 location).

Otherwise we could have just let the line 15 as it and only replace the `=` by `+=`. But again, this depends what the maintainer wanted to do. Especially as from the tests I've made:
* the non SysWow64 location is used
  * on a 32bits system when using the 32 bits version of Firefox;
  * on a 64 bits system when using the 64 bits version of Firefox.
* the SysWow64 location is used
  * only on a 64 bits system when using the 32 version of Firefox.

I realized if we install firefox on a 64 bits system using Chocolatey, the 64 bits version is installed, and if we install manually the 32 bits version afterwards, Chocolatey using the Auto Uninstaller feature is removing the 64 bits version correctly.

*However* for some packages I'm currently updating, how can I force my uninstall script to be run when the user left the Auto Uninstaller feature enabled? This feature is rather annoying since the customizations I've made in my install scripts are *NOT* reverted when the AutoUninstaller feature is run :-/ 

Any idea?

majkinetor:
> However for some packages I'm currently updating, how can I force my uninstall script to be run when the user left the Auto Uninstaller feature enabled? This feature is rather annoying since the customizations I've made in my install scripts are NOT reverted when the Auto Uninstaller feature is run :-/

AFAIK, `chocolateyUninstall.ps1` runs before auto uninstaller. So do you talk about packages that do not have it ?

wget:
@majkinetor From the tests I performed on my side, it looks like this is not what is happening. Actually I'm still testing locally using `cpack` then `choco uninstall -fdv .\openvpn.nuspec --yes`, so I don't know if this changes anything :-/

majkinetor:
Is there any such package in this repo ? If not, I suggest you to PR one so we can take a look.

> Actually I'm still testing locally using cpack then choco uninstall -fdv .\openvpn.nuspec --yes, so I don't know if this changes anything :-/

You should take a look at AU's `Test-Package` that can test both locally and inside VM which is good for reproducibility on this referent machine.

wget:
@majkinetor Actually this is the package OpenVPN I'm talking about. A dev version of the new version I want to publish is available on my [GitHub repo](https://github.com/wget/chocolatey_openvpn)

gep13:
@wget @ferventcoder @majkinetor so, what was the consensus on this?  Happy to merge?

ferventcoder:
@wget I think we determined that chocolateyUninstall.ps1 always runs before autoUninstaller in our conversation. We also found there may be a small bug in autoUninstaller as well. https://github.com/chocolatey/choco/issues/1035

@gep13 This PR LGTM!

wget:
Damn. I had to RTFM to understand what LGTM means :D

gep13:
>@wget said...
Damn. I had to RTFM to understand what LGTM means :D

Ha ha! 😄

PR #380, opened 2016-11-08
Moved CutePDF to AU.

mikecole:


automatic/PDFCreator/PDFCreator.nuspec
```diff
@@ -17,8 +17,4 @@ PDFCreator – PDFCreator easily creates PDFs from any Windows program. Use it l
     <tags>pdf admin</tags>
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages</packageSourceUrl>
```
majkinetor: https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/PDFCreator

majkinetor: Its also good chance (although not mandatory) to update other metadata:
- Give some features
- Add some more tags such as `editor`, `office` etc.
- Add docs, bugtrackerurl, mailinglist etc.

majkinetor: Also, its open source so, `projectsSourceUrl`: https://github.com/pdfforge/PDFCreator

automatic/PDFCreator/update.ps1
```diff
@@ -0,0 +1,24 @@
...
+     }
+}
+
+function global:au_GetLatest {
+    Write-Host $url
+
+    $temp_file = $env:TEMP + '\PDFCreator.exe'
+    Invoke-WebRequest $url -OutFile $temp_file
+
+    $version = (Get-Command $temp_file).Version
```
majkinetor: Interesting way to obtain the version. I like it.

However, its not appropriate here. The file size is ~27MB and the version is available on download page.

So you should find out something that produces faster update:
- Download page has info about the version.
- Blog seems to serve as release new information: http://www.pdfforge.org/blog, so you can take in their RSS feed (http://www.pdfforge.org/blog/rss) and filter items with PDFCreator in title and get the version.

majkinetor:
In advance, please use separate PR for each package,  please.

automatic/PDFCreator/PDFCreator.nuspec
mikecole: I was going more for a 1-1 conversion from ketarin to AU, but I'll add these.

automatic/PDFCreator/update.ps1
mikecole: The download page gives me 2.4. The blog gives me 2.4.0. Getting the version from the installer gives me 2.4.0.213. I wasn't sure what level of detail was most appropriate, so I went with most detailed. What is your opinion on which version is appropriate?

mikecole:
Sorry, didn't realize two packages snuck into this pull request.

automatic/PDFCreator/update.ps1
majkinetor: Its a good find. I had a same problem with LibreOffice - on site it was one version and it adds revision number. It was important to recognize what you had in order not to download if not needed. However, you solution downloads too, so it doesn't make a difference. But its great soluion when its very hard to parse the version or if its impossible.

Just go with the site 2.4 or blog 2.4.0 (site is probably what is expected by users tbh). Its better not to add revision as it can be used for choco fix.

gep13: I would go with 2.4.0, I.e. What comes from the blog post.

automatic/PDFCreator/PDFCreator.nuspec
majkinetor: This migration to AU is perfect time to tidy up metadata. You need to revisit everything anyway, and those information are really handy to have in `choco info <package>` or on site directly.

automatic/PDFCreator/update.ps1
mikecole: OK, I can look at the RSS feed for the first title that contains "released" and extract the version. It might not be perfect, but it's probably good enough.

automatic/PDFCreator/PDFCreator.nuspec
mikecole: 👍

automatic/PDFCreator/update.ps1
mikecole: Yeah that's probably the right choice.

majkinetor: I wouldnt add 0 just like that. What happens when they tomorrow release 2.5.0.1. You would get invalid version. I think u should use site version.

2.5 is not equal to 2.5.0 BTW.

gep13: @majkinetor ok, I have to ask, why is 2.5 not the same as 2.5.0?  In one, the patch version is explicitly being stated as 0, and the other, it just isn't being shown.

The suggestion wasn't to simply add a .0, but rather, if the version number is a 2 part version number, add the .0 onto it.

majkinetor: > The suggestion wasn't to simply add a .0, but rather, if the version number is a 2 part version number, add the .0 onto it.

I think this is irrelevant. 

> ok, I have to ask, why is 2.5 not the same as 2.5.0? 

Because `[version]2.5 -ne [version]2.5.0` :)

majkinetor:
@mikecole  I referenced few issues reported for those packages. Would u please take a look?

mikecole:
Sure will. Thanks!

automatic/PDFCreator/update.ps1
gep13: @majkinetor ah, I see the problem here... We are not talking about the same thing.  I was talking about Semantic Versioning.

image

AdmiringWorm: I see the PDF creator uses minimum tree part version, probably best to follow the existing convention.
https://chocolatey.org/packages/PDFCreator

majkinetor: That makes sense.

mikecole:
@majkinetor I believe I fixed both linked issues.

automatic/PDFCreator/update.ps1
```diff
@@ -0,0 +1,27 @@
+import-module au
+
+$download_page_url = 'http://download.pdfforge.org/download/pdfcreator/list'
+
+function global:au_SearchReplace {
+    @{
+        'tools\ChocolateyInstall.ps1' = @{
+            "(^[$]checksum\s*=\s*)('.*')"   = "`$1'$($Latest.Checksum32)'"
```
majkinetor: You don't have an URL here. It will never update it.

automatic/cutepdf/update.ps1
```diff
@@ -0,0 +1,24 @@
...
+        'tools\ChocolateyInstall.ps1' = @{
+            "(^[$]checksum\s*=\s*)('.*')"   = "`$1'$($Latest.Checksum32)'"
+        }
+     }
+}
+
+function global:au_GetLatest {
+    Write-Host $url
+
+    $temp_file = $env:TEMP + '\CuteWriter.exe'
```
majkinetor: The same as before here, without downloading the files if other things are possible.

mikecole:
@majkinetor fixed and fixed.

automatic/PDFCreator/PDFCreator.nuspec
```diff
@@ -14,11 +14,9 @@
 PDFCreator – PDFCreator easily creates PDFs from any Windows program. Use it like a printer in Word, StarCalc or any other Windows application.
     </description>
     <summary>PDFCreator - PDFCreator easily creates PDFs from any Windows program.</summary>
-    <tags>pdf admin</tags>
-    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages</packageSourceUrl>
+    <tags>pdf editor printer admin</tags>
```
majkinetor: Should be tagged with [fos](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guideliness-%5BDraft%5D#122-obligatory-tags).

automatic/PDFCreator/update.ps1
```diff
@@ -0,0 +1,28 @@
...
+    $version = $newest_release.innerText.Replace("PDFCreator ", "").Trim()
+
+    $newest_release_url = $html.body.getElementsByTagName("a") | where {$_.href -like "*/download/pdfcreator/$version*"}
+
+    $url = $newest_release_url.href.Replace("about:/download", "http://orange.download.pdfforge.org") + "&download"
+
+    return @{ URL = $url; Version = $version }
+}
+
+update -NoCheckUrl -ChecksumFor 32
```
majkinetor: AU complains about URLs ? hence `NoCheckUrl` ?

automatic/cutepdf/cutepdf.nuspec
```diff
@@ -1,10 +1,9 @@
...
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
   <metadata>
-    <id>{{PackageName}}</id>
+    <id>cutepdf</id>
     <title>CutePDF</title>
-    <version>{{PackageVersion}}</version>
+    <version>3.0</version>
     <authors>Acro Software Inc.</authors>
     <owners>Redsandro</owners>
     <summary>Free PDF printer driver with no watermarks.</summary>
```
majkinetor: Please add '##Features'

automatic/cutepdf/cutepdf.nuspec
```diff
@@ -1,10 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
-<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
```
majkinetor: Add `freeware` among tags.

majkinetor:
Also, add releaseNotes. I see that CutePDF has them at http://www.cutepdf.com/support/updates.asp and the other package certainly has it.

Those are all metadata problems. When you fix it I will merge it.

automatic/PDFCreator/PDFCreator.nuspec
AdmiringWorm: that would actually be `foss`, `fos` means something completely different

majkinetor: Yeah, my bad.. foss

gep13:
@mikecole are you in a position to update this PR with the suggestions, or are you happy for us to make the modifications and merge?

automatic/PDFCreator/update.ps1
mikecole: No, that was copied and pasted from another package. I'll fix it.

mikecole:
@gep13 yes I can fix these in the next few days, unless you'd like them sooner.

PR #386, opened 2016-11-11
(codeblocks) Migrated to AU [AU codeblocks]

AdmiringWorm:


majkinetor:
Failed:

> The term 'Get-WebContent' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

majkinetor:
The problem here is that `Get-WebContent` is from the core package. This is one of the reasons developing with extensions that are all around and not in single module wrong @gep13. Its contraintuitive of how things usually work.

majkinetor:
Now, the foshub is dependend on it but not we include it in update.ps1 as a module. Then you have to replicate dependency mechanism in code which is horrible. 

With merged core extension this would still happen in this case as FosHub is not part of it tho.

Another solution would be to preinstall all core extensions before running update on server then we can just call updater without importing any module.

AdmiringWorm:
Odd, worked fine on my computer.
https://gist.github.com/AdmiringWorm/41bc9eef6bb84a3fcae21fda1ab8eb44/efbc2dc17170de8717ceb9dce3acaa53b0dc26fd
I'll check out what happened. (Can't now though, the power is out)

majkinetor:
It worked because you have it installed in your local choco.

majkinetor:
Lets see if it works now.

majkinetor:
OK, worked now  [report](https://gist.github.com/bd2eaa76f2a9ab739ca0544c502dca6e/84baa33fbf1b9c757bc26cc5f560ae165004b068)

PR #390, opened 2016-11-12
Core extension

gep13:


extensions/chocolatey-core.extension/extensions/Get-AppInstallLocation.ps1
```diff
@@ -0,0 +1,70 @@
...
+        [string] $AppNamePattern
+    )
+
+    function strip($path) { if ($path.EndsWith('\')) { return $path -replace '.$' } else { $path } }
+
+    $ErrorActionPreference = "SilentlyContinue"
+
+    Write-Verbose "Trying local and machine (x32 & x64) Uninstall keys"
+    [array] $key = Get-UninstallRegistryKey $AppNamePattern
+    if ($key.Count -eq 1) {
```
AdmiringWorm: Couldn't this possibly cause a problem?
Let's say that the `Get-UninstallRegistryKey` finds two keys with the same name, which points to
lets say the exact same path.
Personally I think that should be handled correctly, for instance by allowing a user
to override the exact count expected?

gep13: @majkinetor what are your thoughts?

majkinetor: How it could cause a problem ? If there are two keys, it will just skip this entire section. I could select -first 1 also but the point of the function was to require no special input. 

What I did it in the past is change as soon as some package appears that it can't find but could actually find it with minor modifications. 

So lets keep it like this now, its already quite tested, and will see in the future when problem arises.

majkinetor:
BTW, Uninstall-ChocolateyPath (#310) is something that naturally belongs here.

majkinetor:
I fixed one small error. It should be good to push now I believe.

gep13:
>@majkinetor said...
Or maybe not, I didn't know there is so much activity about it recently. It will prolly get soon included among helpers.

Yes, I believe that we should wait before doing anything in this area.

PR #391, opened 2016-11-12
(codeblocks) Removing unnecessary dependency

gep13:
@majkinetor @AdmiringWorm why was this extra dependency required?  The fosshub extension already has a dependency on the core extension, and should be resolved that way.

### DIFF
--- a/automatic/codeblocks/codeblocks.nuspec
+++ b/automatic/codeblocks/codeblocks.nuspec
@@ -31,7 +31,6 @@ This package downloads the installer with the included GCC compiler.
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/codeblocks</packageSourceUrl>
     <dependencies>
       <dependency id="chocolatey-fosshub.extension" version="0.2.0" />
-      <dependency id="chocolatey-core.extension" version="1.0" />
     </dependencies>
   </metadata>
   <files>
--- a/automatic/codeblocks/update.ps1
+++ b/automatic/codeblocks/update.ps1
@@ -1,6 +1,6 @@
 import-module au
 import-module "$PSScriptRoot/../../extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1"
-import-module "$PSScriptRoot/../../extensions/chocolatey-core.extension/extensions/Get-WebContent.psm1"
+. "$PSScriptRoot/../../extensions/chocolatey-core.extension/extensions/Get-WebContent.ps1"
 
 $releases = 'http://www.codeblocks.org/downloads/26'
 


AdmiringWorm:
when running AU it would fail because it wasn't pulled in automatically in the update.ps1 script.
no need to add the dependency in the nuspec file, must be fixed in the update.ps1 file since the extension of `Get-WebContent` changed.

AdmiringWorm:
that is, the extension changed from `.psm1` to `.ps1`
so the line in `update.ps1` must be changed from:
`import-module "$PSScriptRoot/../../extensions/chocolatey-core.extension/extensions/Get-WebContent.psm1"`

to
`import-module "$PSScriptRoot/../../extensions/chocolatey-core.extension/extensions/Get-WebContent.ps1"`

gep13:
@AdmiringWorm would it not just be a dot source of the file?  Rather than an Import-Module?

AdmiringWorm:
I don't know, possibly. I tried it locally with import though, and that works.

AdmiringWorm:
Just tried it with a dot source, and it also works. So apparently it doesn't matter if it is dot source or import-module

gep13:
@AdmiringWorm in this case, I think dot source is more correct, as it isn't actually a module.

gep13:
@AdmiringWorm @majkinetor I have added another commit to this PR.

AdmiringWorm:
just tested to be absolutely sure, and it works now. Want to merge this now, or wait for majkinetor?

gep13:
@AdmiringWorm let's wait and get final confirmation.

AdmiringWorm:
@gep13 I looked through the history of codeblocks.nuspec.
It originally had chocolatey-uninstall.extension as a dependency, so it was changed when you ran the updating of chocolatey-uninstall to chocolatey-core

gep13:
@AdmiringWorm Yes, that is correct.  Is that a problem?

AdmiringWorm:
nah, since the `Get-UninstallRegistryKey` function is in the core extension on it shouldn't be.
Just mentioned it since you initially asked why that dependency was required.
It was just auto updated, that's all.

gep13:
Gotcha. I updated it when I was updating all the others, for consistency sake, and then raised the question about whether it was actually required.

majkinetor:
LGTM.

majkinetor:
Honestly, I wouldn't pinpoint specific ps1. I would just use this:

```
import-module "$PSScriptRoot/../../extensions/chocolatey-core.extension/extensions/chocolatey-core.psm1"
```

majkinetor:
> Don't know why it was initially added to the nuspec dependency element though.
@majkinetor ?

Not sure, probably a mistake. Foshub already depends on it so its redudant. 

I will merge this now.

majkinetor:
5b3e0777ab8a4f3b2e98d26bde810ad2d36

PR #392, opened 2016-11-12
(fosshub) Updated depdendencies

gep13:
@majkinetor @AdmiringWorm we can pull this one in once the outstanding question about the core-extension has been answered.

### DIFF
--- a/automatic/audacity/audacity.nuspec
+++ b/automatic/audacity/audacity.nuspec
@@ -17,7 +17,7 @@
     <releaseNotes></releaseNotes>
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages</packageSourceUrl>
     <dependencies>
-      <dependency id="chocolatey-fosshub.extension" version="0.3.0" />
+      <dependency id="chocolatey-fosshub.extension" version="0.4.0" />
     </dependencies>
   </metadata>
   <files>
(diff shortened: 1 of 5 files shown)


gep13:
@majkinetor @AdmiringWorm do you think we should force push all the packages in this PR so that they update with the latest dependencies?

AdmiringWorm:
I think that should be possible.
Not sure though.

AdmiringWorm:
but is it really necessary?
Sure the uninstall extension is deprecated, but it has a dependency on the core extension so no harm not updating them?

gep13:
Oh, yeah, everything should just work, I am not worried about that side of things.

More, just whether it is worth pushing out an updated package fix with the most recent dependencies, or just wait for the natural order of things to sort things out.

AdmiringWorm:
personally, I feel this is such a small change that it isn't wort pushing out a package fix.
But that may just be me.

gep13:
>@AdmiringWorm said...
personally, I feel this is such a small change that it isn't wort pushing out a package fix.
But that may just be me.

Nope. that is my opinion as well.  Just wanted to make sure 😄

majkinetor:
I agree, if the package doesn't require new version of extension lets keep it that way. I update pacakge dep only when bug fixing extensions.

PR #396, opened 2016-11-13
WinPcap

majkinetor:
I fixed this one, I beleive few tools depend on it.

I couldn't execute AHK using `Start-ChocolateyProcessAsAdmin` so I invoked AHK as is.
If somebody wants to fix it, please do. If not, I will push it as it is.


majkinetor:
O yeah, it also depends on core extension 1.0.

majkinetor:
I thought its not released yet. Ignore.

gep13:
>@majkinetor said...
I thought its not released yet. Ignore.

Ah, gotcha.  I pushed it last night, and that is why I merged @AdmiringWorm PR for aptana-studio.

automatic/WinPcap/update.ps1
```diff
@@ -0,0 +1,27 @@
+import-module au
+
```
AdmiringWorm: Missing import for the `Get-PackageCacheLocation` function?

majkinetor:
No... i imported core.. no need to cheery pick.

AdmiringWorm:
@majkinetor 
sure, no need to cherry pick. but there isn't any import for core though.

majkinetor:
Meh... lost during branching and merging I guess, it was there :). 

Added.

majkinetor:
If you are not against this line:

```
AutoHotkey $ahkScript $packageArgs.fileFullPath
```

I would merge. I think its what @ferventcoder called antipattern, but I am not 100% sure.

I tried above line as 

```
Start-ChocolateyProcessAsAdmin  "$ahkScript '$($packageArgs.fileFullPath)'" 'Autohotkey'
```

But AHK complained that it can't find the path. I fail to see how can I make it work hence I used direct call.

gep13:
What does $toolsPath contain at runtime? An absolute path, or a relative path?

AdmiringWorm:
@majkinetor I tried a few different ways to call the `Start-ChocolateyProcessAsAdmin`, no luck though.

@gep13 I think the `$toolsPath` is an aboslute path during runtime.

majkinetor:
I almost always use absolute paths. That is the only way to make sure script run no matter what is current dir.

I think we should merge this. If the need arises we can always later fix it.

AdmiringWorm:
I agree.

@gep13 ?

PR #401, opened 2016-11-15
Encfs4win

majkinetor:
Can somebody besides me test this please, just to make sure ? It requires ~20GB of disk for one and good network connection due to deps.

At the end, console should have `encfs` command line and WIN+R `encfsw` should work too.

Restart might be required.


majkinetor:
In my test, it doesn't work after restart either. But it works when I reinstall the app after the restart. It seems that installing dokany library isn't successifull until service pack is active.

Please confirm, and we can merge it I guess.

majkinetor:
BTW, this is the reason for needed storage space:

`Progress: 100% - Completed download of C:\Users\Administrator\AppData\Local\Temp\chocolatey\KB2919355\1.0.20160915\Windows8.1-KB2919355-x64.msu (690.78 MB).`

AdmiringWorm:
@majkinetor I don't know why it depends directly on the KB2919355, it should actually depend on the `Universal C Runtime in Windows (KB2999226)`, but this do depend on KB2919355

that is according to microsofts answer here: https://connect.microsoft.com/VisualStudio/Feedback/Details/1574311

majkinetor:
Yeah, I did put dep only on 2999226

majkinetor:
> Posted by Microsoft on 8/12/2016 at 4:04 PM

> The 2015 VC Redist installs and has dependencies on the Update for Universal C Runtime in Windows
> (https://support.microsoft.com/en-us/kb/2999226)
> 
> To install this update, you must have April 2014 update rollup for Windows RT 8.1, Windows 8.1, and Windows Server 2012 R2 (2919355) installed in Windows 8.1 or Windows Server 2012 R2. (http://support.microsoft.com/kb/2919355)

AdmiringWorm:
that's the answer I mentioned.
anyways, still trying to install the KB219355 so I can test this package

AdmiringWorm:
just managed to install everything, and `encfs` is available on the run dialog for me.
However, trying to start the program gives the following System Error:
```
The program can't start because dokanfuse1.dll is missing from your computer.
Try reinstalling the program to fix this problem.
```

tried restarting as well, but no luck. same system error.

I think the dokanfuse1.dll library is provided by the `dokany` package, but I could be wrong

PR #403, opened 2016-11-16
Cygwin enhacements

majkinetor:
Please check it out due to popularity of the package:

- Added bunch of parameters requested here and there (see commit message)
- Made it work with cyg-get again
- Made it able to use proxy
- Uninstaller will delete cygwin files


majkinetor:
I accidentally commited here virtualbox small change, ignore it.

majkinetor:
And then I messed up some more rebasing. Ignore, will fix it before merge.

Issue #70, opened 2015-04-11
seafile-client: 1603 error


TomOne:
The seafile-client package throws an 1603 error when the same version is already installed.

TomOne:
Fixed with https://github.com/chocolatey/chocolatey-coreteampackages/commit/0a505678f3001cb0e88f94276e7257150ea4f0ac

PR #410, opened 2016-11-20
(seafile-client) Migrated to AU [AU seafile-client]

AdmiringWorm:


automatic/seafile-client/tools/chocolateyInstall.ps1
```diff
@@ -1,15 +1,24 @@
...
+
+  checksum     = '64f0e7192acbbbad4e6ac34d5b531ff27697172b9109f2c00cf61ebae923a642'
+  checksumType = 'sha256'
+  url          = 'https://bintray.com/artifact/download/seafile-org/seafile/seafile-6.0.0-en.msi'
+
+  silentArgs   = '/passive'
+  validExitCodes = @(0)
+}
 
 $app = Get-WmiObject -Class Win32_Product | Where-Object {
```
majkinetor: Doesn't this ignores the version ?
Also, this wmi class is **very** slow. Perhaps it would be better to check registry key as in uninstaller ? It would also give version.

But is it really necessary ? What is the position on this anyway @gep13 ? I usually remove those checks if files are not big (less then 100MB for me) and keep them on big downloads.

majkinetor: AH, ignore that about the version, I must be blind.

AdmiringWorm: It probably would be better to use a registry key, but I just reused this as it was the check already used originally.
The check is necessary because of #70, sure could add 1603 as a valid exit code.
But since 1603 is considered a fatal error, that wouldn't be correct.

PR #412, opened 2016-11-24
(Quicktime) Migrated to AU [AU Quicktime]

AdmiringWorm:


majkinetor:
It updated correctly but after that:

```
  Quicktime ERROR: 
     au_GetLatest failed 
      
     Access Denied 
      
     Access Denied 
       
     You don't have permission to access "http://support.apple.com/kb/DL837?" on this server. 
     Reference #18.8f9bfea5.1479983000.271da549 
      
      
     
```

AdmiringWorm:
that's odd, that didn't occur when I was testing it.
I wonder if they have some kind of rate limit or something.
Lets check on it in a few hours and if the issue still persist, I need to find out something....

AdmiringWorm:
looks like it resolved itself, most likely some restrictions apple had put up.

PR #422, opened 2016-11-27
(firefox) Migrated to AU +++

AdmiringWorm:
I tried to make this generic enough so it can eventually be reused by the thunderbird package.


majkinetor:
That was quite an effort!

majkinetor:
LGTM

majkinetor:
One thing is troublening me tho. I see you have non search and replace related code in `au_SearchReplace`. It would be more correct to move it to `au_BeforeUpdate`.

AdmiringWorm:
good call, I definitely should move the the checksum file creation to afterupdate.
I'm on my way out to shovel some snow, but I'll take a look it when I'm done.

majkinetor:
Snow... mmmm.

AdmiringWorm:
@majkinetor could you review the latest changes?

majkinetor:
Why do you have fixed version in installer (50.0) ? You could use $Env:ChocolateyPackageVersion ?

Otherwise looks good.

AdmiringWorm:
if I use `$env:ChocolateyPackageVersion` that would also include the package fix version, and that won't work.
And since it is already being replaced in searchandreplace function, it doesn't matter if it is a fixed version or not.

majkinetor:
Didn't think about that, you are right.

AdmiringWorm:
I merged it now, I didn't push it though.
Plan to do that at the same time as when the thunderbird package is finished.

majkinetor:
> if I use $env:ChocolateyPackageVersion that would also include the package fix version, and that won't work.

I came to say it doesn't matter as AU will change the version to fix one but I see now that you use `RemoteVersion` for that. I guess that few other packages have this 'fix' problem.

PR #427, opened 2016-11-28
smartftp: added KB2533623 dependency

matbech:
KB2533623 is only required on Windows 7 and Windows Server 2008 R2.
Note: package author update


gep13:
@matbech happy to merge, but as a suggestion, you might want to think about specifying a version of the package to depend on.

matbech:
@gep13 I was thinking about it as well. In the end the reason I left it out was that the dependency is a Windows KB from 2014 which won't be updated anymore. So once the package is installed, it should never be touched again. Even if the package version is changed, there isn't a need to "reinstall" the underlying KB (.msu).

PR #428, opened 2016-11-29
Adds Windows kb2670838 - Platform update

matbech:
https://www.microsoft.com/en-us/download/details.aspx?id=36805

### DIFF
--- a/manual/kb2670838/Tools/ChocolateyInstall.ps1
+++ b/manual/kb2670838/Tools/ChocolateyInstall.ps1
@@ -1,0 +1,46 @@
+$kb = "KB2670838"
+$silentArgs = "/quiet /norestart /log:`"$env:TEMP\$kb.Install.evt`""
+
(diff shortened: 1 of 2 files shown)


jberezanski:
LGTM!
I wonder if AU will push a package which does not exist in the community repo yet...?

gep13:
>@jberezanski said...
I wonder if AU will push a package which does not exist in the community repo yet...?

As long as the `chocolatey` user has been added as maintainer of the package, then I don't see why not 😄

jberezanski:
There was no package on chocolatey.org yet, so there was nothing to add the `chocolatey` user to. This was a completely new package.

But... it worked :) https://chocolatey.org/packages/kb2670838

gep13:
Ah, I see what you mean.  In which case yes, it is just a push of a new package, and will work in the same way as if doing it from your local machine, just uses the chocolatey user API key.

PR #437, opened 2016-12-01
(mkvtoolnix) Guidelines - Added to PATH - No Foshub [AU mkvtoolnix]

majkinetor:
No fix is needed, software updates quite frequently, and also this requires foshub extension fixed.


majkinetor:
One less foshub pacakge ^_^

PR #440, opened 2016-12-01
Fixed fosshub requiring referrer for downloads

### LINKED ISSUE #414: (fosshub.extension) Download URL within "xdts"-attribute
The package does no longer work for me because the download url of the binaries is stored within the xdts-attribute of the link now instead of the data-attribute.

https://github.com/chocolatey/chocolatey-coreteampackages/blob/e17820954cdce01bf1979a4ecaa7b7baa0612706/extensions/chocolatey-fosshub.extension/extensions/Get-UrlFromFosshub.psm1#L33

welwood08:
fixes #414 again

Hopefully this doesn't turn into whack-a-mole (ie if fosshub really doesn't want chocolatey doing its thing they come here and say so).

### DIFF
--- a/automatic/audacity/tools/chocolateyInstall.ps1
+++ b/automatic/audacity/tools/chocolateyInstall.ps1
@@ -8,6 +8,7 @@ $packageArgs = @{
   packageName    = $packageName
   fileType       = 'exe'
   url            = $url32
+  options        = @{ Headers = @{ Referer = 'https://www.fosshub.com/' } }
   checksum       = $checksum32
   checksumType   = 'sha256'
   silentArgs     = '/VERYSILENT'
--- a/automatic/avidemux/tools/chocolateyInstall.ps1
+++ b/automatic/avidemux/tools/chocolateyInstall.ps1
@@ -11,6 +11,7 @@ $packageArgs = @{
   fileType               = 'EXE'
   url                    = $url32
   url64bit               = $url64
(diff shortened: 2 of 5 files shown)


welwood08:
Thinking again, is there a more generic way to do this that doesn't need all consumers of the fosshub helper to make modifications?

AdmiringWorm:
unfortunately this is something I've been trying for some time now.
sometimes it works sometimes it doesn't.
Looks to be because fosshub sets such a short time before the links expires (around 4-5 seconds, could be faster)

AdmiringWorm:
> Thinking again, is there a more generic way to do this that doesn't need all consumers of the fosshub helper to make modifications?

Not that I know of, except implementing a new function users can call to get the headers.

welwood08:
Well if it would be something that required consumers of the fosshub extension to make changes it could surely just be a wrapper around `Install-ChocolateyPackage` that accepted the same parameters and performed the necessary preparation logic?

In other words, consumers would just call `Install-ChocolateyPackageFromFosshub`. That function could worry about finding the correct url/url64, setting the right headers, anything else fosshub might require in the future before calling `Install-ChocolateyPackage` itself.

AdmiringWorm:
that is of course an option, I'm off to bed for the night now but I'll try looking into something like that tomorrow.

majkinetor:
Outdated, see https://github.com/chocolatey/chocolatey-coreteampackages/pull/441

PR #441, opened 2016-12-02
[Abondoned] Implementing fosshub downloading

AdmiringWorm:
This is still a very much work in progress, but so far it works although au update is broken by this (need a way to detect if it's a au update run or not)

**NOTE** I know I'm missing comments, wanted this tested before I continued with it.

@gep13 @majkinetor


majkinetor:
Why do we need `Get-UrlFromFosshub $fosshubUrl`

`Install-ChocolateyFoshub` could call it ?

AdmiringWorm:
`Install-ChocolateyFosshub` do actually call it, or rather `Get-FosshubWebFile` does.
But yeah, it could probably be made a private function

extensions/chocolatey-fosshub.extension/extensions/Get-FosshubWebFile.ps1
```diff
@@ -0,0 +1,139 @@
...
+      }
+    }
+  }
+
+  if ($needsDownload) {
+    if (!($options["Headers"].ContainsKey('Referer'))) {
+      $referer = GetFosshubReferer $url
+      $options["Headers"].Add('Referer', $referer)
+    }
+    $downloadUrl = Get-UrlFromFosshub $url
```
majkinetor: Sorry, I didn't see it right. This is OK and func should probably not get exported (just name it with lower letters AFAIK, see psm1 convention). 

This is something we can do now since older packages do not work that depend on this func.

AdmiringWorm: 👍 I've made the changes locally, I'm waiting to push it until the rest have been reviewed and tested.
aslo `Get-FosshubWebFile` is already public I believe.
I've changed the lightalloy package to use it directly, so it is available at least.

majkinetor:
After skim read, I see one maintenance issue. You copied the code from helpers and when they change, this will not be in sync.

You need proxy functions for situations like this: https://mcpmag.com/articles/2013/11/05/blacksmith-part-7.aspx.

Now, there is no problem to go that way, especially in this moment, it can always be done later, just pointing it out.

majkinetor:
> need a way to detect if it's a au update run or not

I could provide this in AU, some env var for example. I never wanted for AU to be visible inside chocolateyInstall tho, will think about this.

AdmiringWorm:
> After skim read, I see one maintenance issue. You copied the code from helpers and when they change, this will not be in sync.

Yeah I know, I did see that as a necessety though since there is some code in the original that won't work for fosshub, and some things that need to be changed to make it work for fosshub.

> You need proxy functions for situations like this: https://mcpmag.com/articles/2013/11/05/blacksmith-part-7.aspx.

Did not know of this, but definitely something to be considered in the future.

> I could provide this in AU, some env var for example. I never wanted for AU to be visible inside chocolateyInstall tho, will think about this.

I can understand that, and usually it wouldn't be a need for it either.
and honestly I could probably work around this anyhow, by trying to use the `Get-FosshubWebFile` directly in `update.ps1` instead.

AdmiringWorm:
@majkinetor I think perhaps we should keep `Get-UrlFromFosshub` public, would make it easier to use in au update scripts.
In them there isn't really a need to call `Install-ChocolateyFosshub`

AdmiringWorm:
I dropped keeping Get-UrlFromFosshub public, it didn't wont work without changes to make it work without using chocolatey functions.

Added an example in the qbittorrent file to how the checksum would have to be generated.

majkinetor:
I will take a look today. 

We could switch qbittorrent to sourceforge. Audacity could be used as testing dummy for fosshub since it doesn't have official alternative download.

AdmiringWorm:
yeah, we could change qbittorrent to sourceforge although the official chocolatey guidelines mention to avoid sourceforge if another download mirror exists.

Another option we could do with packages that allow redistribution is do exactly that,
internalize those packages.
And to prevent the executables/archives to be checked in we could add a gitignore file in the automatic
folder ignore such files.

majkinetor:
Comparing to fosshub sourceforge is gold ATM.

> Another option we could do with packages that allow redistribution is do exactly that,
internalize those packages. And to prevent the executables/archives to be checked in we could add a gitignore file in the automatic folder ignore such files.

I like this idea A LOT. 

Mechanism for package to set specific plugin options would be great here - package could announce to Git plugin that it wants files ignored or not. For example `[AUPackage].Options` could contain HashTable with specific directions to plugins, that could be passed by update. For example:

    update -Options @{ Internalize = $true; GitIgnore = '*.exe' }

I am not sure if Internalize should go here or like this:

    update -Internalize -Options @{ GitIgnore = '*.exe' }

since its not plugin related.

This IMO doesn't compete with chocolatey PRO functionality of internalization as this one would work only in the context of update and only for some packages (GNU etc.) while corporative internalization can be done on anything.

@ferventcoder, please take a look, we already talked about it before few times but lets repeat some of the ideas and problems publically here.

AdmiringWorm:
> Comparing to fosshub sourceforge is gold ATM.

True, I think it is mostly in the guidelines because of the history of changed the installer or something (can't remember exactly).

> This IMO doesn't compete with chocolatey PRO functionality of internalization as this one would work only in the context of update and only for some packages (GNU etc.) while corporative internalization can be done on anything.

That would depend on exactly AU would be doing, but if giving an internalize option
only gets AU to download the file from specified urls and sets the file name of those files to `$Latest`,
then I don't think it would compete IMO.

This would mean the maintainer still would have to make the appropriate changes to the install script,
making this not a true internalization?
right?

majkinetor:
I updated the above answer.

majkinetor:
> This would mean the maintainer still would have to make the appropriate changes to the install script,
making this not a true internalization?
> right?

- Idea 1 would be done completelly by AU.
- Idea 2 wouldn't require AU changes.

majkinetor:
But yes, not true internalization.

ferventcoder:
Prefer the word "Embed" over "Internalize" - internalizing means it wasn't before and now I am. Embedding is more appropriate as I am simply embedding the software binaries in the package.

ferventcoder:
>>Another option we could do with packages that allow redistribution is do exactly that,
internalize those packages. And to prevent the executables/archives to be checked in we could add a gitignore file in the automatic folder ignore such files.


>  I like this idea A LOT.

I proposed it a couple of years ago as a way to have more reliability in the community packages. Other folks were not as interested. I think anything that can be embedded probably should be so we don't unknowingly upset other communities or hosts.

ferventcoder:
Let's not make any hosting service out to be a bad guy here. All of them have financial interests to protect, they can't provide hosting for free, so they need to be able to make money somehow. 

I just learned yesterday Fosshub is upset about Chocolatey packages that download from Fosshub because they believe "we" are doing so to to avoid bandwidth costs and instead have them pay the costs.

It makes me wonder what other hosting services or software vendors believe when they don't understand how the community repository works. And it's clear that some folks really do not understand software redistribution or the legal side of that.

majkinetor:
> I just learned yesterday Fosshub is upset about Chocolatey packages that download from Fosshub because they believe "we" are doing so to to avoid bandwidth costs and instead have them pay the costs.

Yeah, for others, here is the [link](https://www.reddit.com/r/sysadmin/comments/5g5npg/fosshub_message_for_chocolatey_please_stop_with/).

I wont comment this, I have better things to do, one of them being `update -Embed`.

ferventcoder:
https://www.reddit.com/r/sysadmin/comments/5g5npg/fosshub_message_for_chocolatey_please_stop_with/daqnamb/

This comment here is the path forward. With packages using FossHub, I'm going to review and contact maintainers. They will need to review licensing for distribution rights and move forward with embedding or seeking alternative actions. The final alternative is to remove those packages from the community site if there is no alternative.

AdmiringWorm:
I'm abondoning this extension, as we should work more against embedding software we can.

ferventcoder:
https://github.com/chocolatey/package-validator/issues/153

FYI - once implemented this will not allow a package to be accepted.

ferventcoder:
https://github.com/chocolatey/package-validator/wiki/ScriptsDoNotDownloadFromFossHub

PR #442, opened 2016-12-03
Updated qbittorrent to embed installer

AdmiringWorm:
qBittorrent uses GPL license which allows redistribution

I'm unsure about the creation of embedded packages, so any feedback would be appreciated.

/cc @gep13 @majkinetor @ferventcoder

### DIFF
--- a/automatic/qbittorrent/tools/VERIFICATION.txt
+++ b/automatic/qbittorrent/tools/VERIFICATION.txt
@@ -1,0 +1,17 @@
+VERIFICATION
+Verification is intended to assist the Chocolatey moderators and community
+in verifying that this package's contents are trustworthy.
+
+The installer have been downloaded from the alternative sourceforge mirror listed on <http://www.qbittorrent.org/download.php>
+and can be verified like this:
+
+1. Go to <http://sourceforge.net/projects/qbittorrent/files/qbittorrent-win32/qbittorrent-3.3.7/qbittorrent_3.3.7_setup.exe/download>
+   to download the installer
+2. You can use one of the following methods to obtain the checksum
+  - Use powershell function 'Get-Filehash'
+  - Use chocolatey utility 'checksum.exe'
+
+  checksum type: sha256
+  checksum: 49AE9A0ADFC3272BEC38822C528F732D9495B79A2A7CA934F8C6635237B15D07
+
+File 'LICENSE.txt' is obtained from <https://github.com/qbittorrent/qBittorrent/blob/0070dcf5509e43c2c3457c4e3f1ed0b1ae087e36/COPYING>
--- a/automatic/qbittorrent/update.ps1
+++ b/automatic/qbittorrent/update.ps1
@@ -1,28 +1,44 @@
 import-module au
-import-module "./../../extensions/extensions.psm1"
 
-$releases = 'https://www.fosshub.com/qBittorrent.html'
+$releases = 'http://www.qbittorrent.org/download.php'
+
+function global:au_BeforeUpdate {
+  Remove-Item "$PSScriptRoot\tools\*.exe"
+
+  $client = New-Object System.Net.WebClient
+
+  $filePath = "$PSScriptRoot\tools\$($Latest.FileName)"
+
+  $client.DownloadFile($Latest.URL32, "$filePath")
+  $client.Dispose()
+
+  $Latest.ChecksumType = 'sha256'
+  $Latest.Checksum = Get-FileHash -Algorithm $Latest.ChecksumType -Path $filePath | % Hash
+}
 
 function global:au_SearchReplace {
    @{
         ".\tools\chocolateyInstall.ps1" = @{
-            "(^[$]fosshubUrl\s*=\s*)('.*')"   = "`$1'$($Latest.URL32)'"
-            "(^[$]checksum\s*=\s*)('.*')"     = "`$1'$($Latest.Checksum32)'"
-            "(^[$]checksumType\s*=\s*)('.*')" = "`$1'$($Latest.ChecksumType32)'"
+            "(?i)(`"`[$]toolsDir\\).*`"" = "`${1}$($Latest.FileName)`""
+        }
+        ".\tools\verification.txt"      = @{
+          "(?i)(1\..+)\<.*\>"          = "`${1}<$($Latest.URL32)>"
+          "(?i)(checksum type:\s+).*"   = "`${1}$($Latest.ChecksumType)"
+          "(?i)(checksum:\s+).*"       = "`${1}$($Latest.Checksum)"
         }
     }
 }
 
 function global:au_GetLatest {
-    $download_page = Invoke-WebRequest -Uri $releases
+    $download_page = Invoke-WebRequest -Uri $releases -UseBasicParsing
 
-    $re    = '\.exe$'
-    $urlPart   = $download_page.links | ? href -match $re | select -First 1 -expand href
-    $url = "https://www.fosshub.com$urlPart"
+    $re    = 'setup\.exe'
+    $url   = $download_page.links | ? href -match $re | select -First 1 -expand href
 
     $version  = $url -split '[_]' | select -Last 1 -Skip 1
+    $fileName = $url -split '/' | select -last 1 -skip 1
 
-    return @{ URL32 = $url; Version = $version }
+    return @{ URL32 = $url; Version = $version; FileName = $fileName }
 }
 
-update -ChecksumFor 32 -NoCheckUrl
+update -ChecksumFor none
--- a/automatic/.gitignore
+++ b/automatic/.gitignore
@@ -1,0 +1,9 @@
+# Ignores binary files that is necessary to be stored in the repository
+*.exe
+*.msi
+*.dll
+*.zip
+*.7z
+*.rar
+
+# And probably many more
--- a/automatic/qbittorrent/qbittorrent.nuspec
+++ b/automatic/qbittorrent/qbittorrent.nuspec
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks,
+you should use an editor that supports UTF-8, not this one. -->
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>qbittorrent</id>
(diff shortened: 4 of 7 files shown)


gep13:
Have a look at the ChocolateyGUI package. It is quite old, and needs updating, but the premise of embedding the msi within the package is shown in there.

gep13:
Heading out for the night, but will check in tomorrow to see if there are any questions.

AdmiringWorm:
gotcha.
From looking at the chocolateyGUI package, this should work then (with that I mean it should be accepted).
I was mostly curious if there was something additional I needed to include, the file is downloaded by au
in the BeforeUpdate function and replaces the path/filename specified as arguments to `Install-ChocolateyInstallPackage`.

gep13:
Nope, don't think so. As long as the nuspec files section isn't explicitly including files, then the downloaded file should get picked up and included in the package that is generated by AU. If we are going down this route, we will need to include a verification.txt to allow moderators to verify that the included file is legitimate.

AdmiringWorm:
> If we are going down this route, we will need to include a verification.txt to allow moderators to verify that the included file is legitimate.

Like where they can download the file and verify the checksum?

gep13:
Yip. Check the output of `Choco new` to see what is expected. It is mainly used by moderators, but can be used by end users as well to additionally verify the included installer.

automatic/qbittorrent/update.ps1
```diff
@@ -1,28 +1,36 @@
 import-module au
 import-module "./../../extensions/extensions.psm1"
```
majkinetor: should be removed

automatic/qbittorrent/update.ps1
```diff
@@ -1,28 +1,36 @@
...
 
-$releases = 'https://www.fosshub.com/qBittorrent.html'
+$releases = 'http://www.qbittorrent.org/download.php'
+
+function global:au_BeforeUpdate {
+  $download_page = Invoke-WebRequest -uri $Latest.URL32 -UseBasicParsing
+  $url = $download_page.links | ? class -eq "direct-download" | select -expand href
+
+  $client = New-Object System.Net.WebClient
```
majkinetor: Why not using iwr for this too ?

majkinetor:
LGTM

> Like where they can download the file and verify the checksum?

You can just put $Latest table or parts of it in verification.txt

Some questions:

- Do we keep the setup in the tools dir forever or not ? 
- Will the tools dir build up old installers ? 
- Windows installer dir is already cacheing it too. Could we delete it after installation to save space ?

automatic/qbittorrent/update.ps1
AdmiringWorm: because iwr is notoriously slow for downloading files, so to save time if there is multiple packages
needing to download files, I'm using .net instead.

gep13:
>@majkinetor said...
You can just put $Latest table or parts of it in verification.txt

Not sure what you mean by this. Can you elaborate?

The idea of the verification.txt is that it is a list of human followable steps to verify that the bundled installer, and associates checksum are correct. I.e. Go to this website, click on download link, run this command to generate the hash, verify against chocolateyInstall.ps1.

AdmiringWorm:
> @majkinetor said
Do we keep the setup in the tools dir forever or not ?

I don't see any reason to do that, and since the .exe file is already is already on the ignore list for the repo it shouldn't be commited (unless AU overrides it though).

> @majkinetor said
Will the tools dir build up old installers ?

Good question, I don't really know. definitely something that needs to be tested though.

> @gep13 said
verify against chocolateyInstall.ps1.

I'm guessing you meant verifying against the installer/archive, right?

gep13:
>@AdmiringWorm said...
 guessing you meant verifying against the installer/archive, right?

Well yes, both. I.e. Verify against the checksum that is contained within the installation script as well.

AdmiringWorm:
@gep13 It would be enough with something like this right?

```
The installer have been downloaded from the alternative sourceforge
mirror listed on <http://www.qbittorrent.org/download.php>
and can be verified like this:

1. Go to <download location> and download the installer
2. Use powershell function 'Get-FileHash' or the chocolatey utility 'checksum.exe'
   to verify the checksum matches 'checksum hash'
```

the checksum in the installation script? isn't the checksum only in the installation script if
we download it from the during installation, not when it has been embedded.
Or am I misunderstanding something perhaps?

gep13:
@AdmiringWorm ah, sorry, wasn't thinking there.

Yes, you are right. The expected checksum would go into the verification.txt file, and wouldn't be in the install script.

My bad...

automatic/qbittorrent/tools/VERIFICATION.txt
```diff
@@ -0,0 +1,11 @@
...
+Verification is intended to assist the Chocolatey moderators and community
+in verifying that this package's contents are trustworthy.
+
+The installer have been downloaded from the alternative sourceforge mirror listed on <http://www.qbittorrent.org/download.php>
+and can be verified like this:
+
+1. Go to <http://sourceforge.net/projects/qbittorrent/files/qbittorrent-win32/qbittorrent-3.3.7/qbittorrent_3.3.7_setup.exe/download>
+   and download the installer
+2. Use powershell function 'Get-FileHash' or the chocolatey utility 'checksum.exe'
+   to verify that the installer matches the sha256 checksum '49AE9A0ADFC3272BEC38822C528F732D9495B79A2A7CA934F8C6635237B15D07'
```
ferventcoder: BEAUTIFUL! Thanks for catching this as a necessity

ferventcoder:
You will also need a LICENSE.txt file. The reason we ask for this is protection of maintainers and the community. A remote licenseUrl cannot be controlled and could be changed. You can see how this could lead to problems down the road if someone changes the license and then claims that the license did not allow for distributions.

For protection, also place the LICENSE.txt file in the package with the embedded software.

AdmiringWorm:
@ferventcoder alright, will do

ferventcoder:
> If we are going down this route, we will need to include a verification.txt to allow moderators to verify that the included file is legitimate.

This is not just for moderators. This is also so someone looking can take the checksum for the binary (listed on the package page under files) and use it to verify the remote location is exactly the same.

FWIW - when you push a package, during the gathering of all included files, if the file is a binary, the Gallery grabs the checksum values for md5, sha1, sha256, and sha512 to ease verifying it with the remote source.

ferventcoder:
I'd still add the value to the verification.txt file as well, that adds value IMHO.

ferventcoder:
I wish github would publish checksums when folks put up binaries under releases. That would be most helpful.

AdmiringWorm:
> @ferventcoder said
 I wish github would publish checksums when folks put up binaries under releases. That would be most helpful.

Same here, I actually find it a little odd that they doesn't

automatic/qbittorrent/tools/VERIFICATION.txt
AdmiringWorm: no problem, is there anything else should be in the verification.txt file?
or any additional steps I need to include?

majkinetor:
> Do we keep the setup in the tools dir forever or not ?

I didn't mean about repo, but about installation. Installer can delete the setup after installation for example to save space, otherwise, every one will be x3 (tools, TEMP, windows cache).

> automatic/qbittorrent/tools/VERIFICATION.txt

We could have a template for this, could be part of the package directory and AU can modify it during update. Something like:

```
+1. Go to $releases page and download the installer
+2. Use powershell function 'Get-FileHash' or the chocolatey utility 'checksum.exe'
+   to verify that the installer matches the sha256 checksum is $Latest.Chekcusm
```

Then this could be used as such for majority of packages.

majkinetor:
About keeping the binaries in repository I wouldn't do that for start, but that is probably the best solution. But really, if Choco ever goes out of buisnis I guess there will be migration plan and it will certainly not be shut down over night so we can make crawlers to downlolad older packages that we need.

And honestly, in huge majority of cases you don't really need particular version but latest one so older versions are most of the time IMO irrelelvant (i mean, its one thing to have version for library and who needs specific version of lets say qbittorent).

majkinetor:
You will have to take into account that `tools` folder should be cleaned before eacn new install. On AV its already empty, but local runs may keep outdated files.

AdmiringWorm:
Sorry, my mind went straight to the repo not on the users machine.

regarding saving space, it wouldn't be stored in TEMP though but it will be stored in tools directory and possibly in windows cache (although my windows\installer directory only contains msi files, not .exe files).

> We could have a template for this, could be part of the package directory and AU can modify it during update.

That would definitely be useful.

> About keeping the binaries in repository I wouldn't do that for start, but that is probably the best solution. 

There is no need to store them in the repository, with this PR I've added a .gitignore file in the automatic directory to ignore common binary files.
They are packed inside the nupkg package, so there is just no need to store them anywhere else. IMHO

AdmiringWorm:
> You will have to take into account that tools folder should be cleaned before eacn new install. On AV its already empty, but local runs may keep outdated files.

That may be true, I haven't really tested updating yet, but I will.

majkinetor:
> regarding saving space, it wouldn't be stored in TEMP though

Right, forgot you are using different choco installer. 

> They are packed inside the nupkg package, so there is just no need to store them anywhere else. IMHO

I know, the point was that then we depend on chocolatey.org existence. If chocolatey.org is out one day packages on this repo become unusable.

majkinetor:
But its certainly better then what we have now where exes dissapear from vendor site.

majkinetor:
I think chocolateyUninstall.ps1 should be modernized.

AdmiringWorm:
> I think chocolateyUninstall.ps1 should be modernized.

Using the uninstall helper in `chocolatey-core` extension instead?

ferventcoder:
We could explore git-lfs for binaries.

ferventcoder:
Then you get best of both worlds here

AdmiringWorm:
> We could explore git-lfs for binaries.

That's definitely a possibility, can't say I've ever tried git-lfs before though so I don't know how it is used (YET).
I also think lfs storage is free for the first GB, but not entirely sure about that though.

AdmiringWorm:
I think it is done now, only thing to find out first it whether we should store the binaries in the repo or not.
If we are I need to remove the additional .gitignore file I've added.

majkinetor:
> I think it is done now, only thing to find out first it whether we should store the binaries in the repo or not.

Not for now. I just tried git-lfs and its really epic but we need to either use Github enterprise for chocolatey user or mount lfs server on existing chocolatey infrastructure. Without deduplication size will not be trivial - 15 versions of smplayer alone for example will take ~1GB. I think that 1 TB of space without deduplication should be enough for a couple of years for all packages that could ever become embeded. With dedup, I beleive this could be order of magnitude lower.

Other then that I think git-lfs is epic as given the chocolatey client, repo becomes another form of decentralized gallery that is easy to utilize independent of the chocolatey.org and it solves the 404 problem completelly.

majkinetor:
So far I see just one downside of embeding - you will typically download x2 of what you need (as package contains both x32 and x64 and non-embeded packages download just one). This is not ideal for big tools but again those should probably not be embeded. Alternative is to use separate x32 and x64 package (like we already have for some packages) which is something I dont think we should generally practice.

majkinetor:
> I think it is done now

Qbittorrent does not to register itself.  Could you add at the end:

 ```  
$packageName = $packageArgs.packageName
$installLocation = Get-AppInstallLocation $packageName
if ($installLocation)  {
    Write-Host "$packageName installed to '$installLocation'"
    Register-Application "$installLocation\$packageName.exe" qbit
    Write-Host "$packageName registered as qbit"
}
else { Write-Warning "Can't find $PackageName install location" }
```

The details are on https://github.com/chocolatey/choco/issues/1072. This is quite practical for testing too as it allows you to very quickly launch it.

Other then that this is ready for push.

AdmiringWorm:
@majkinetor I've added the applicationr registration and uninstall of that key.

If it looks good, I'll squash and merge this now...

majkinetor:
LGTM

PR #448, opened 2016-12-05
(nomacs*) Guidelines (fixes #395)

### LINKED ISSUE #395: (nomacs) How should software get updated
Perhaps switch to using Fosshub, as suggested in disqus comment:

https://chocolatey.org/packages/nomacs#comment-2997181741

majkinetor:
This is ready except for proxy (see https://github.com/chocolatey/chocolatey-coreteampackages/issues/395#issuecomment-264851727). 

Waiting for feedback from authors on how to set it up since it uses online installer.

### DIFF
--- a/automatic/nomacs.portable/nomacs.portable.nuspec
+++ b/automatic/nomacs.portable/nomacs.portable.nuspec
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    <id>nomacs.portable</id>
+    <version>3.4.0.20161205</version>
+    <title>nomacs Image Lounge</title>
+    <owners>chocolatey</owners>
+    <authors>Markus Diem, Stefan Fiel, Florian Kleber</authors>
+    <licenseUrl>https://www.gnu.org/licenses/gpl.html</licenseUrl>
+    <projectUrl>https://nomacs.org/</projectUrl>
+    <iconUrl>https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/bcd6e5bad67c9a360b1e9bc9a1ccfedfd4e57fdd/icons/nomacs.svg</iconUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>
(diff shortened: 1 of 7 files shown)


majkinetor:
I think that this should be pushed now even tho proxy does not work. I added a note for the time being for the online installer package and will wait up for the answer from the vendor. 

The reasons are

- current package doesn't work due to checksum
- it never worked with proxy anyway
- portable package doesn't need proxy and works

If there is no answer for the proxy from authors I can make AHK script for online installer that sets it.

@gep13 @AdmiringWorm

majkinetor:
@gep13 I have a problem to push nomacs as it currently occupies [3.4.1](https://chocolatey.org/packages/nomacs) version and current remove one is 3.4.0.  What do you suggest ? Keep in mind that I would like both nomacs & nomacs.portable which currently doesn't exist to have the same version.

Maybe we should unlist 3.4.1 and I could push 3.4.0.fix for both ?

gep13:
@majkinetor yes, that sounds sensible to me.  I would unlist both the 3.4 and the 3.4.1 packages, and push 3.4.0.

majkinetor:
Sounds great.

gep13:
@majkinetor are you working on this just now?  You want them unlisted just now?

majkinetor:
Yup

majkinetor:
You can see by the labels - its done and waiting for you to finish.

automatic/nomacs.portable/nomacs.portable.nuspec
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    <id>nomacs.portable</id>
+    <version>3.4.0.20161205</version>
```
gep13: @majkinetor is the package fix notation required here?  I thought the plan was to try to push 3.4.0?

gep13:
@majkinetor I have unlisted 3.4 and 3.4.1 on chocolatey.org

automatic/nomacs.portable/nomacs.portable.nuspec
gep13: Gotcha, no worries.

PR #449, opened 2016-12-05
(tightvnc) Removed ketarin file

gep13:
@majkinetor think this was missed when you migrated the tightvnc package.

Or was it left for a reason?


AdmiringWorm:
It was probably just missed, no reason to keep it anyways. will still be available in the history

majkinetor:
No reason, its even stranger since I scan all ketarin files from time to time to see what we missed.

PR #450, opened 2016-12-05
(codeblocks) Updated to embedd installer

AdmiringWorm:


majkinetor:
U r missing import to core. OR we na ditch that mechanism and preinstall it on AV. It will not be the same as package can use unpublished stuff this way but that rarelly happens and that package can still manually import.

AdmiringWorm:
Since this one doesn't use au to download the the file, a reference to core extension shouldn't be needed.

majkinetor:
Oh, yes, forgot about it, sorry.

AdmiringWorm:
@majkinetor 
Is there anything else?

majkinetor:
[Looks clear](https://www.youtube.com/watch?v=YVFgI_F4R7Y)

PR #451, opened 2016-12-06
(autohotkey*) embedded 

majkinetor:


majkinetor:
Will wait for new AHK version for this one to get pushed because of the versioning problem.

PR #452, opened 2016-12-06
Create chocolateyUninstall.ps1

sippi90:
since silent uninstall isn't working for notepadplusplus.
I've created a chocolateyUninstall.ps1


majkinetor:
Thanks sippy !

majkinetor:
This actually requires some changes. If the package is uninstalled  by other means it will fail.

sippi90:
perfectly done

Issue #454, opened 2016-12-07
Overlapping functions in chocolatey-core.extension and mm-choco.extension need more thought


ferventcoder:
https://chocolatey.org/packages/virtualbox#comment-3040800131

If there are, you need to ensure that chocolatey-core.extension wins.

ferventcoder:
@majkinetor @AdmiringWorm FYI

majkinetor:
mm-choco.extension is now completelly contained in chocolatey-core.extension and I will deprecate it ASAP.

ferventcoder:
Cool, so the answer is yes, you do need to account for the fact that folks may have that extension installed.

ferventcoder:
mm-choco is currently after chocolatey-core in the alphabet, so all functions in the other package win and override the same ones in chocolatey-core. 

Typically you would set an alias for the ones you want to win.

ferventcoder:
> Typically you would set an alias for the ones you want to win.

This is not meant to always be used, as it is a heavy handed approach. It should only be used in cases where this could occur.

majkinetor:
https://chocolatey.org/packages/mm-choco.extension/0.0.4.1

I am not sure if people will have to explicitelly uninstall it now

ferventcoder:
The problem was the versionless or minimum version installed dependencies from other packages. Anyone coming along now or upgrading all will be put in a good place.

This is one of the reasons I drug my feet in ever documenting extensions, because they do require a lot of upfront thought sometimes in planning whether a function should be loaded or not.

ferventcoder:
Like checking to see if a command is already available and only loading the function is not already loaded by other means (an automatic deprecation to functions that are later built into newer Chocolatey versions).

majkinetor:
core extension actually does this. It wont load if its already available.

ferventcoder:
It's trickier when you move functions from one extension into another. You have to consider all scenarios and try to provide the best experience to lower a user's issues. I tend to prefer having no hard to determine issues. Like this one. It may not be obvious at first (it certainly wasn't to me until I thought about it).

ferventcoder:
I just looked through the code in core extension and I don't see this at all. Can you point me to the code that does something like this:

https://github.com/ferventcoder/chocolatey-packages/blob/f22175b28804dfa9bddbf07edadf08a77c6d7b78/manual/pik/tools/chocolateyInstall.ps1#L7-L10

ferventcoder:
[...]

PR #456, opened 2016-12-07
(chocolatey-core.extension) Guard loading Register-Application

ferventcoder:
When Register-Application is already loaded by other means, write a 
debug message for debugging purposes but don't load the extension.

Relates to #454 

This is an example of what I'm looking for with extension function loading.


majkinetor:
Is there any special reason you did just for just one function ?

gep13:
>@majkinetor said...
Is there any special reason you did just for just one function ?

I suspect that this was just done as an example.

ferventcoder:
Just an example, but the code in the PSM1 already does this.

Issue #443, opened 2016-12-04
(everything) Migrate from user


majkinetor:
This is best tool in the world for searching and it belongs here. It requires modernization and checksum ATM. License allows redistribution so it should be embeded, especially given the very low size. 

I contacted maintainer to give ownership to chocolatey user. 

https://chocolatey.org/packages/Everything

majkinetor:
@gep13 , user responded but doesn't know how, could you do it ? 

> Hi,
> 
> For some reason, I am not seeing this discussion in the Everything comments so I am emailing you directly.  I would be glad to add "Chocolatey" as a user to the repo for this package but I am not seeing "Chocolatey" as a Github username.  Can you please verify the Github ID.
>

gep13:
@majkinetor sounds like they are trying to add a "chocolatey" user as an owner to the GitHub Repo where the source files for the Chocolatey Package resides.  I am assuming that you are talking about adding the chocolatey user as a co-maintainer on chocolatey.org, and then your intention is to move the Everything package over to the Core Team Repo, right?

majkinetor:
Yeah.

BTW, this conversation took place several days back.

gep13:
As long as you email him back to confirm exactly what is being done, I will go ahead and add the user as a co-maintainer.  A PR to his repo to remove the source files would also make sense, assuming you can find his repo, as it is not immediately obvious where that is.

majkinetor:
> Date: Wed, 7 Dec 2016 12:27:59 +0100
> Message-ID: <CAB6nfs4GNWGNhin-sp2sPFcbjZCSDqV-oK-zGnJ92TmFx-mU=A@mail.gmail.com>
> Subject: Re: [Chocolatey.org] Message for maintainers of the package 'Everything'
> From: Miodrag Milic <***************>
> To: Mark Michaelis <*****************>
> Content-Type: multipart/alternative; boundary=047d7bfcfb9226d16b05430fd0a6
> 
> 
> Choco moderators added chocolatey user as maintainer.
> Please remove the package from your automatic routines.

PR #459, opened 2016-12-08
(everything) new package

majkinetor:
Check it out (#443).

Notice the [Convert-ToEmbeded](https://github.com/chocolatey/chocolatey-coreteampackages/blob/9dc73656c3384c017e20957a3a3ced12dda8c71b/automatic/everything/update.ps1#L21-L48) function and how it is used. Perhaps somebody else will find it useful and if so, I may add it to AU.

### DIFF
--- a/automatic/everything/Everything.nuspec
+++ b/automatic/everything/Everything.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>everything</id>
-    <version>1.4.1809.20161208</version>
+    <version>0.0</version>
     <title>Everything</title>
     <authors>David Carpenter</authors>
     <owners>chocolatey, Mark Michaelis</owners>
@@ -36,6 +36,8 @@ You type in a search filter to limit what files and folders are displayed.
     <dependencies>
       <dependency id="chocolatey-core.extension" version="1.0.2" />
     </dependencies>
-
   </metadata>
+  <files>
+    <file src="tools\**" target="tools" />
+  </files>
 </package>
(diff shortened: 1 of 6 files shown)


automatic/everything/Everything.nuspec
```diff
@@ -0,0 +1,41 @@
...
+    <docsUrl>https://www.voidtools.com/faq/</docsUrl>
+    <mailingListUrl>https://www.voidtools.com/forum/</mailingListUrl>
+    <bugTrackerUrl>https://www.voidtools.com/forum/viewforum.php?f=6</bugTrackerUrl>
+    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/everything</packageSourceUrl>
+    <dependencies>
+      <dependency id="chocolatey-core.extension" version="1.0.2" />
+    </dependencies>
+
+  </metadata>
+</package>
```
AdmiringWorm: looks like you're missing the files element

AdmiringWorm:
btw, I found IMO a simpler way to download the files that could perhaps be used.
I tested out in my own repository reusing chocolatey helpers (namely `Get-WebFileName` and `Get-WebFile`), and found this to be quite simple.

I'm currently using that method for the InnoSetup package as can be seen here:
https://github.com/AdmiringWorm/chocolatey-packages/blob/master/automatic/InnoSetup/update.ps1

automatic/everything/Everything.nuspec
majkinetor: looks like it isn't needed ...

AdmiringWorm: if you don't have the files element, it will also pack the `update.ps1` file

PR #470, opened 2016-12-12
(git-lfs) Migrate git-lfs packages

pascalberger:
This PR is to move the git-lfs packages from https://github.com/pascalberger/chocolatey-packages to the Core Team Packages.

I invited the Chocolatey user as a maintainer.


majkinetor:
Would u embed it ?

pascalberger:
@majkinetor Sorry don't understand what you mean. Could you please explain

majkinetor:
Please see [1.1.4 - Embed a package if allowed](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#114-embed-a-package-if-allowed).

pascalberger:
@majkinetor Do you mean the installer embedded in the git-lfs.install package or an additional package?

majkinetor:
Installer embedded in the git-lfs.install package.

pascalberger:
@majkinetor OK, I see the point of having an embedded package. I'll contact GitHub about getting the distribution rights. Is there a reason why this didn't came up for the Visual Studio Code or SourceTree package? Wouldn't it also make sense to make them embedded packages too?

majkinetor:
>  I'll contact GitHub about getting the distribution rights. 

There is no need for that. Its MIT license, just include it. 

> Is there a reason why this didn't came up for the Visual Studio Code or SourceTree package? 

Yes, SourceTree doesn't allow redistribution. VSCode seems to be not redistributtalbe too. For those you would need to ask vendor.

majkinetor:
To be honest it would be good to have VSCode embedded here, its probable that they will give redistribution rights for binaries (source code is license MIT AFAIK).

BTW, it VSCode currently fails updater on [force test](https://gist.github.com/gep13/b733b29e320a8c05bc99cc0bf59cd8d1#visualstudiocode) because it must import extension in updater. If it is embedded, this is no longer required.

pascalberger:
@majkinetor I updated it to be an embedded package

majkinetor:
Looks good.

Just set the nuspec version of virtual package to be the same as for install package. I will force update for both.

pascalberger:
@majkinetor Updated the `git-lfs.nuspec`

majkinetor:
Failed now https://gist.github.com/bd2eaa76f2a9ab739ca0544c502dca6e/91468e81cfffeccfbf018901fa089f013701a3e3

pascalberger:
@majkinetor @gep13 You need to accept the invite as package moderator. I'll have a look at the other error.

gep13:
Chocolatey user has been added to the package.

pascalberger:
@majkinetor Any idea what the problem with the git-lfs is in the force tests? As it seems to work actually since they were updated and also works for me locally.

majkinetor:
Definitelly network related, should be ignorred. Both packages are now pusshed.

PR #475, opened 2016-12-15
(activepresenter) embedded package [AU activepresenter]

majkinetor:


automatic/activepresenter/tools/chocolateyInstall.ps1
```diff
@@ -1,13 +1,23 @@
 $ErrorActionPreference = 'Stop'
 
+$fileType      = 'exe'
+$toolsDir      = Split-Path $MyInvocation.MyCommand.Definition
+$embedded_path = gi "$toolsDir\*.$fileType"
```
AdmiringWorm: why use a globbing pattern? why not pass the filename from the update script?

AdmiringWorm: fair enough.

majkinetor: Now the only thing that troubled me is that it it changes installer file name. This was never problematic but if you actually donwload software which includes x64 and x32 this will add just one of those to the end which might be confusing to somebody looking into the package after unpacking it - one may think its x32 or x64 while its actualy both.

I decided so far that this issue does not deserves attention.

majkinetor:
Its not blank... its scrambled. I don't know how it happened. I used cleopatra to sign it up with my pgp key but it was in separate file which I later decided not to push. .

AdmiringWorm:
try adding `*.pdf binary` to the .gitignore file and resubmit the pdf file.
Could be it tried to convert it somehow instead of treating it as a binary file.

majkinetor:
Yeah, git does it, it did it again.

majkinetor:
No change.

majkinetor:
[Redistribution.pdf](https://github.com/chocolatey/chocolatey-coreteampackages/files/656885/Redistribution.pdf)

I uploaded file here for others to try.

AdmiringWorm:
I mentioned the wrong file, i meant .gitattributes.
Hold on.

majkinetor:
It was suspicious that it is in gitignore.

AdmiringWorm:
now it works

AdmiringWorm:
one question more though.
[AU activepresenter:5.3.1.22] is wrong right? the latest version is 6.03

majkinetor:
Yeah, it was absolute uninstaller.

PR #483, opened 2016-12-18
(mkdocs) Check PyPi for new package releases instead of GitHub

pascalberger:
Not all releases from GitHub are necessarily pushed to the Python Package Index from where the Chocolatey Package will install the tool.


AdmiringWorm:
had to change this package and the mkdocs-material package to parse the web page itself instead of the json result.
It failed almost every time when parsing the json.

Commit: https://github.com/chocolatey/chocolatey-coreteampackages/commit/280757ddc20b03eb8a36a849dc5f138f3722d57c

PR #489, opened 2016-12-21
(git*) migrated from ferventcoder

pascalberger:
* Package source: https://github.com/ferventcoder/chocolatey-packages/pull/224
* [6 open issues](https://github.com/ferventcoder/chocolatey-packages/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20git)

## Changes
Compared to the version from ferventcoder I changed the following stuff:
* All packages:
    * Added AU script
    * Update NuSpec metadata to be compliant with core team guidelines
    * Change to embedded package
* git.install
    * Use Core.Extension for parameter parsing
* git.portable
    * Changed ID of `git.commandline` package to `git.portable`
    * Add to PATH
    * Add `chocolateyUninstall.ps1` script
    * Deprecation of the existing `git.commandline` package is handled in https://github.com/ferventcoder/chocolatey-packages/pull/225


majkinetor:
That was next on my list. Thanks for making an effort.

majkinetor:
@ferventcoder, should we create `.portable` packages now and deprecate `.commandline` ?

ferventcoder:
@majkinetor I think that would be fine.

pascalberger:
@ferventcoder @majkinetor @gep13 Can you please review this. There were quite some changes in the scripts and metadata. See PR description for an overview of the changes compared to the original version from ferventcoders repo.

automatic/git.install/git.install.nuspec
```diff
@@ -0,0 +1,51 @@
...
+## Features
+
+* **Git BASH**: Git for Windows provides a BASH emulation used to run Git from the command line. *NIX users should feel right at home, as the BASH emulation behaves just like the "git" command in LINUX and UNIX environments.
+* **Git GUI**: As Windows users commonly expect graphical user interfaces, Git for Windows also provides the Git GUI, a powerful alternative to Git BASH, offering a graphical version of just about every Git command line function, as well as comprehensive visual diff tools.
+* **Shell Integration**: Simply right-click on a folder in Windows Explorer to access the BASH or GUI.
+
+## Package parameters
+The following package parameters can be set:
+
+* `/GitOnlyOnPath` - this puts gitinstall\cmd on path. This is also done by default if no package parameters are set.
```
majkinetor: Remove `/` prefix on parameters description.

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,107 @@
...
+$filePath32 = "$toolsDir\Git-2.11.0-32-bit.exe"
+$filePath64 = "$toolsDir\Git-2.11.0-64-bit.exe"
+$fileArgs = $(
+  '/VERYSILENT /NORESTART /NOCANCEL /SP- ' +
+  '/COMPONENTS="icons,icons\quicklaunch,ext,ext\shellhere,ext\guihere,assoc,assoc_sh" /LOG'
+)
+
+# Parse package parameters
+$pp = Get-PackageParameters
+$gitCmdOnly = $pp.GitOnlyOnPath
```
majkinetor: Why do you have those variables ? Just use $pp.XXX bellow

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,107 @@
...
+$unixTools = $pp.GitAndUnixToolsOnPath
+$useWindowsTerminal = $pp.WindowsTerminal
+$noAutoCrlf = $pp.NoAutoCrlf
+
+$registryKeyName = 'Git_is1'
+$is64bit = (Get-ProcessorBits) -eq 64
```
majkinetor: Get-ProcessorBits 64

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,107 @@
...
+$installKey = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
+if ($is64bit -eq $true -and $env:chocolateyForceX86 -eq $true) {
```
majkinetor: `$is64bit -and $env:chocolateyForceX86 -eq 'true'`

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,107 @@
...
+$userInstallKey = "HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
+if (Test-Path $userInstallKey) {
+  $installKey = $userInstallKey
+}
+
+if ( -not (Test-Path $installKey)) {
+  New-Item -Path $installKey | Out-Null
+}
+
+if ($gitCmdOnly) {
```
majkinetor: I would move entire next `if (...) NewItemProperty` section into helpers.ps1.

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,107 @@
...
+  SilentArgs     = $fileArgs
+  ValidExitCodes = @(0)
+}
+
+Install-ChocolateyInstallPackage @packageArgs
+
+if (Test-Path $installKey) {
+  $keyNames = Get-ItemProperty -Path $installKey
+
+  if ($gitCmdOnly -eq $false -and $unixTools -eq $false) {
```
majkinetor: `if (! ($gitCmdOnly -or $unixTools) )`

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,107 @@
...
+      Install-ChocolateyPath $gitPath 'Machine'
+    }
+  }
+}
+
+# Lets remove the installer as there is no more need for it
+Remove-Item -Force $filePath32
+Remove-Item -Force $filePath64
+
+Write-Warning "Git installed - You may need to close and reopen your shell for PATH changes to take effect."
```
majkinetor: Redudant message, choco already writes that if needed.

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,32 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$filePath32 = "$toolsDir\PortableGit-2.11.0-32-bit.7z.exe"
+$filePath64 = "$toolsDir\PortableGit-2.11.0-64-bit.7z.exe"
+$filePath = if (Get-ProcessorBits -eq 64) { $filePath64 } else { $filePath32 }
```
majkinetor: Not enough, must check for forcex86 too.

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,32 @@
...
+$installLocation = Join-Path $(Get-ToolsLocation) 'git'
+
+Get-ChocolateyUnzip -FileFullPath $filePath -Destination $installLocation
+
+# Add to path
+Write-Host "Add Git to system PATH"
```
majkinetor: Redudant message

automatic/git.portable/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,11 @@
+$ErrorActionPreference = 'Stop'
+
+$installLocation = Join-Path $(Get-ToolsLocation) 'git'
```
majkinetor: This wont do. It might not be on toolslocation.

pascalberger: @majkinetor What is considered best practices here? Honstely I just ported this from the current implementation (just replaced the deprecated function with `Get-ToolsLocation`)

automatic/git.install/git.install.nuspec
pascalberger: Done

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done (also on some other places)

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.portable/tools/chocolateyInstall.ps1
pascalberger: Done. Also added the same logic for the git.install package: https://github.com/chocolatey/chocolatey-coreteampackages/pull/489/files#diff-38b497b16f7f35551e33f1702ceffb25R69

automatic/git.portable/tools/chocolateyInstall.ps1
pascalberger: Done

majkinetor:
> Changed ID of git.commandline package to git.portable

This must be done differently. `git.commandline` must be deprecated and to point to `git.portable`.

majkinetor:
Ignore above, I didn't see that it is deprecated @ ferventcoder repo.

automatic/git.portable/tools/chocolateyUninstall.ps1
majkinetor: I do sometning like:

```powershell
$pp = Get-PackageParameters
$installDir = Get-ToolsLocation
if ($pp.InstallDir ) { $InstallDir = $pp.InstallDir}
Write-Host "$packageName is going to be installed in '$installDir'"
```

Using `Get-ToolsLocation` requires

```xml
    <dependencies>
      <dependency id="chocolatey" version="0.9.10" />
    </dependencies>
```

However, this becomes releveant **ONLY when choco#797 is implemented**. Until then people should not use this parameter as it will not be remembered when updating.

Now the PATH becomes problematic as if user changes the dir you don't know where is it installed later unless you remember it somewhere. You would have to remember the argument somewhere, such as in tools directory, and read it on uninstall.

So, I wouldn't do that now. `Get-ToolsLocation` is OK (just add dependency) and once 797 is up we can fix this and other stuff by adding `InstallDir`.

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,96 @@
...
+  '/COMPONENTS="icons,icons\quicklaunch,ext,ext\shellhere,ext\guihere,assoc,assoc_sh" /LOG'
+)
+
+# Parse package parameters
+$pp = Get-PackageParameters
+
+$registryKeyName = 'Git_is1'
+$is64bit = Get-ProcessorBits 64
+$installKey = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
+if ($is64bit -and $env:chocolateyForceX86 -eq $true) {
```
majkinetor: `$env:chocolateyForceX86 -eq 'true'`

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,96 @@
...
+$filePath64 = "$toolsDir\Git-2.11.0-64-bit.exe"
+$fileArgs = $(
+  '/VERYSILENT /NORESTART /NOCANCEL /SP- ' +
+  '/COMPONENTS="icons,icons\quicklaunch,ext,ext\shellhere,ext\guihere,assoc,assoc_sh" /LOG'
+)
+
+# Parse package parameters
+$pp = Get-PackageParameters
+
+$registryKeyName = 'Git_is1'
```
majkinetor: https://github.com/pascalberger/chocolatey-coreteampackages/blob/ff38b7dd3ed75ba574218cc41ab5a823901f690f/automatic/git.install/tools/chocolateyInstall.ps1#L14-L48

I would move this to `helper.ps1`. It defocuses from the task in question with verbosity and details. For example, you might call it like this `Set-GitExtraPararameters $pp` because as far as I can see those are not supported via silent options. Or maybe `Set-GitRegistryOptions $pp`.

Also

https://github.com/pascalberger/chocolatey-coreteampackages/blob/ff38b7dd3ed75ba574218cc41ab5a823901f690f/automatic/git.install/tools/chocolateyInstall.ps1#L52-L61

could perhaps be named `Remove-QuickLaunch4SystemUser`.

https://github.com/pascalberger/chocolatey-coreteampackages/blob/ff38b7dd3ed75ba574218cc41ab5a823901f690f/automatic/git.install/tools/chocolateyInstall.ps1#L63-L67

could be `Stop-GitSSHAgent`

You can also incorporate `Get-AppInstallLocation 'Git version*'` instead of handling registry directly via Git_is1'

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,96 @@
...
+    $installLocation = $keyNames.InstallLocation
+    if ($installLocation -ne '') {
+      $gitPath = Join-Path $installLocation 'cmd'
+      Install-ChocolateyPath $gitPath 'Machine'
+    }
+  }
+}
+
+# Lets remove the installer as there is no more need for it
+Remove-Item -Force $filePath32
```
majkinetor: Add `-ea 0` on both. We don't want installer to fail if for some reason file is locked or whatever.

majkinetor:
Requires new parameter from https://github.com/ferventcoder/chocolatey-packages/issues/197: `NoShellIntegration`

majkinetor:
The https://github.com/ferventcoder/chocolatey-packages/issues/149 issue would either have to be done manually (see what git adds and where and remove it by hand) or should be closed on remote repo because it should be solved upstream.


Other then adding an option for disableing `ShellIntegration`, I think this resolves all open issues.

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: I refactored the script and put functionallity in external helper.ps1

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.install/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.install/tools/helpers.ps1
```diff
@@ -0,0 +1,64 @@
+function Get-InstallKey() {
+    $registryKeyName = 'Git_is1'
+    $installKey = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
+    if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -eq 'true') {
```
pascalberger: The logic in this seems wrong to me. But it is the same logic as in the original script [here](https://github.com/ferventcoder/chocolatey-packages/blob/master/automatic/git.install/tools/chocolateyInstall.ps1#L67). 

Shouldn't this be `if ((Get-ProcessorBits 32) -or $env:chocolateyForceX86 -eq 'true')`?

@ferventcoder Any idea about the logic behind this?

automatic/git.portable/tools/chocolateyUninstall.ps1
pascalberger: Added dependeny to `"chocolatey" version="0.9.10"`

automatic/git.install/tools/helpers.ps1
majkinetor: Yes, it should be:
`if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true')`

majkinetor:
@pascalberger could you move verification stuff to the `legal` folder ?

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,32 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$is64bit = Get-ProcessorBits 64
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$filePath32 = "$toolsDir\PortableGit-2.11.0-32-bit.7z.exe"
+$filePath64 = "$toolsDir\PortableGit-2.11.0-64-bit.7z.exe"
+$filePath = if ($is64bit -and $env:chocolateyForceX86 -eq $false) { $filePath64 } else { $filePath32 }
```
majkinetor: `$env:chocolateyForceX86 -ne 'true'`

majkinetor: Its the best to have something like this:

```powershell
$embedded_path = if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true') {
         Write-Host "Installing 64 bit version"; gi "$toolsDir\*-64-*.exe"
} else { Write-Host "Installing 32 bit version"; gi "$toolsDir\*-32-*.exe" }
```

Write-Host is good to have because choco doesn't show anything here unlike with online version.

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,32 @@
...
+$installLocation = Join-Path $(Get-ToolsLocation) 'git'
+
+Get-ChocolateyUnzip -FileFullPath $filePath -Destination $installLocation
+
+# Add to path
+$gitBinPath = Join-Path $installLocation 'bin'
+Install-ChocolateyPath $gitBinPath 'Machine'
+$env:Path = "$($env:Path);$gitBinPath"
+
+# Generate ignore files
```
majkinetor:
```powershell
# Generate ignore files
Write-Host "Generate ignore files"
$files = get-childitem $installLocation -include *.exe -recurse
foreach ($file in $files) {
  if (!($file.FullName -match '\\cmd\\git.exe$|\\git-bash.exe$|ssh[^\\]$')) {
    New-Item "$file.ignore" -type file -force | Out-Null
  }
}
```

This is not needed, its required when in chocolatey lib folder. You can remove this.

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,32 @@
...
+    New-Item "$file.ignore" -type file -force | Out-Null
+  }
+}
+
+# Remove old installations
+$deprecatedInstallDir = Join-Path $env:systemdrive 'git'
+if (Test-Path $deprecatedInstallDir) { Remove-Item $deprecatedInstallDir -recurse -force -ErrorAction SilentlyContinue }
+
+# Lets remove the ZIP files as there is no more need for it
+Remove-Item -Force $filePath32
```
majkinetor: -ea 0

automatic/git.portable/tools/chocolateyInstall.ps1
majkinetor: With online version it shows if it installs x32 or x64 and with embedded packages it doesn't.

majkinetor: Notice that I use wildcars above, but its probably better to use full name. (if removing older installer fails, gi will return multiple files).

 I use wildcards so that same installer works with other tools.

majkinetor: The remaining files problem might not exist. I am not sure what kind of changes choco recently has @ferventcoder . AFAIK old lib directory of the package is cleaned on new install now. I remember that on older choco some stuff would pille up here.

pascalberger:
@majkinetor I've added a `NoShellIntegration` parameter as requested in ferventcoder/chocolatey-packages#197. Also refactored handling of installer arguments a little bit to handle separators more easily.

majkinetor:
Great for new parameter.

I think your new argument handling is overkill. This code acomplishes the same:

`$fileArgs = '/verysilent', '/norestart', '/nocancel', '/sp-', '/log'`

The same with components

When you later put the array into string - ` "$fileArgs"` - it will get joined with spaces.

pascalberger:
@majkinetor Licenses and verification files moved to `legal` folder

pascalberger:
@majkinetor The reason for the argument handling with lists is that it is easy this way to add and remove entries (as it is done in `Remove-QuickLaunchForSystemUser` and `Remove-ShellIntegration`) without needing to take care of separators for each add / remove (avoids stuff like [this](https://github.com/ferventcoder/chocolatey-packages/pull/169/files#diff-38b497b16f7f35551e33f1702ceffb25R125))

majkinetor:
That is even easier and more in Powershell spirit otherwise:

`$x = 'param1', 'param2', 'param3'`

or 

```
$x =  'param1',
      'param2',
      'param3'
```

- Add:  $x =+ 'param4'
- Remove: $x =  $x -ne 'param1' 

I am not sure if Remove syntax works in Posh2. However `$x = $x | ? { $_ -ne 'param1' }` does.

automatic/git.portable/tools/chocolateyInstall.ps1
pascalberger: Done

automatic/git.portable/tools/chocolateyInstall.ps1
pascalberger: Removed

automatic/git.portable/tools/chocolateyInstall.ps1
pascalberger: Adopted. Also changed for `git.install` [here](https://github.com/chocolatey/chocolatey-coreteampackages/pull/489/files#diff-38b497b16f7f35551e33f1702ceffb25R49)

automatic/git.install/tools/helpers.ps1
pascalberger: Fixed

majkinetor:
OK, looks good now all in all. Please just consider about args, that looks ugly and non-powershelly. There is rarely need to resort to non-native stuff in Powershell. Since primary audiency for choco are not programmers but admins, they are usually scared by the code that has mumbo jumbo such as `System.Collections.Generic.List[string]`.

```
$fileArgs = New-Object "System.Collections.Generic.List[string]"
$fileArgs.Add("/VERYSILENT")
$fileArgs.Add("/NORESTART")
$fileArgs.Add("/NOCANCEL")
$fileArgs.Add("/SP-")
$fileArgs.Add("/LOG")

$components = New-Object "System.Collections.Generic.List[string]"
$components.Add("icons")
$components.Add("icons\quicklaunch")
$components.Add("ext")
$components.Add("ext\shellhere")
$components.Add("ext\guihere")
$components.Add("assoc")
$components.Add("assoc_sh")
```

I also consider this to be wall of text :)

pascalberger:
@majkinetor Yeah, already at rewriting the argument stuff. Will push soon

majkinetor:
Great. Thx for the patience.

majkinetor:
Invited others for review because of the importance of the tool. I know that @AdmiringWorm is on vac. but he can take a look after this is merged.

pascalberger:
@majkinetor I changed argument handling. I also added `/SUPPRESSMSGBOXES` always to fileArgs instead of only if run as system user since IMHO I also don't want to see any message boxes if installed as user. Any objection against this?

majkinetor:
Yeah, no GUI stuff if possible forever.

Yeah, args are correct now.

Looking at helpers:

```powershell
    if ($Components.IndexOf("ext\shellhere") -ge 0) {
        $Components = $Components -ne "ext\shellhere"
    }
```

- If is not needed TBH. Many IFs in this world can be removed especially if (Test-Path)'s which is plain wrong.
- Have you checked is this Posh2 compatible ? 

I have problem to do that now as of:
```
PS> powershell -version 2
Version v2.0.50727 of the .NET Framework is not installed and it is required to run version 2 of Windows PowerShell.
```

pascalberger:
@majkinetor Removed the unecessary Ifs. Yeah, checked with Posh2.

Also fixed another bug where a variable was only checked for not being empty instead of not being empty or null [here](https://github.com/chocolatey/chocolatey-coreteampackages/pull/489/files#diff-38b497b16f7f35551e33f1702ceffb25R66).

majkinetor:
I wonder why people `Join-Path` at all. It can't accept multiple arguments a la `[io.path]::combine('c:\', 'foo', 'bar')` and doesn't do anything for platform specific path delimiter as Powreshell alraedy handles both / and \.

automatic/git.install/tools/helpers.ps1
```diff
@@ -0,0 +1,69 @@
+function Get-InstallKey() {
+    $registryKeyName = 'Git_is1'
+    $installKey = "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
+    if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true') {
```
AdmiringWorm: this looks wrong to me, basically you're saying if we are running on 64bit and 32bit install isn't forced, use the Wow6432Node registry location.
I believe it should be:
```
(Get-ProcessorBits 64) -and $env:chocolateyForceX86 -eq 'true'
```

automatic/git.portable/tools/chocolateyUninstall.ps1
AdmiringWorm: doesn't `Get-ChocolateyUnzip` output a text file in the package directory with the full path to the extracted files? if so, couldn't this file be used instead?

if it does you can call the `Uninstall-ChocolateyZipPackage` function with package name and zip file name (including extension)

automatic/git.install/tools/helpers.ps1
majkinetor: Yeah its wrong, but not in above but order `$installKey` values should be ordered the opposite way.

automatic/git.install/tools/helpers.ps1
ferventcoder: There used to only be a 32 bit version. The logic here can likely be refactored/removed entirely

automatic/git.install/tools/chocolateyInstall.ps1
ferventcoder: To resuce bugs look for both keys

ferventcoder: Reduce

ferventcoder:
Yeah - maybe best to add to the wiki that it is an anti-pattern to set requires powershell -version 2. It breaks newer Windows that need it added later

majkinetor:
Closing this one, want to do refactoring in a new core branch.

pascalberger:
Reopening, ss there still are open discussions in #499, so that we first can take over the code and do any refactorings on top of it.

automatic/git.install/tools/helpers.ps1
pascalberger: Fixed this

pascalberger:
I fixed all open issues on this one? 

@ferventcoder @majkinetor @gep13 Can you have another look and if it is OK I would like to merge this, #499 should be replaced by a new PR for just the refactorings on top of it (or rebased)

ferventcoder:
This looks good to me without testing the actual install bits.

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,30 @@
...
+Install-ChocolateyPath $gitBinPath 'Machine'
+$env:Path = "$($env:Path);$gitBinPath"
+
+# Remove old installations
+$deprecatedInstallDir = Join-Path $env:systemdrive 'git'
+if (Test-Path $deprecatedInstallDir) { Remove-Item $deprecatedInstallDir -recurse -force -ErrorAction SilentlyContinue }
+
+# Lets remove the ZIP files as there is no more need for it
+Remove-Item -Force $filePath32 -ea 0
+Remove-Item -Force $filePath64 -ea 0
```
AdmiringWorm: I believe there should be a call to `post-install.bat` file located in the extracted git directory (called through git-bash.exe).

automatic/git.install/update.ps1
```diff
@@ -0,0 +1,75 @@
+import-module au
+
+$releases = "https://git-scm.com/download/win"
+
+function global:au_BeforeUpdate {
```
pascalberger: Fixed

automatic/git.portable/tools/chocolateyInstall.ps1
pascalberger: Do you have any documentation about this? Anyway I would suggest to merge this with the features it currently has, which already has a lot of improvements compared to ferventcoders version. Especially since it is already deleted in ferventcoders repo and the package won't be updated anymore until we merge this. Anything additional can later be done in additional PRs, otherwise I fear this will end in a neverending story

AdmiringWorm: I don't have any documentation of it other than what is documented in the bat file itself.
And considering they say themself in the bat file
```
This script is intended to be run after installation of Git for Windows
(including the portable version)
```

but yeah, this can be done later if/when an issue is reported that the bat file may fix.
I'll merge it for now, and we'll see what happens.

pascalberger:
@AdmiringWorm I think we need now to manually push the new `git.portable` package, afterwards @ferventcoder needs to push the deprecation of `git.commandline`, correct?

AdmiringWorm:
@pascalberger the git.portable package was pushed earlier.

pascalberger:
@AdmiringWorm Mmh, https://chocolatey.org/packages?q=git.portable doesn't returns anything for me. Maybe load balancing issue?

AdmiringWorm:
just hasn't passed moderation yet.
https://chocolatey.org/packages/git.portable
since it's a new package it needs to be approved first.

pascalberger:
Ah, makes sense :)

ferventcoder:
https://chocolatey.org/packages/git.portable I left a review note requesting a fix. You can't call administrative methods in portable packages. Trying to install to machine path is an administrative method. See Administrative access functions at - https://chocolatey.org/docs/helpers-reference

ferventcoder:
Also, which issue on the list made the change for extracting git portable out of the package directory? We've been leaning towards keeping things IN the package directory, especially portables. And then optionally allowing someone to install them outside of the directory with a package parameter - see sysinternals package.

pascalberger:
@ferventcoder Sorry, didn't see what you're referencing to regarding changed directory. [Previously](https://github.com/ferventcoder/chocolatey-packages/pull/224/files#diff-08de1ebeb44823ede4f1952f7f80aa8fL9) it was this code:

```
$binRoot = Get-BinRoot
$gitInstallDir = Join-Path $binRoot 'git'
```

In the new package I replaced the deprecated `Get-BinRoot` with `Get-ToolsLocation`:

```
$installLocation = Join-Path $(Get-ToolsLocation) 'git'
```

Everything else should keep the same, except that it is now an embedded package. Did I miss anything?

Sorry, missed that `Install-ChocolateyPath` requires admin privileges. Which function is intended for unzipping local files in portable packages?

ferventcoder:
> Sorry, didn't see what you're referencing to regarding changed directory. 

Ah, right. My bad. Just looked and now I remember that was changed awhile back. Ignore me on that.

ferventcoder:
> Sorry, missed that Install-ChocolateyPath requires admin privileges. Which function is intended for unzipping local files in portable packages?

You should use `Install-ChocolateyPath` without "Machine".

pascalberger:
@ferventcoder See https://github.com/chocolatey/chocolatey-coreteampackages/pull/516

PR #499, opened 2016-12-25
(git) refactoring after migration

majkinetor:
Original PR: https://github.com/chocolatey/chocolatey-coreteampackages/pull/489


majkinetor:
I was not happy about the complexity of this code so I quite changed everything. Please take a look.

majkinetor:
**Force update test**

```
## git~> fupdate
git.install - checking updates using au version 2016.12.25.101651
URL check
  https://github.com/git-for-windows/git/releases/download/v2.11.0.windows.1/Git-2.11.0-32-bit.exe
  https://github.com/git-for-windows/git/releases/download/v2.11.0.windows.1/Git-2.11.0-64-bit.exe
nuspec version: 2.11.0
remote version: 2.11.0
No new version found, but update is forced
New version is available
Automatic checksum skipped
Running au_BeforeUpdate
Updating files
  $Latest data:
    Checksum32      (String)    2A6083479538C4FE454336660FCE96346EE3CF46F99CE08A666D4635539239D7
    Checksum64      (String)    FD1937EA8468461D35D9CABFCDD2DAA3A74509DC9213C43C2B9615E8F0B85086
    ChecksumType    (String)    sha256
    FileName32      (String)    Git-2.11.0-32-bit.exe
    FileName64      (String)    Git-2.11.0-64-bit.exe
    FileType        (String)    exe
    NuspecVersion   (String)    2.11.0
    PackageName     (String)    git.install
    URL32           (String)    https://github.com/git-for-windows/git/releases/download/v2.11.0.windows.1/Git-2.11.0-32-bit.exe
    URL64           (String)    https://github.com/git-for-windows/git/releases/download/v2.11.0.windows.1/Git-2.11.0-64-bit.exe
    Version         (String)    2.11.0.20161225

  git.install.nuspec
    setting id:  git.install
    using Chocolatey fix notation: 2.11.0 -> 2.11.0.20161225
  .\legal\verification.txt
    (?i)(checksum type:\s+).* = ${1}sha256
    (?i)(32-Bit.+)\<.*\> = ${1}<https://github.com/git-for-windows/git/releases/download/v2.11.0.windows.1/Git-2.11.0-32-bit.exe>
    (?i)(checksum32:\s+).* = ${1}2A6083479538C4FE454336660FCE96346EE3CF46F99CE08A666D4635539239D7
    (?i)(checksum64:\s+).* = ${1}FD1937EA8468461D35D9CABFCDD2DAA3A74509DC9213C43C2B9615E8F0B85086
    (?i)(64-Bit.+)\<.*\> = ${1}<https://github.com/git-for-windows/git/releases/download/v2.11.0.windows.1/Git-2.11.0-64-bit.exe>
  .\tools\chocolateyInstall.ps1
    (^[$]filePath32\s*=\s*"[$]toolsPath\\)(.*)" = $1Git-2.11.0-32-bit.exe"
    (^[$]filePath64\s*=\s*"[$]toolsPath\\)(.*)" = $1Git-2.11.0-64-bit.exe"
Attempting to build package from 'git.install.nuspec'.
Successfully created package 'C:\work\_chocolatey\chocolatey-coreteampackages\automatic\git.install\git.install.2.11.0.20161225.nupkg'
Package updated
```

**`Test-Package`**

```
==> default: ============================================================
==> default:
==> default: TESTING FOLLOWING PACKAGES: git.install:2.11.0.20161225
==> default: ============================================================
==> default:
==> default: ------------------------------------------------------------
==> default: PACKAGE: git.install:2.11.0.20161225
==> default:
==> default: OPTIONS: c:\packages\git.install.2.11.0.20161225.xml
==> default:
==> default: Name                           Value
==> default: ----                           -----
==> default: Install                        True
==> default: Uninstall                      False
==> default: Parameters
==> default: ------------------------------------------------------------
==> default:
==> default: TESTING INSTALL FOR git.install:2.11.0.20161225
==> default: CMD: choco install -fy git.install --allow-downgrade --version 2.11.0.20161225 --source "'c:\packages;http://chocolatey.org/api/v2/'"
==> default: Chocolatey v0.10.3
==> default: Installing the following packages:
==> default:
==> default: git.install
==> default: By installing you accept licenses for the packages.
==> default: git.install v2.11.0.20161225 (forced)
==> default: git.install package files install completed. Performing other installation steps.
==> default: Installing 64 bit version
==> default: Installing git.install...
==> default: git.install has been installed.
==> default: Environment Vars (like PATH) have changed. Close/reopen your shell to
==> default:
==> default:  see the changes (or in powershell/cmd.exe just type `refreshenv`).
==> default:  The install of git.install was successful.
==> default:
==> default:   Software installed to 'C:\Program Files\Git\'
```

**`Test-Package -Parameters '/NoShellIntegration'`**

```
==> default: ============================================================
==> default: TESTING FOLLOWING PACKAGES: git.install:2.11.0.20161225
==> default: ============================================================
==> default:
==> default: ------------------------------------------------------------
==> default: PACKAGE: git.install:2.11.0.20161225
==> default:
==> default: OPTIONS: c:\packages\git.install.2.11.0.20161225.xml
==> default:
==> default: Name                           Value
==> default: ----                           -----
==> default: Install                        True
==> default: Uninstall                      False
==> default: Parameters                     /NoShellIntegration
==> default: ------------------------------------------------------------
==> default:
==> default: TESTING INSTALL FOR git.install:2.11.0.20161225
==> default: CMD: choco install -fy git.install --allow-downgrade --version 2.11.0.20161225 --source "'c:\packages;http://chocolatey.org/api/v2/'"  --params '/NoShellIntegration'
==> default: Chocolatey v0.10.3
==> default: Installing the following packages:
==> default:
==> default: git.install
==> default: By installing you accept licenses for the packages.
==> default: git.install v2.11.0.20161225 (forced)
==> default: git.install package files install completed. Performing other installation steps.
==> default: Installing 64 bit version
==> default: Parameter: no git shell integration
==> default:
==> default: Installing git.install...
==> default: git.install has been installed.
==> default:
==> default: Environment Vars (like PATH) have changed. Close/reopen your shell to
==> default:  see the changes (or in powershell/cmd.exe just type `refreshenv`).
==> default:  The install of git.install was successful.
==> default:
==> default:   Software installed to 'C:\Program Files\Git\'
```

NoShellIntegration worked OK.

**NOTE**: 
I removed Install-ChocolateyPath as it is redudant with package args.

But as shown here, `Git\cmd` ends up on the path even tho no parameter is passed (blank test machine) which is not correct behavior. I think this should be default anyway and that params should be reversed.  **@ferventcoder needs to speed up choco#797 A LOT**. Its definite that `Install-ChocolateyPath` in the script was there just for the update as installer args are not saved (which is again wrong).

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,25 @@
...
+$packageArgs = @{
+    PackageName    = 'git.install'
+    FileType       = 'exe'
+    SoftwareName   = 'Git version *'
+    File           = $installFile
+    SilentArgs     = "/VERYSILENT", "/SUPPRESSMSGBOXES", "/NORESTART", "/NOCANCEL", "/SP-", "/LOG", (Get-InstallComponents $pp)
+    ValidExitCodes = @(0)
+}
+Install-ChocolateyInstallPackage @packageArgs
+rm -Force $filePath32, $filePath64 -ea 0
```
pascalberger: Avoid aliases as per [§1.1.8](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#118-clean-code). Use `Remove-Item` instead

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,25 @@
...
+    FileFullPath = $filePath
+    Destination  = "$(Get-ToolsLocation)\git"
+}
+Get-ChocolateyUnzip @packageArgs
+Install-ChocolateyPath $packageArgs.Destination
+
+$deprecatedInstallDir = Join-Path $env:systemdrive 'git'
+if (Test-Path $deprecatedInstallDir) {
+    Write-Host "Removing old installation at: $deprecatedInstallDir"
+    rm $deprecatedInstallDir -recurse -force -ea 0
```
pascalberger: Avoid aliases as per [§1.1.8](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#118-clean-code). Use `Remove-Item` instead

automatic/git.install/tools/helpers.ps1
```diff
@@ -0,0 +1,49 @@
...
+    $isSystem = ([System.Security.Principal.WindowsIdentity]::GetCurrent()).IsSystem
+    if ( !$isSystem ) { $res += "icons\quicklaunch" }
+
+    if ($res.Length -eq 0) { return }
+    '/COMPONENTS="{0}"' -f ($res -join ",")
+}
+
+function Stop-GitSSHAgent()
+{
+    if (!(ps ssh-agent -ea 0)) { return }
```
pascalberger: Avoid aliases as per [§1.1.8](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#118-clean-code). Use `Get-Process` instead

automatic/git.install/tools/helpers.ps1
```diff
@@ -0,0 +1,49 @@
...
+
+    Write-Host "Killing any running git ssh-agent instances"
+    ps ssh-agent | ? {$_.Path -ilike "*\git\usr\bin\*"} | kill
```
pascalberger: Avoid aliases as per [§1.1.8](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#118-clean-code). Use `Get-Process` instead

automatic/git.install/git.install.nuspec
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <metadata>
+    <id>git.install</id>
+    <title>Git (Install)</title>
+    <version>2.11.0</version>
+    <authors>The Git Development Community</authors>
+    <owners>chocolatey</owners>
+    <summary> Git or Windows offers a native set of tools that bring the full feature set of the Git SCM to Windows</summary>
```
pascalberger: `Git or Windows` makes no sense: Software is called `Git for Windows`. Also remove trailing space introduced at the beginning

automatic/git/git.nuspec
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <metadata>
+    <id>git</id>
+    <title>Git</title>
+    <version>2.11.0</version>
+    <authors>The Git Development Community</authors>
+    <owners>Chocolatey</owners>
+    <summary>Git or Windows offers a native set of tools that bring the full feature set of the Git SCM to Windows</summary>
```
pascalberger: `Git or Windows` makes no sense: Software is called `Git for Windows`. Also remove trailing space introduced at the beginning

automatic/git/git.nuspec
```diff
@@ -6,29 +6,27 @@
     <version>2.11.0</version>
     <authors>The Git Development Community</authors>
     <owners>Chocolatey</owners>
-    <summary>Git (for Windows) – We bring the awesome Git SCM to Windows</summary>
+    <summary>Git or Windows offers a native set of tools that bring the full feature set of the Git SCM to Windows</summary>
     <description>
 Git for Windows focuses on offering a lightweight, native set of tools that bring the full feature set of the Git SCM to Windows while providing appropriate user interfaces for experienced Git users and novices alike.
 
-## Features
```
pascalberger: Why remove the feature section as this is required by the guidelines ([$1.2.3](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#123-informative-description)

automatic/git.portable/git.portable.nuspec
```diff
@@ -6,7 +6,7 @@
     <version>2.11.0</version>
     <authors>The Git Development Community</authors>
     <owners>Chocolatey</owners>
```
pascalberger: You changed this to lowercase in the other packages should therefore be lower case here too

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: No. It says:

> Use Aliases sparingly

`rm` is WAY MORE KNOWN then `Remove-Item`; The same goes for others. I can go into neurolinguistics to prove that using cross-platform aliases is less confusing and requires less effort on brain.

automatic/git.install/git.install.nuspec
majkinetor: LOL, ofc it doesn't make sense, its a typo :)

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: That was added by @ferventcoder and should probably get removed. It doesn't represent guideline if you have an option not to use it - use sparingly is open to interpretation. Other reasons mentioned are moot:

>They tend to make PowerShell harder to read, 

- It depend on alias (cross-platform or not?)

> Sometimes they are not available in all supported versions of PowerShell

This is redudant rule - we already have posh2 compatibility rule.

> and someone can override aliases locally 

The argument is invalid. Someone can override anything locally, cmdlets included. The aliases do not fix that but only Full Quoalified Cmdlet Name which then makes it unreadable. default aliases are read only so this is even incorrect.

pascalberger: But in this case we should consistently use `rm` over `Remove-Item` and not mix them in the same file as it is currently (`chocolateyInstall.ps` for `git.portable`). And I would also propose to write down in the guidelines which aliases are considered OK, as in the end this is also a matter of taste.

majkinetor: I think this issue doesn't deserve that much attention because its hard to maintain it and isn't actually important.

Yes, everything here is a matter of taste, but how about replacing entire wiki section with this:

> Do not generally use non readonly aliases. Aliases such as `cat`, `rm`, `ps`, `ls` etc are OK as their behavior is well known on multiple OS. Besides that, using `%` and `?` is recommended over non-aliased versions. On the other hand, PowerShell specific commands such as `Invoke-WebRequest` should not be used in alias form (`iwr`) as they tend to be confusing to newcomers or (`wget`) implement radically different interface.

majkinetor: Actually, it can be enforced. Supposed to be in this issue https://github.com/majkinetor/au/issues/55. But I don't know when.

pascalberger: @majkinetor Sounds good to me. IMHO important is to be consistent (eg. not mix aliases with non-aliases in the same file, since this can be quite confusing)

majkinetor: I agree, I didn't catch it, changed now.

automatic/git.install/tools/helpers.ps1
ferventcoder: I do think Get-Process in this case. Only folks who have significant xplat experience will recognize ps.

automatic/git.install/tools/helpers.ps1
```diff
@@ -0,0 +1,48 @@
...
+    if ($pp.NoShellIntegration) {
+        Write-Host "Parameter: no git shell integration"
+    } else { $res += "ext", "ext\shellhere", "ext\guihere" }
+
+    # Make our install work properly when running under SYSTEM account (Chef Cliet Service, Puppet Service, etc)
+    $isSystem = ([System.Security.Principal.WindowsIdentity]::GetCurrent()).IsSystem
+    if ( !$isSystem ) { $res += "icons\quicklaunch" }
+
+    if ($res.Length -eq 0) { return }
+    '/COMPONENTS="{0}"' -f ($res -join ",")
```
ferventcoder: Look at this - https://github.com/iTob191/chocolatey-packages/commit/7cdc7cc98cda3e1f909ddd118d38b00c3c0f0d94

I prefer we move towards this.

ferventcoder: @majkinetor what do you think?

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,25 @@
...
+
+$packageArgs = @{
+    PackageName    = 'git.install'
+    FileType       = 'exe'
+    SoftwareName   = 'Git version *'
+    File           = $installFile
+    SilentArgs     = "/VERYSILENT", "/SUPPRESSMSGBOXES", "/NORESTART", "/NOCANCEL", "/SP-", "/LOG", (Get-InstallComponents $pp)
+    ValidExitCodes = @(0)
+}
+Install-ChocolateyInstallPackage @packageArgs
```
ferventcoder: Maybe too much refactoring. You can't completely remove `Install-ChocolateyPath` - otherwise there may be instances you will get nothing on PATH (unless that has been fixed in the installer).

Not always will people pass a package parameter for the PATH set, so the old package was handling those scenarios as well (not great, but handling).

automatic/git.install/tools/chocolateyInstall.ps1
ferventcoder: > Do not generally use non read-only aliases. Aliases such as cat, rm, ps, ls, kill etc are OK as their behavior is well known on multiple OS.

Consider your audience. This is a Windows package manager. You and I are familiar with these commands, but that doesn't mean that others are. So I would disagree on these being well-known for the Windows community.

However I do agree that  `%`, `?`, `select` are widely used in scripts and in frameworks.

automatic/git.install/tools/helpers.ps1
majkinetor: **BABYSITTING ALARM**

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: And in this case you get something on PATH even you don't want it. This is so confusing that even I don't know what param to use here.

majkinetor: Its trivial for me to return this back. However, I think you should concentrate on choco#797 rather then us inventing strnage mumbo jumbo.

automatic/git.install/tools/helpers.ps1
gep13: @majkinetor what exactly do you mean by this?

majkinetor: You have a habit or repeating that sentence @gep13  :)

gep13: So what does that tell you?

automatic/git.install/tools/helpers.ps1
majkinetor: @ferventcoder 

I think its 100+ lines of code more for the same thing. While LOC is not particularly good measure for code quality that code is basically unreadable and breaks number of paradigms of good programing. This is like PHP stuff - lets put everything in a single file and hope that nobody will ever change it. 

Contrary to that, refactored code has a benefit of
- SRP (single responsibility principle)
- DRY (do not repeat)
- Progressive disclosure - anybody looking the code will see critical phases - set parameters, stop agent , run installer, set path - on a single screen. If you want to know particular details on how arguments are created you go into helpers and look single function handling single responsibility.

Write-Hosts announcing parameters is a good thing but I think its better to do that in core extension with something like `Get-PackageParameters -Announce`. Its not the same as you can provide more descriptive information  but in most cases paramters are self describing (do I really need to write _'you want git on path'_ if parameter is `GitOnPath` and is shown during installation in a manner like _User provided parameters: GitOnPath, NoQuickLaunch_.)

automatic/git.install/tools/helpers.ps1
majkinetor: Diagnosis could include OCD, ADHD, FFI, FLD ... number of things really. 😀

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: I returned back code that adds PATH: https://github.com/chocolatey/chocolatey-coreteampackages/pull/499/commits/1e1964afa6d5aaf468d6f37167dd07fcd955d852

majkinetor: > Not always will people pass a package parameter for the PATH set,

Thats why the parameters should be reversed - `Noxxxx` (i.e. `NoGitOnPath`). With repositories people expect that behavior.

ferventcoder: > However, I think you should concentrate on choco#797 rather then us inventing strange mumbo jumbo.

Not strange mumbo jumbo. I'm also not a fan of being told what to do.

ferventcoder: > And in that case you get something on PATH even you don't want it. This is so confusing that even I don't know what param to use here.

To install git with no params and expect `git` to work immediately is always desired. I did remove the install PATH at one point and so many people complained I put it back. Many packages do this for most installers. If they do not add PATH, lots of packages will add it - see VirtualBox and other packages. Plus

automatic/git.install/tools/helpers.ps1
ferventcoder: Sorry, I didn't mean just as is. I meant the list of available components and then switching them on and off. It's all together for those components so it increases readability and is dry.

Apologies that I need to get technical, but I mean specifically:

~~~powershell
$components = @{
     "icons"             = $true;
     "icons\quicklaunch" = $true;
     "ext"               = $true;
     "ext\shellhere"     = $true;
     "ext\guihere"       = $true;
     "assoc"             = $true;
     "assoc_sh"          = $true;
 };
~~~

ferventcoder: Apologies that you misunderstood what I was saying.

automatic/git.install/tools/helpers.ps1
ferventcoder: The point is to make the code readable, discoverable to even those with little PowerShell experience so they can follow along with what is going on. 

Making things easier for others is one of the biggest reasons that Chocolatey has been so successful. ;)

ferventcoder:
> @ferventcoder needs to speed up choco#797 A LOT. Its definite that Install-ChocolateyPath in the script was there just for the update as installer args are not saved (which is again wrong).

Why would this be removed? The installer doesn't remove the current path on upgrade.

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: > Not strange mumbo jumbo. I'm also not a fan of being told what to do. This is disrespectful IMHO

I am not telling you. I said: _**I think** you should concentrate on choco#797_.

majkinetor:
> Why would this be removed? The installer doesn't remove the current path on upgrade.

I reverted that.

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: > To install git with no params and expect git to work immediately is always desired

Majkinetor wrote:
> ..parameters should be reversed - Noxxxx (i.e. NoGitOnPath).

ferventcoder:
I meant, why would an upgrade remove a PATH value?

automatic/git.install/tools/chocolateyInstall.ps1
ferventcoder: I agree with that - there was a time when I was moving to asking folks to explicitly request for the PATH to be added. Folks didn't want it. So GitOnPath is similar to LinuxToolsOnPath, but really one only needs the LinuxTools for the additional path add, which may not be desired. So I agree with "NoGitOnPATH".

automatic/git.install/tools/helpers.ps1
majkinetor: Why so much appoligies ? 😕 

Lets analyze:

```powershell
function Get-InstallComponents( [HashTable]$pp )
{
    $res = "icons", "assoc", "assoc_sh"   ; #1

    if ($pp.NoShellIntegration) {  #2
        Write-Host "Parameter: no git shell integration"
    } else { $res += "ext", "ext\shellhere", "ext\guihere" }

    # Make our install work properly when running under SYSTEM account (Chef Cliet Service, Puppet Service, etc)
    $isSystem = ([System.Security.Principal.WindowsIdentity]::GetCurrent()).IsSystem
    if ( !$isSystem ) { $res += "icons\quicklaunch" } #3

    if ($res.Length -eq 0) { return }
    '/COMPONENTS="{0}"' -f ($res -join ",") 
}
```

- 1 means those opotions are ALWYAS applied
- 2 and 3 are contextual. So you could look your proposed version of the code and espect that changing default there would be enough, but it wouldn't. This code states that intention.
- Above code works as a blackbox and you don't have to read anything else in the package to understand it or change it. Contary to that, you need to be aware or 100+ lines of code to change those 7 lines.

majkinetor:
> I meant, why would an upgrade remove a PATH value?

Upgrade shouldn't. But parameters are not good. If I have an option `GitOnPath` that begs the question **why do I have it as it works even if I don't use it** ? Since everybody wants git on path, the option should then be reversed to `NoGitOnPath`. But if its reveresed **and not remembered,** the update could remove existing path depending on how you implement it - NoGitOnPath could ignore any current situation or could actually **remove** it from the PATH if you have it there (I beleive this is correct solution). 

In short, current parameters are confusing and behave unintuitive.

ferventcoder:
> In short, current parameters are confusing and behave unintuitive.

Or at least GitOnPath. I agree with the assessment to have NoGitOnPath. 👍

automatic/git.install/tools/helpers.ps1
ferventcoder: That's fine as well. :D

pascalberger:
Looking at the number of comments and considering there are still open points to discuss, wouldn't it be better to merge first #489, and then do any refactorings on top of it in subsequent PRs instead of trying to put everything in a single PR?

ferventcoder:
Sounds good to me.

pascalberger:
I reopened #489 to get the git package code into this repository, without the refactorings of this PR

AdmiringWorm:
I just merged master into git to resolve the file conflicts, looks like I messed up your review in the process @pascalberger . Sorry about that.

AdmiringWorm:
@pascalberger , @ferventcoder 
could you guys perhaps review this one again?
I've expanded the aliases and merged the latest git.portable changes.

automatic/git/git.nuspec
```diff
@@ -6,28 +6,26 @@
     <version>2.11.0.2</version>
     <authors>The Git Development Community</authors>
     <owners>chocolatey</owners>
-    <summary>Git (for Windows) â€“ We bring the awesome Git SCM to Windows</summary>
+    <summary>Git for Windows offers a native set of tools that bring the full feature set of the Git SCM to Windows</summary>
     <description>
 Git for Windows focuses on offering a lightweight, native set of tools that bring the full feature set of the Git SCM to Windows while providing appropriate user interfaces for experienced Git users and novices alike.
 
-## Features
```
pascalberger: Why remove features?

automatic/git.portable/tools/chocolateyUninstall.ps1
```diff
@@ -8,4 +8,4 @@ Remove-Item $installLocation -Recurse -Force -ea 0
 $newPath = $Env:Path.Replace(";$installLocation", '')
 if ($newPath -eq $Env:PATH) { return }
 Write-Host "Removing Git from system PATH"
-[System.Environment]::SetEnvironmentVariable('PATH', $newPath, 'Machine')
\ No newline at end of file
+[System.Environment]::SetEnvironmentVariable('PATH', $newPath, 'Machine')
```
pascalberger: Not directly related to changes in this PR, but is it OK to change machine wide settings in portable package?

automatic/git.install/tools/helpers.ps1
```diff
@@ -1,69 +1,48 @@
-function Get-InstallKey() {
-    $registryKeyName = 'Git_is1'
-    $installKey = "HKLM:\SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
-    if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true') {
```
pascalberger: Wrong logic. See #526

automatic/git.install/tools/helpers.ps1
```diff
@@ -1,69 +1,48 @@
...
-        New-Item -Path $InstallKey | Out-Null
-    }
+function Set-InstallerRegistrySettings( [HashTable]$pp )
+{
+    $installkey = Get-InstallKey
+    $ino = "Inno Setup CodeFile:"
 
-    if ($GitOnlyOnPath) {
-        # update registry so installer picks it up automatically
-        New-ItemProperty $InstallKey -Name "Inno Setup CodeFile: Path Option" -Value "Cmd" -PropertyType "String" -Force | Out-Null
```
pascalberger: Was the `| Out-Null` intentaionally removed from all `New-ItemProperty` calls?

automatic/git.portable/tools/chocolateyInstall.ps1
```diff
@@ -1,30 +1,26 @@
...
 
-Get-ChocolateyUnzip -FileFullPath $filePath -Destination $installLocation
+$filePath   = if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true') {
+                     Write-Host "Installing 64 bit version"; $filePath64 }
+              else { Write-Host "Installing 32 bit version"; $filePath32 }
 
-# Add to path
-$gitBinPath = Join-Path $installLocation 'bin'
-Install-ChocolateyPath $gitBinPath
-$env:Path = "$($env:Path);$gitBinPath"
```
pascalberger: Why remove this line?

automatic/git.install/tools/chocolateyInstall.ps1
```diff
@@ -1,75 +1,33 @@
...
 Install-ChocolateyInstallPackage @packageArgs
 
-if (Test-Path $installKey) {
-    $keyNames = Get-ItemProperty -Path $installKey
+$packageName = $packageArgs.packageName
+$installLocation = Get-AppInstallLocation $packageArgs.SoftwareName
+if (!$installLocation)  { Write-Warning "Can't find $packageName install location"; return }
+Write-Host "$packageName installed to '$installLocation'"
 
-    if (!$pp.GitOnlyOnPath -and !$pp.GitAndUnixToolsOnPath) {
```
pascalberger: The new logic seems different to me than the old one. Eg. `Install-ChocolateyPath` is no called nevertheless the values of `GitOnlyOnPath` and `GitAndUnixToolsOnPath`

automatic/git.install/tools/helpers.ps1
```diff
@@ -1,69 +1,48 @@
...
-    $userInstallKey = "HKCU:\Software\Microsoft\Windows\CurrentVersion\Uninstall\$registryKeyName"
-    if (Test-Path $userInstallKey) {
-        $installKey = $userInstallKey
-    }
-
-    return $installKey
+function Get-InstallKey()
+{
+    $keyName = 'Git_is1'
+    $installKey = if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true') {
```
pascalberger: Wrong logic. See #526

automatic/git.portable/tools/chocolateyInstall.ps1
majkinetor: Its not needed.

automatic/git.install/tools/helpers.ps1
majkinetor: Yeah

automatic/git.install/tools/helpers.ps1
majkinetor: I think it outputs nothing anyway

automatic/git.install/tools/chocolateyInstall.ps1
majkinetor: Its not needed. Its handled by `InstallerRegistrySettings`

majkinetor:
Be all as it may, In the meantime, I lost interest in being the maintainer for entire repository for specific reasons. My input in the future will be minimal comparing to what happened in previous year. 

You guys should feel free to change anything here before eventual merge or just close the PR, I will leavve that up to you.

automatic/git.portable/tools/chocolateyInstall.ps1
majkinetor: See Install-ChocolateyPath source code for details.

automatic/git.portable/tools/chocolateyUninstall.ps1
majkinetor: This way package will work for all users AFAIK.

automatic/git/git.nuspec
majkinetor: Probably a mistake.

automatic/git.portable/tools/chocolateyUninstall.ps1
AdmiringWorm: @majkinetor that may be, but portable packages are supposed to work without needing administrative privileges.

AdmiringWorm:
> You guys should feel free to change anything here before eventual merge or just close the PR, I will leavve that up to you.

I was planning to make some changes and continue on this eventually.

AdmiringWorm:
@pascalberger I squashed all commits, and I believe changed/added comment on what you reviewed.

pascalberger:
Changes look good to me without actually testing it

pascalberger:
But we should wait with merging this, since git 2.11.0.3 was already pushed and approved, but git.install 2.11.0.3 timed out and needs to be repushed. So we should push git.install 2.11.0.3 again, and after it is approved can merge this.

pascalberger:
Mmh, git.install 2.11.0.3 seems be now also be pushed, even though it reported an error https://gist.github.com/gep13/bd2eaa76f2a9ab739ca0544c502dca6e#git.install

AdmiringWorm:
@pascalberger even though it timed out, it was still pushed.
https://chocolatey.org/packages/git.install/2.11.0.3

It just timed out because it couldn't get a request back in time.

AdmiringWorm:
@pascalberger I've fixed the merge conflicts, so whenever you get around to test the changes this can then be merged.

PR #505, opened 2017-01-02
Fixing an install issue on CutePDF Writer when print spooler service is not up and running

Thilas:
Chocolatey verifier service environment happens to have this service disabled. That's why cutepdf was failing.

Otherwise, I found out that uninstall wasn't silent and cannot be right now without an AutoIt script (see http://superuser.com/questions/580555/cutepdf-writer-problems-with-silent-uninstall). I didn't change a thing on this side.

### DIFF
--- a/automatic/cutepdf/cutepdf.nuspec
+++ b/automatic/cutepdf/cutepdf.nuspec
@@ -6,7 +6,7 @@
     <title>CutePDF Writer</title>
     <version>3.1</version>
     <authors>Acro Software Inc.</authors>
-    <owners>Redsandro,Thilas</owners>
+    <owners>chocolatey,Redsandro,Thilas</owners>
     <summary>Free PDF printer driver with no watermarks.</summary>
     <description>
 Portable Document Format (PDF) is the de facto standard for the secure and reliable distribution and exchange of electronic documents and forms around the world.  CutePDF Writer is the free version of commercial PDF converter software. CutePDF Writer installs itself as a "printer subsystem". This enables virtually any Windows applications (must be able to print) to convert to professional quality PDF documents - with just a push of a button!
--- a/automatic/cutepdf/cutepdf.nuspec
+++ b/automatic/cutepdf/cutepdf.nuspec
@@ -3,10 +3,10 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
   <metadata>
     <id>cutepdf</id>
-    <title>CutePDF</title>
-    <version>3.0</version>
+    <title>CutePDF Writer</title>
+    <version>3.1</version>
     <authors>Acro Software Inc.</authors>
-    <owners>Redsandro</owners>
+    <owners>Redsandro Thilas</owners>
     <summary>Free PDF printer driver with no watermarks.</summary>
     <description>
 Portable Document Format (PDF) is the de facto standard for the secure and reliable distribution and exchange of electronic documents and forms around the world.  CutePDF Writer is the free version of commercial PDF converter software. CutePDF Writer installs itself as a "printer subsystem". This enables virtually any Windows applications (must be able to print) to convert to professional quality PDF documents - with just a push of a button!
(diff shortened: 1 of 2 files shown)


automatic/cutepdf/cutepdf.nuspec
```diff
@@ -3,10 +3,10 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
   <metadata>
     <id>cutepdf</id>
-    <title>CutePDF</title>
-    <version>3.0</version>
+    <title>CutePDF Writer</title>
+    <version>3.1</version>
     <authors>Acro Software Inc.</authors>
-    <owners>Redsandro</owners>
+    <owners>Redsandro Thilas</owners>
```
AdmiringWorm: should be comma separated.

automatic/cutepdf/tools/chocolateyInstall.ps1
```diff
@@ -1,16 +1,30 @@
...
 $url = 'http://www.cutepdf.com/download/CuteWriter.exe'
 
 $packageArgs = @{
   packageName   = 'cutepdf'
-  unzipLocation = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
   fileType      = 'exe'
   url           = $url
   silentArgs    = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-' # Inno Setup Package
   validExitCodes= @(0)
-  softwareName  = 'cutepdf*'
```
AdmiringWorm: Don't remove software name, this is used in pro/business version of choco to detect if it's already installed.

automatic/cutepdf/tools/chocolateyInstall.ps1
```diff
@@ -1,16 +1,30 @@
...
 }
 
+# Make sure Print Spooler service is up and running
+try {
+  $serviceName = 'Spooler'
+  $spoolerService = Get-WmiObject -Class Win32_Service -Property StartMode,State -Filter "Name='$serviceName'"
+  if ($spoolerService -eq $null) { throw "Service $serviceName was not found" }
+  Write-Host "Print Spooler service state: $($spoolerService.StartMode) / $($spoolerService.State)"
+  if ($spoolerService.StartMode -ne 'Auto' -or $spoolerService.State -ne 'Running') {
+    Set-Service $serviceName -StartupType Automatic -Status Running
```
AdmiringWorm: if we had to manually change the status, it should be changed back to the previous state when the installation have finished.
If the user have disabled the service, they probably don't want to have it enabled unless when absolutely necessary.

AdmiringWorm:
@Thilas are you able to make the requested changes?

automatic/cutepdf/tools/chocolateyInstall.ps1
Thilas: I understand your point of view but if the service is disabled back at the end of the install, cutepdf won't work. So what is the point to install it successfully if it doesn't work in the end?

An other idea would be to cancel the install and warn the user about the reason so that he fixes this but then it means that this package couldn't pass verifier service (since the service is disabled on this environment).

What do you think?

Thilas:
@AdmiringWorm, sorry I forgot. Could you have a look to my feedbacks.

automatic/cutepdf/tools/chocolateyInstall.ps1
AdmiringWorm: ah, I misunderstood the reason why it was being enabled.
I thought perhaps it just needed to be enabled to install, not to run.

Let me check out a few things regarding cutepdf before I merge this.

AdmiringWorm:
@Thilas 
> @AdmiringWorm, sorry I forgot. Could you have a look to my feedbacks?

No worries.
Before I merge this, I want to test out a few things regarding the CutePDF.
this will probably be merged either today or tomorrow.

AdmiringWorm:
seems like I'm done testing it now.
Thanks for contributing this.

AdmiringWorm:
Seems I need to wait until cutepdf v 3.1 is unrejected before your updates can be pushed.

Thilas:
No problem. Thanks.

AdmiringWorm:
@Thilas cutepdf v3.1 is now live.
Thanks again for fixing the spooler issue.

PR #508, opened 2017-01-02
(tor-browser) Migrate from Ketarin to AU

### LINKED ISSUE #506: (tor-browser) Not updating
It seems like the Ketarin script for tor-browser no longer is updating the package. The community feed is still at [6.0.5](https://chocolatey.org/packages/tor-browser) while current version is [6.0.8](https://www.torproject.org/download/download-easy.html.en#windows).

pascalberger:
Migrate tor-browser from Ketarin to AU. 

Fixes #506


AdmiringWorm:
@pascalberger 
so far it looks good to me, however we don't have the benefit of providing checksums.
can you please take a look at the firefox/thunderbird package and see if there is something there you maybe can reuse to provide checksums for all installers for this package?

pascalberger:
@AdmiringWorm Thanks for the review. Added missing checksum

AdmiringWorm:
@pascalberger I'll have a look again a little later, but one thing that I noticed that will not behave correctly is the checksum as it currently stands.
The checksum will only be valid for the installer if the culture is `en-US`, but will be wrong for any other culture.
You'll need to store the checksum for all the installers, or drop support for all cultures except `en-US`.

pascalberger:
@AdmiringWorm Good point! I'll thikn about a solution, maybe I try to change to an embedded package, which would solve the issue. Based on size of the packages

AdmiringWorm:
@pascalberger you can take a look at how I handled it in the [Firefox](https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/Firefox) package, I believe most of the helpers there can be reused for this package (with minimal modifications)

majkinetor:
Related: https://github.com/chocolatey/chocolatey-coreteampackages/issues/282

majkinetor:
@AdmiringWorm since this is hanging for too long, perhaps we make a package less robust in order to get the latest version. 

For example removing all languages except english ? Better to have latest english TOR then outdated multilang tor.

AdmiringWorm:
> For example removing all languages except english ? Better to have latest english TOR then outdated multilang tor.

yeah, probably just as well.

> @AdmiringWorm since this is hanging for too long, perhaps we make a package less robust in order to get the latest version.

It have been hanging because I've not really had any willpower to finish it up (I'm still looking for it).

AdmiringWorm:
I'll be going back on what I said.
Since the code for grabbing and storing different languages is already implemented, I just need to update the install script (mostly just that) to finish this one up.

AdmiringWorm:
@majkinetor I've updated this PR now with what I believe to be a working package.
I probably won't be touching this anymore now (unless for minor fixes).

PR #513, opened 2017-01-11
(youtube-dl) Changed to embedded package [AU youtube-dl]

AdmiringWorm:


automatic/youtube-dl/update.ps1
```diff
@@ -1,20 +1,36 @@
...
+    New-Item -ItemType Directory "$PSScriptRoot\tools"
+  }
+  $Latest.FileName = Get-WebFileName $Latest.URL32 'youtube-dl.exe'
+  $filePath = "$PSScriptRoot\tools\$($Latest.FileName)"
+  Get-WebFile $Latest.URL32 $filePath
+
+  # Let us check if the checksum actually matches
+  $actualChecksum = Get-FileHash -Algorithm $Latest.ChecksumType32 $filePath | % Hash
+
+  if ($actualChecksum -ne $Latest.Checksum32) {
```
pascalberger: @AdmiringWorm I'm a little bit confused about this one :) `$Latest.Checksum32` is set by `Get-WebFile`? In this case can this comparison every fail?

AdmiringWorm: it isn't set by `Get-WebFile`
`Get-WebFile` is a choco function to just download the file (you're thinking of `Get-RemoteFiles`).
the checksum is set using the provided checksum by the software developers.
So lets say something happens while downloading the file, but an error isn't reported the generated checksum wont match the provided checksum.

The check is probably almost unnecessary, but I feel it's better to have 1 check to many than 1 to few

AdmiringWorm: the software provided checksum is aquired here:
https://github.com/chocolatey/chocolatey-coreteampackages/pull/513/files#diff-9fff76ad80fcbf680f7450c0663434d4R44

since there wasn't anything else, I'll merge this now and push out the fix

pascalberger: But where in the script is `$Latest.Checksum32` set? Or is this handled by AU in some way?

PR #515, opened 2017-01-12
(visualstudiocode) Disable auto update by default

pascalberger:
Disable auto update by default. Added "EnableAutoUpdate" parameter for enable auto update

### DIFF
--- a/automatic/visualstudiocode/visualstudiocode.nuspec
+++ b/automatic/visualstudiocode/visualstudiocode.nuspec
@@ -40,6 +40,8 @@ For example: `--params '"/NoDesktopIcon /DontAddToPath"'`.
 
 ## Notes
 The package uses default install options except that it adds context menu entries and Visual Studio Code isn't started after installation.
+
+For disabling the auto-update functionality of Visual Studio Code see the [Visual Studio Code Auto Update Deactivation package](https://chocolatey.org/packages/visualstudiocode-disableautoupdate).
     </description>
     <tags>microsoft visualstudiocode vscode editor ide javascript typescript admin foss cross-platform</tags>
     <dependencies>
--- a/manual/visualstudiocode-disableautoupdate/tools/chocolateyInstall.ps1
+++ b/manual/visualstudiocode-disableautoupdate/tools/chocolateyInstall.ps1
@@ -1,0 +1,6 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$toolsPath = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+. $toolsPath\helpers.ps1
+
+Set-UpdateChannel("none")
--- a/manual/visualstudiocode-disableautoupdate/tools/chocolateyUninstall.ps1
+++ b/manual/visualstudiocode-disableautoupdate/tools/chocolateyUninstall.ps1
@@ -1,0 +1,6 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$toolsPath = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+. $toolsPath\helpers.ps1
+
+Set-UpdateChannel("stable")
--- a/manual/visualstudiocode-disableautoupdate/tools/helpers.ps1
+++ b/manual/visualstudiocode-disableautoupdate/tools/helpers.ps1
@@ -1,0 +1,79 @@
+function Set-UpdateChannel() {
+  param(
+    [Parameter(Mandatory=$true)]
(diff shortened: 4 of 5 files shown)


automatic/visualstudiocode/Tools/helpers.ps1
```diff
@@ -0,0 +1,44 @@
...
+            $storageFileContent.updateChannel = "none"
+            Write-Output "Updated 'updateChannel' to '$UpdateChannel'."
+        }
+        catch
+        {
+            Write-Output "Add new 'updateChannel' node with value '$UpdateChannel'."
+            $storageFileContent | Add-Member -Name "updateChannel" -value $UpdateChannel -MemberType NoteProperty
+        }
+
+        $storageFileContent | ConvertTo-Json -Depth 999 -Compress | Set-Content $storageFilePath
```
AdmiringWorm: using compress is wrong, I was just looking through the file and it is uncompressed

automatic/visualstudiocode/Tools/helpers.ps1
```diff
@@ -0,0 +1,44 @@
...
+{
+  "updateChannel": "$UpdateChannel"
+}
+"@
+
+        $storageFileContent | Set-Content $storageFilePath
+    } else {
+        Write-Output "Settings file '$storageFilePath' already exists."
+
+        $storageFileContent = ConvertFrom-Json "$(Get-Content $storageFilePath -Encoding UTF8)"
```
AdmiringWorm: You need to have a fallback for ConvertFrom-Json as I believe it isn't available in POSH v2

automatic/visualstudiocode/Tools/helpers.ps1
AdmiringWorm: also need to a fallback for ConvertTo-Json. I believe it isn't available in POSH v2

AdmiringWorm:
hold off for now regarding the json convertto/from, I'm currently testing this package on a windows 7 base environment to see if it will work or not.
if it doesn't work, you could try the following from stackoverflow:
http://stackoverflow.com/a/29689642

it requires .net 3.5, but since .net 4.5.2 will still be installed, it shouldn't matter that much.

AdmiringWorm:
@pascalberger so running this on a windows 7 vm throws an error.
relevant part:
```
2017-01-12 03:46:25,350 [INFO ] - Settings path 'C:\Users\vagrant\AppData\Roaming\Code' does not exist. Creating it now.
2017-01-12 03:46:25,374 [INFO ] - Settings file 'C:\Users\vagrant\AppData\Roaming\Code\storage.json' does not exist. Creating it now.
2017-01-12 03:46:25,387 [INFO ] - 
2017-01-12 03:46:25,387 [WARN ] - Type
2017-01-12 03:46:25,387 [WARN ] -   Confirmation (`-y`) is set.
  Respond within 30 seconds or the default selection will be chosen.
2017-01-12 03:46:52,106 [ERROR] - ERROR: Cannot process argument because the value of argument "type" is invalid. Change the value of the "type" argument and run the operation again.
```

ferventcoder:
I disagree with disabling updates by default. Let's not make folks insecure in the name of having the right package version linked.

pascalberger:
@ferventcoder Don't you want the functionallity at all or just keep auto update enabled by default and switch the parameter?

ferventcoder:
Switch the parameter yes. Allow me to decide to remove the default behavior by adding something additional.

pascalberger:
Thinking a little bit more about this, I would prefer to create a new manual package for just disabling auto update in Visual Studio Code instead of adding this functionallity as a parameter. Advantages of this would be that the disable-autoupdate package can be used as a dependency within another package for setting up an environment. And it keeps thethe Visual Studio Code package clean by not adding functionality to the package which is not part of their installer experience. And we can still mention the new package for disabling auto update in the package description of the Visual Studio Code package so that people easily can find it.

@AdmiringWorm @ferventcoder Any thoughts and preferences about this?

gep13:
@pascalberger I like the idea of a separate package.  Seems cleaner to me.

gep13:
Sorry, didn't mean to close the PR 😄

ferventcoder:
:+1: for separate package!

pascalberger:
Changed functionallity to disable auto-updater to its own package. 

@AdmiringWorm Should be ready for another review

AdmiringWorm:
@pascalberger I'll have a look when I get home (around 5-6 hours from now)

ferventcoder:
Looks good. I didn't check the PowerShell code in detail though.

PR #518, opened 2017-01-13
(waterfox) Migrated and updated

AdmiringWorm:
I've taken over the existing package [waterfox](https://chocolatey.org/packages/waterfox) and is updating this to an embedded package.

Unfortunately I'm currently unable to get the uninstall script to work as it should,
when trying to launch the NSIS uninstaller (and if not run silently) shows a messagebox
saying the file could not be opened (same thing happens when the auto uninstaller tries to run it).
Manually uninstalling it  and if the uninstall script is run outside of chocolatey works though, so basically I'm lost here.

EDIT:
I've opened a issue regarding the uninstaller here:
https://github.com/chocolatey/choco/issues/1135


automatic/waterfox/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,27 @@
+﻿$ErrorActionPreference = 'Stop';
+
+$packageName = 'waterfox'
+
+$uninstalled = $false
```
RedBaron2: This is not used anywhere else in the script

AdmiringWorm: I know it's not used anywhere, however it's part of the default template if you run `choco new`.
I don't really know if it's perhaps used somehow by choco or one of its helpers.

Maybe @ferventcoder can educate us on why it's in the default template?

RedBaron2: I will await the answer

PR #521, opened 2017-01-13
Flash Player PPAPI missed file

RedBaron2:
@gep13 @AdmiringWorm 
Missed file for name change


pascalberger:
Thanks @RedBaron2!

It seems like there's also an error in the AU script for this package as version couldn't be determined:

`Invalid version:`

See https://gist.github.com/gep13/bd2eaa76f2a9ab739ca0544c502dca6e#flashplayerppapi.

Maybe you can change this too in this PR?

pascalberger:
Thanks @RedBaron2

PR #523, opened 2017-01-13
(git.install) Update git.install after changes were pushed

pascalberger:
Update git.install after changes were pushed to chocolatey.org by AU but not commited to repository.

Or is there another way to have the files updated after a failing run?


AdmiringWorm:
@pascalberger 
as far as I know, there currently isn't a way other than manually committing to repository.

FYI, you don't need PR for this, just commit it directly

pascalberger:
@AdmiringWorm Thanks! Was just not sure if this is the correct way to fix this, therefore prefered a PR over directly commiting :)

PR #526, opened 2017-01-14
(git.install) Revert registry detection logic back as it originally was

### LINKED ISSUE #524: (git.install) Registry key error on install
I'm getting an error "The registry key at the specified path does not exist." when attempting to install the git package (which installs git.install).

Here's the gist:
https://gist.github.com/michaelray/0796d6c6f7efadf31d51315cfd733713

pascalberger:
Revert logic of registry key detection back as it was before migration to core team repo and refactorings

Fixes #524


AdmiringWorm:
@pascalberger I've merged this now, but I need to look into padding the revision number before pushing this out.

AdmiringWorm:
I noticed a new release have been released.
This will be pushed the next time the updater is run.

AdmiringWorm:
However, we should probably try to have issue #499 included before then.

PR #527, opened 2017-01-14
(iTunes) Update to latest version and fix install errors

lennartb-:
This fixes the invalid cast exception from Object[] to Version[], adds a link to the 32 bit version and points both 32 and 64 bit to the latest version

### DIFF
--- a/automatic/iTunes/tools/chocolateyInstall.ps1
+++ b/automatic/iTunes/tools/chocolateyInstall.ps1
@@ -14,12 +14,12 @@ $packageArgs = @{
   checksumType64 = 'sha256'
   silentArgs     = "/qn /norestart"
   validExitCodes = @(0, 2010, 1641)
-  unzipLocation = Get-PackageCacheLocation
+  unzipLocation  = Get-PackageCacheLocation
 }
 
 $app = Get-UninstallRegistryKey -SoftwareName $packageArgs.softwareName | select -first 1
 
-if ($app -and ([version]$app.Version -ge [version]$version)) {
+if ($app -and ([version]$app.DisplayVersion -ge [version]$version) -and ($env:ChocolateyForce -ne $true)) {
   Write-Host "iTunes $version or higher is already installed."
   Write-Host "No need to download and install again"
   return;
@@ -37,4 +37,4 @@ foreach ($msiFile in $msiFileList) {
   Install-ChocolateyInstallPackage @packageArgs
 }
 
-Remove-Item $packageArgs.unzipLocation -Recurse
+Remove-Item $packageArgs.unzipLocation -Recurse -Force -ea 0
--- a/automatic/iTunes/update.ps1
+++ b/automatic/iTunes/update.ps1
@@ -1,4 +1,8 @@
+<<<<<<< HEAD
 ﻿Import-Module AU
+=======
+Import-Module AU
+>>>>>>> (itunes) Added AU update script
 
 $releases     = 'https://www.apple.com/itunes/download/'
 $softwareName = 'iTunes'
--- a/automatic/iTunes/update.ps1
+++ b/automatic/iTunes/update.ps1
@@ -1,8 +1,4 @@
-<<<<<<< HEAD
 ﻿Import-Module AU
-=======
-Import-Module AU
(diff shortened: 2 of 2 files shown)


gep13:
@lennartb- thank you for taking the time to update this, however, there would still need to be some work done here before we could merge it in. iTunes is actually currently what we call a ketatin automatic package, and we are in the process of moving these sorts of packages to be AU packages. You have done part of the work to get this package there. Would you be interested in doing the rest?

lennartb-:
Sure, are there docs on how to do this?

pascalberger:
@lennartb- Ketarin was the old tool used. Now we're in the process to rewrite everything to a new tool (AU). You can find documentationo about it [here](https://github.com/majkinetor/au).

There are also quite a few package in this repo which already use AU, which you can use as a sample.

And please also check our [contribution guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines).

And if you have any specific questions just let us know :)

lennartb-:
What happens if there isn't a direct download to the required installer? I poked around the iTunes download page and it seems the URL is constructed via JS. The actual link the permanent, but there is no *.exe file linked (as in the sample au-packages-template project). I look around the existing packages, but those who had an update.ps1 mostly contained fixed links to the installers. I took a look with Fiddler, but I couldn't figure out the actual download request.

AdmiringWorm:
@lennartb- if there is no way to automate the update, then this would have to be moved to the manual directory and manually be updated when it becomes out of date.

lennartb-:
Maybe someone more savvy with JS can take a look at it then, I (hopefully) can manage the rest, but I'm not much of a web tech guy.

AdmiringWorm:
@lennartb- I can add a update script for this before it's merged (I believe I found a way to get the version and urls), if you could update the rest it would be great.

AdmiringWorm:
@lennartb- can you also rebase your changes please

lennartb-:
@AdmiringWorm Yep, done

AdmiringWorm:
I've added the AU update script, updated the install script and the metadata.
Please review this one again.

/cc @gep13 @pascalberger

automatic/iTunes/iTunes.nuspec
```diff
@@ -1,28 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
-<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
-    <id>{{PackageName}}</id>
-    <version>{{PackageVersion}}</version>
-    <title>iTunes</title>
+    <id>iTunes</id>
```
pascalberger: ID should be in lowercase (which would also include directory name)

automatic/iTunes/tools/chocolateyInstall.ps1
```diff
@@ -1,37 +1,40 @@
...
+  $_.Name -notmatch 'AppleSoftwareUpdate*.msi'
+})
+
+foreach ($msiFile in $msiFileList) {
+  $packageArgs.packageName = $msiFile.Name
+  $packageArgs.file = $msiFile.FullName
+  Install-ChocolateyInstallPackage @packageArgs
+}
+
+Remove-Item $packageArgs.unzipLocation -Recurse
```
pascalberger: Ignore errors here?

automatic/iTunes/iTunes.nuspec
AdmiringWorm: true, it should.
But until https://github.com/chocolatey/choco/issues/1040 is fixed, it shouldn't be changed for existing packages.

automatic/iTunes/tools/chocolateyInstall.ps1
AdmiringWorm: will do

PR #529, opened 2017-01-14
(all) Change encoding of all .nuspec files to UTF8

pascalberger:
Change encoding of all .nuspec files to UTF8 (without BOM) and added same test characters to all files


AdmiringWorm:
until the PR I have open for AU, the encodings this PR fixes will be reverted when the package is updated.

https://github.com/majkinetor/au/pull/63

PR #533, opened 2017-01-15
(waterfox) Update nuspec to be more compliant with guidelines

pascalberger:


automatic/waterfox/waterfox.nuspec
```diff
@@ -14,19 +14,16 @@
     <licenseUrl>https://github.com/MrAlex94/Waterfox/blob/master/LICENSE.md</licenseUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <projectSourceUrl>https://github.com/MrAlex94/Waterfox</projectSourceUrl>
-    <tags>waterfox admin web-browser firefox 64-bit</tags>
+    <tags>waterfox web-browser firefox 64-bit admin cross-platform foss</tags>
     <summary>Waterfox is basically a 64-bit version of Firefox.</summary>
     <description>
-## How and why Waterfox is here...
```
AdmiringWorm: why removing this?
It's been added because it's part of their about page.

automatic/waterfox/waterfox.nuspec
```diff
@@ -14,19 +14,16 @@
...
-I had a fascination for the web and wanted to help expand on the ideals of what Mozilla had for a free and open web.
-And so I decided to make Waterfox, a 64-Bit browser based on Mozilla's free and open source platform.
+Waterfox was started back in March 2011 by [Alex Kontos](https://github.com/MrAlex94), a 16 year old student.
+He had a fascination for the web and wanted to help expand on the ideals of what Mozilla had for a free and open web.
+And so he decided to make Waterfox, a 64-Bit browser based on Mozilla's free and open source platform.
 
 Waterfox was one of the first widely distributed 64-Bit browsers on the web and quickly gained a loyal following.
 At a time Waterfox had one thing in mind: speed, but now Waterfox also attempts to be an ethical, user-oriented browser
 
-![](https://www.waterfoxproject.org/img/storm-speed.png)
```
AdmiringWorm: same question as above.

automatic/waterfox/waterfox.nuspec
```diff
@@ -4,8 +4,8 @@
   <metadata>
     <id>waterfox</id>
     <version>50.1.0</version>
-    <packageSourceUrl>https://github.com/AdmiringWorm/chocolatey-packages/tree/master/waterfox</packageSourceUrl>
```
AdmiringWorm: thanks for fixing this, was originally planning to add it to my repository.
Guess I forgot to change it.

automatic/waterfox/waterfox.nuspec
pascalberger: Same as above (here it looks even more out of place, as alignment is different than on the waterfox page, and also background color doesn't match)

automatic/waterfox/waterfox.nuspec
pascalberger: IMHO this looks strange on the chocolatey.org page. 

Also there a lot of other packages which have differences to the about page of the project (eg. even here you already stripped the "About Waterfox" heading). 
But if you want to keep it I will revert this change.

automatic/waterfox/waterfox.nuspec
AdmiringWorm: ah yes, I didn't notice that before.
Skip this one then.

automatic/waterfox/waterfox.nuspec
AdmiringWorm: Personally I don't find it strange at all.
Yeah, I stripped out `About Waterfox` heading beacuse that's what the description is all about, and that is the main header while this one is a sub header.
I would like to keep it.

PR #541, opened 2017-01-20
clover: checksum incorrect

jmbeach:


AdmiringWorm:
@jmbeach I'm afraid that PRs which only changes the checksum and nothing else usually isn't accepted here.
The checksum is handled by AU (Chocolatey Automatic Package Updater Module) and will automatically be replaced when a new build run is initiated (in this case a forced run is needed).

I'll initiate a new forced update run on this package quite soon.

jmbeach:
Gotcha. Didn't know that, sorry.

jmbeach:
@AdmiringWorm, you merged my commit. Was that on accident?

AdmiringWorm:
@jmbeach nah, it wasn't on accident.
I just used your PR to force the update of the clover package instead of pushing it manually from my local computer.

PR #542, opened 2017-01-23
flashplayeractivex : replace outdated download url with the new one

Pyrrvs:
The base url used to download the flashplayer has changed from :
_https://download.macromedia.com/pub/flashplayer/pdc/${CurrentVersion}/_
to
_https://fpdownload.macromedia.com/pub/flashplayer/pdc/${CurrentVersion}/_

### DIFF
--- a/automatic/flashplayeractivex/tools/chocolateyInstall.ps1
+++ b/automatic/flashplayeractivex/tools/chocolateyInstall.ps1
@@ -29,10 +29,10 @@ if ($allRight) {
 $packageArgs = @{
   packageName   = $packageName
   fileType      = 'msi'
-  url           = 'https://download.macromedia.com/get/flashplayer/current/licensing/win/install_flash_player_24_active_x.msi'
+  url           = 'https://fpdownload.macromedia.com/pub/flashplayer/pdc/24.0.0.194/install_flash_player_24_plugin.msi'
   silentArgs    = '/quiet /norestart REMOVE_PREVIOUS=YES'
   softwareName  = 'Adobe Flash Player ActiveX'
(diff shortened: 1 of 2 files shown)


AdmiringWorm:
@Pyrrvs 
Actually it was changed from:
`https://download.macromedia.com/get/flashplayer/current/licensing/win/install_flash_player_<majorVersion>_active_x.msi`
to
`https://download.macromedia.com/pub/flashplayer/pdc/${CurrentVersion}/install_flash_player_${majorVersion}_plugin.msi`

The changes however was never pushed because of the upstream version scheme.
This is also why the url and checksum haven't been changed in the install script.

I've been planning to look into being able to update the flashplayer packages, but unfortunately have yet to get around to it.
The url you have used in this PR is to the round robin servers,
as such I'm afraid I'll have to close this PR

Pyrrvs:
Thank you for this explanation and sorry if I've bothered you with this PR. I would have a question however about the 'upstream version scheme' you were talking about because I'm not quite sure to understand what you mean by that. So if you have a moment to explain it to me, I might be able to help you in the future.

Regards

AdmiringWorm:
Don't worry about bothering us with the PR.

What I mean by upstream versioning scheme, is that Adobe releases flashplayer using a 4-part version.
i.e `24.0.0.194`, and since chocolatey only supports maximum 4 parts we are unable to update these packages without changing this version.

I've added a PR just now that adds a script we can use while updated such packages by padding the revision number (the last part of a 4-part versioning) so we can update all packages making use of that script.

Pyrrvs:
Thank you, it makes much more sense now !

PR #543, opened 2017-01-23
(build) Pad version script

AdmiringWorm:
I've added a script to allow for easier access to padding the revision number of packages
which already uses 4-part version scheme.
The script also checks if the global variable 'au_force' is set to true and then increases the revision number based on what is shown as the latest in the corresponding nuspec file.

Also added to the flashplayer* packages to use those scripts (plans to add it to autoit packages as well).

I know I've not added any documentation of the script, and honestly didn't want to use time on it yet either.

/cc @pascalberger @gep13 @ferventcoder

### DIFF
--- a/automatic/flashplayeractivex/update.ps1
+++ b/automatic/flashplayeractivex/update.ps1
@@ -1,7 +1,9 @@
-
+﻿
 import-module au
+. "$PSScriptRoot\..\..\scripts\Get-Padded-Version.ps1"
 
 $releases = "https://get.adobe.com/en/flashplayer/" # URL to for GetLatest
+$padVersionUnder = '24.0.1'
 
 function global:au_BeforeUpdate {
   # We need this, otherwise the checksum won't get created
@@ -13,7 +15,7 @@ function global:au_BeforeUpdate {
 function global:au_SearchReplace {
   @{
     ".\tools\chocolateyInstall.ps1" = @{
-      "(^[$]version\s*=\s*)('.*')"= "`$1'$($Latest.Version)'"
+      "(^[$]version\s*=\s*)('.*')"= "`$1'$($Latest.RemoteVersion)'"
       "(^[$]majorVersion\s*=\s*)('.*')"= "`$1'$($Latest.majorVersion)'"
       "(^[$]packageName\s*=\s*)('.*')"= "`$1'$($Latest.PackageName)'"
       "(?i)(^\s*url\s*=\s*)('.*')" = "`$1'$($Latest.URL32)'"
@@ -35,7 +37,9 @@ function global:au_GetLatest {
 
   $url32 = "https://download.macromedia.com/pub/flashplayer/pdc/${CurrentVersion}/install_flash_player_${majorVersion}_plugin.msi"
 
-  return @{ URL32 = $url32; Version = $CurrentVersion; majorVersion = $majorVersion; }
+  $packageVersion = Get-Padded-Version $CurrentVersion $padVersionUnder
+
+  return @{ URL32 = $url32; Version = $packageVersion; RemoteVersion = $CurrentVersion; majorVersion = $majorVersion; }
 }
 
-update -ChecksumFor 32
+update -ChecksumFor none
--- a/scripts/Get-Padded-Version.ps1
+++ b/scripts/Get-Padded-Version.ps1
@@ -1,9 +1,47 @@
+<#
+.SYNOPSIS
+  Pad the revision number if the passed version is using a 4-part version scheme.
+
+.DESCRIPTION
+  Adds padded zeroes to the revision part of the version if the passed
+  Version parameter is using a 4-part scheme and the Parameter OnlyBelowVersion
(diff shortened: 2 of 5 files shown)


AdmiringWorm:
@gep13 what do you mean?
The maximum version passed to the `Get-Padded-Version` script?

IMO a padded version should only be used when it's needed, in such a case the maximum version should be increased whenever there is a need for it. (Update the maximum version in `update.ps1` of the package).

For instance, we need to fix a package that has the version of `24.0.0.x`, setting the maximum padded version (exclusive) to `24.0.1` would mean only versions below `24.0.1` will be padded, but not after.

There is also the possibility to not pass a maximum version which means all versions using a 4-part versioning scheme will be padded.

AdmiringWorm:
I'll honestly say that I could probably come up with a better parameter name than `MaximumVersion`, but couldn't think of one.

automatic/flashplayeractivex/update.ps1
```diff
@@ -1,7 +1,9 @@
-
+﻿
 import-module au
+. "$PSScriptRoot\..\..\scripts\Get-Padded-Version.ps1"
 
 $releases = "https://get.adobe.com/en/flashplayer/" # URL to for GetLatest
+$padVersionUnder = '24.0.1'
```
gep13: @AdmiringWorm the question was really about this.  This is currently set at 24.0.1.  Who or what is responsible for changing this, once the maximum version number allowed/possible goes above this?

AdmiringWorm: This would be needed to be changed manually (by whoever fixes the package) when/if the last released package is equal to or above that version.

AdmiringWorm: I can of course remove this, which would mean that every version that is a 4-part version would be padded with zeros.

AdmiringWorm:
@gep13 
I've added some basic documentation with examples that hopefully makes a little sense.

automatic/flashplayeractivex/update.ps1
gep13: @AdmiringWorm nope, I think this makes sense.  I just wanted to make sure that I wasn't missing something.  The docs you have added really help with understanding.

ferventcoder: I understood it and haven't read the docs just yet.

scripts/Get-Padded-Version.ps1
```diff
@@ -0,0 +1,79 @@
...
+    return $Version
+  } elseif ($Version -notmatch "^[\d\.]+$") {
+    Write-Warning "Padding of pre release versions is not supported";
+    return $Version
+  }
+
+  $Matches = $null
+
+  if ($OnlyBelowVersion -ne $null -and ([version]$Version) -ge $OnlyBelowVersion) {
+    return $Version;
```
ferventcoder: No write debug here like as above?

scripts/Get-Padded-Version.ps1
```diff
@@ -0,0 +1,79 @@
...
+#>
+
+function Get-Padded-Version() {
+  param(
+    [Parameter(Mandatory = $true, Position = 0)]
+    [string]$Version,
+    [Parameter(Position = 1)]
+    [version]$OnlyBelowVersion = $null,
+    [Parameter(Position = 2)]
+    [int]$RevisionLength = 6
```
ferventcoder: Why 6 as a default?

scripts/Get-Padded-Version.ps1
```diff
@@ -0,0 +1,79 @@
...
+.EXAMPLE
+  Get-Padded-Version -Version '24.0.1.2' -OnlyBelowVersion '24.0.1'
+  Will output the same version that was given without padded zeroes.
+#>
+
+function Get-Padded-Version() {
+  param(
+    [Parameter(Mandatory = $true, Position = 0)]
+    [string]$Version,
+    [Parameter(Position = 1)]
```
ferventcoder: Not a big fan of positional parameters. Always best to positional the first one or two items (especially if they are mandatory), and leave the rest for named parameters. cc @gep13 ?

scripts/Get-Padded-Version.ps1
AdmiringWorm: I just forgot to add one.

scripts/Get-Padded-Version.ps1
AdmiringWorm: No particular reason, I basically picked a random number between 4 and 10.
6 Seemed like a good fit though, IMO

scripts/Get-Padded-Version.ps1
AdmiringWorm: I felt that the OnlyBelowVersion parameter would probably be used most of the time,
as such I believe it should be a positional parameter.
Not sure why I put the RevisionLength parameter as a positional though.

AdmiringWorm:
@ferventcoder I've added the missing `Write-Debug` and also removed the Position from the RevisionLength.

PR #552, opened 2017-01-31
(picard) Migrated Ketarin to AU

AdmiringWorm:
## Changes
- Update the install script to be a little more modern
- Added uninstall script since NSIS isn't uninstalling silently by default
- Updated all missing metadata + added features section
- Removed ketarin file
- Added AU update script

**NOTE: *Requires [AU picard] when merging***


automatic/picard/picard.nuspec
```diff
@@ -19,14 +27,25 @@ Picard supports the majority of audio file formats, is capable of using audio fi
 When tagging files, Picard uses an album-oriented approach. This approach allows it to utilize the MusicBrainz data as effectively as possible and correctly tag your music. For more information, see the illustrated quick start guide to tagging.
 
 Picard is named after Captain Jean-Luc Picard from the TV series Star Trek: The Next Generation.
+
+## Features
+- Multiple Formats
+  Picard supports all popular music formats, including MP3, FLAC, OGG, M4A, WMA, WAV, and more.
```
pascalberger: This won't look good if rendered, since the line break will be removed. I would suggest something like:

```
* **Multiple Formats:** Picard supports all popular music formats, including MP3, FLAC, OGG, M4A, WMA, WAV, and more.
```

AdmiringWorm: That is actually not an issue, unlike the initial commonmark spec, chocolatey actually uses hard break lines instead of soft breaks.

If you look for instance of the [waterfox](https://chocolatey.org/packages/waterfox) package, you see there is a line break in after each dot.
However, I have not added two spaces after them since that isn't necessary.
The same thing is happening in the notes section list of that package.

AdmiringWorm: however, I should probably make the first line of each list item bold anyhow.

pascalberger: OK, didn't know that chocolatey uses hard breaks. The suggested format is the one which for example is used in the SourceTree package.

AdmiringWorm: well, I can certainly change it to that format though.

AdmiringWorm:
@pascalberger I've updated the list formatting to what you suggested.

pascalberger:
@AdmiringWorm LGTM

PR #561, opened 2017-02-02
(qtox) Update AU script and change to an embedded package

pascalberger:
Fixes the error in the AU script and change to an embedded package

### DIFF
--- a/automatic/qtox/qtox.nuspec
+++ b/automatic/qtox/qtox.nuspec
@@ -4,7 +4,7 @@
   <metadata>
     <id>qtox</id>
     <title>qTox</title>
-    <version>1.4.1.1</version>
+    <version>1.7.0</version>
     <authors>qTox developers</authors>
     <owners>chocolatey</owners>
     <summary>Powerful Tox chat client that follows the Tox design guidelines.</summary>
@@ -46,6 +46,7 @@ This package will always install the latest version of qTox, regardless of the p
     <releaseNotes>https://github.com/qTox/qTox/blob/master/CHANGELOG.md</releaseNotes>
   </metadata>
   <files>
+    <file src="legal\**" target="legal" />
     <file src="tools\**" target="tools" />
   </files>
 </package>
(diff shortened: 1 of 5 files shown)


automatic/qtox/qtox.nuspec
```diff
@@ -4,7 +4,7 @@
   <metadata>
     <id>qtox</id>
     <title>qTox</title>
-    <version>1.4.1.1</version>
+    <version>1.7.0</version>
```
AdmiringWorm: please don't change the version to anything newer than what is available on chocolatey.org.
Latest version on chocolatey.org is `1.5.1`

AdmiringWorm:
also had one question, what if the changelog have been updated but the executable haven't yet been built?
We would then package the wrong version.

automatic/qtox/qtox.nuspec
pascalberger: Thanks, missed to reset after last test :). Have it now set to 0.0

pascalberger:
@AdmiringWorm Yes, thats a little bit ugly with determing the version from a different source than the binaries, but unfortunately I couldn't find a better solution. They don't seem to push the Windows Installer any longer to the GitHub releases. Happy to change that if you have a better idea :)

AdmiringWorm:
@pascalberger hmm, perhaps after downloading the executable check the version if it?
If it isn't somewhat equal to the expected version, throw an exception.

automatic/qtox/update.ps1
```diff
@@ -1,34 +1,72 @@
 ﻿import-module au
```
AdmiringWorm: We need an import of the chocolateyInstaller.psm1 file here.
i.e.
`Import-Module "$env:ChocolateyInstall\helpers\chocolateyInstaller.psm1"`

pascalberger:
@AdmiringWorm The version of the downloaded installer unfortunately doesn't correspond to the version of the application (seems to always be at 1.0.0)

AdmiringWorm:
@pascalberger I see, so that's a no go then.
Too bad. well, after you've added the missing import module, we'll just merge this.

automatic/qtox/update.ps1
pascalberger: Ah yes. Added it

pascalberger:
@AdmiringWorm Added the missing import

PR #572, opened 2017-02-07
(yo) Add package for Yeoman CLI

pascalberger:
Add new package for Yeoman CLI. This should also deprecate existing [Yeoman package](https://chocolatey.org/packages/Yeoman)

PR for deprecation of current `Yeoman` package: https://github.com/ferventcoder/chocolatey-packages/pull/231


automatic/yo/yo.nuspec
```diff
@@ -0,0 +1,41 @@
...
+    <bugTrackerUrl>https://github.com/yeoman/yo/issues</bugTrackerUrl>
+    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/yo</packageSourceUrl>
+    <tags>nodejs node javascript web foss cross-platform cli</tags>
+    <licenseUrl>https://opensource.org/licenses/bsd-license.php/</licenseUrl>
+    <copyright>© Google</copyright>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <iconUrl>https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/9e732d13804ce0097f9c52c080241edc560acbf9/icons/yo.png</iconUrl>
+    <releaseNotes>https://github.com/yeoman/yo/releases</releaseNotes>
+    <dependencies>
+      <dependency id="nodejs" version="0.12.0" />
```
gep13: @pascalberger would a dependency on this package: https://chocolatey.org/packages/nodejs-lts be a better option?

pascalberger: @gep13 This would force you to use the LTS version of node. IMHO it should be possible to use the package with either version, so maybe no dependency at all is the best option?

gep13: This is a place where a full virtual package would be ideal, i.e. `node` which would be satisfied by either nodejs or nodejs-lts.  For now, I think what you have is fine.  Just wanted to ask the question.

PR #578, opened 2017-02-11
(clementine) Changed to embedded package

AdmiringWorm:
Feel free to squash all commits on merge.

### DIFF
--- a/automatic/clementine/clementine.nuspec
+++ b/automatic/clementine/clementine.nuspec
@@ -1,17 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
-<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>clementine</id>
     <version>1.3.1</version>
     <title>Clementine</title>
-    <tags>music mp3 audio media foss cross-platform podcast ogg flac transcode player admin</tags>
+    <tags>clementine music mp3 audio media foss cross-platform podcast ogg flac transcode player admin</tags>
     <owners>chocolatey</owners>
     <authors>David Sansome, John Maguire and Arnaud Bienner</authors>
     <licenseUrl>https://github.com/clementine-player/Clementine/blob/master/COPYING</licenseUrl>
     <projectUrl>http://www.clementine-player.org/</projectUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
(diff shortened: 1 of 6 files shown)


automatic/clementine/tools/chocolateyInstall.ps1
```diff
@@ -1,30 +1,27 @@
-$ErrorActionPreference = 'Stop'
+﻿$ErrorActionPreference = 'Stop';
```
pascalberger: Why adding a semicolone? Since there isn't a semicolone used anywhere else IMHO one shouldn't be used here either

automatic/clementine/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,32 @@
+﻿$ErrorActionPreference = 'Stop';
```
pascalberger: Why semicolone?

automatic/clementine/update.ps1
```diff
@@ -1,23 +1,39 @@
-import-module au
+﻿Import-Module AU
 
-$releases = 'https://www.clementine-player.org/downloads'
+$releases     = 'https://www.clementine-player.org/downloads'
```
pascalberger: Remove leading spaces in front of `=`

automatic/clementine/clementine.nuspec
```diff
@@ -1,22 +1,43 @@
...
+- Fetch missing tags from MusicBrainz.
+- Discover and download Podcasts.
+- Download missing album cover art from Last.fm and Amazon.
+- Remote control using and Android device, a Wii Remote, MPRIS or the command-line.
+- Copy music to your iPod, iPhone, MTP or mass-storage USB player.
+- Queue manager.
     </description>
     <dependencies>
-      <dependency id="chocolatey-core.extension" version="1.0.4" />
+      <dependency id="chocolatey-core.extension" version="1.0.6" />
```
pascalberger: Is there any functionally used which is not part of 1.0.4?

AdmiringWorm: no there isn't, however version 1.0.6 resolves a bug in the `Get-AppInstallLocation` where install would fail if the path wasn't found (on POSH 2)

AdmiringWorm: IMO, when updating a package, dependencies should be updated to the latest patch/bug release.

automatic/clementine/tools/chocolateyInstall.ps1
AdmiringWorm: it was just automatically added, I'll remove it. I'll do the same with the other one.

automatic/clementine/clementine.nuspec
pascalberger: OK, makes sense. I don't think there ever will be a package locking to an exact version or max version of `chocolatey-core.extension` which would lead to conflicts.

AdmiringWorm:
@pascalberger I believe I've fixed the formatting issues?

pascalberger:
@AdmiringWorm Thanks, LGTM

PR #579, opened 2017-02-12
Mendeley

Redsandro:


Redsandro:
I'm oldskool (ketarin) and `mendeley` is my first `au` package. It works standalone, but I was wondering if you could (tell me how to) test if this works with the global updater and add it to the update computer @admiringworm @pascalberger.

pascalberger:
@Redsandro For the future please wait for a review before merging your own PR, especially if you have open questions. There are several things not conform to guidelines, and there is already a package with this name (but rejected), therefore this cannot be pushed through the Chocolatey user.

pascalberger:
I made some fixes with https://github.com/chocolatey/chocolatey-coreteampackages/commit/ab988d6acf24f847ceb29f9515f052e5aec26b43

@AdmiringWorm Any idea how to avoid the URL in the AU script? I couldn't find a page on their website which would allow to get the full download link.

AdmiringWorm:
@pascalberger I'm not sure I follow.
Why would you avoid a URL? Even if the website doesn't provide the full url a url would still be required to download the software.

pascalberger:
@AdmiringWorm Sorry I was unclear :) I don't want to avoid the URL, but scraping it from the website instead of hardcoding it in the script would make the script much more stable.

Redsandro:
@pascalberger said
> For the future please wait for a review before merging your own PR

Yes, absolutely. Back when I committed most packages was in a time when this repo was grossly understaffed and getting something done would easily take months. So I'm used to trying not to bother people. But I hesitated because this system is new to me. I didn't expect you to reply so fast! Lesson learned.

> I don't want to avoid the URL, but scraping it from the website instead of hardcoding it in the script would make the script much more stable.

Agreed, but I see no way to do that. The URL is constructed by scripts; the `au` way doesn't work.

> @pascalberger added the Requires Chocolatey User

I added the chocolatey user. What's the policy about maintainers that don't fix their package in over a year? Remove the old maintainer?

pascalberger:
> I added the chocolatey user. What's the policy about maintainers that don't fix their package in over a year? Remove the old maintainer?

@Redsandro I asked this in [Gitter](https://gitter.im/chocolatey?at=58a0d2c000c00c3d4f2d270b). You can remove the old maintainer in this case.

Redsandro:
#done

Redsandro:
Does the update machine automatically import the master branch?

gep13:
>@Redsandro said...
Does the update machine automatically import the master branch?

By update machine, are you referring to the Azure VM which runs the ketarin instance that we still have?  If so, this plays no part in the running of AU.  The AU packages are run on AppVeyor.  Once all the ketarin packages are migrated to AU we will not need that VM at all, and it will all be done through AppVeyor.

pascalberger:
@Redsandro There's actually not really an update machine involved, like with Ketarin. There are just AppVeyor builds, which are triggered to run scheduled every few hours, or of course after a committ/merge on master.

AdmiringWorm:
@pascalberger said:
> @AdmiringWorm Sorry I was unclear :) I don't want to avoid the URL, but scraping it from the website instead of hardcoding it in the script would make the script much more stable.

Ah, I see. In that case I don't know.
Don't really think it matters that much that the url is hard coded in the update script though, since au tests the url for 404 on each build run.

Redsandro:
@gep13 @pascalberger Gotcha. Thanks.

pascalberger:
@Redsandro FYI: AU now pushed the package to Chocolatey and committed the changes back (see https://github.com/chocolatey/chocolatey-coreteampackages/compare/2e50444e1b...0c3478c9e2). You can check the status of the latest AU run also [here](https://gist.github.com/gep13/bd2eaa76f2a9ab739ca0544c502dca6e)

Redsandro:
@pascalberger I wrote a hotfix for the 1223 exit code I received after (succesful) installation. https://github.com/chocolatey/chocolatey-coreteampackages/commit/3d43aa87127b46f5e6efb4350a29c844d7d3164a

Is this where I should have used `AU` in the commit message?

gep13:
@Redsandro yes, you would have needed an AU commit message to force the pushing of the package, otherwise, it will have to wait until the next release of the package.

@AdmiringWorm @pascalberger I think in this case we should use a package fix notation, assuming the package has made it through moderation (I haven't checked yet)?  Thoughts?

AdmiringWorm:
@gep13  it didn't make it through moderation, it fails with error code 1223.
Been trying to figure out why, but no luck. Even fails in the test environment when running with `--notsilent` and it dosn't look like it is missing any dependencies either since it is installed and runs.

AdmiringWorm:
@Redsandro btw, in this case you can force the update of this package in two different ways.
1. Set the version to `0.0` and commit with the `[AU mendeley]` in the title (it will then automatically set the version to the latest upstream).
2. Or add an empty commit message with `[AU mendeley:1.17.7]`, this will force au to use the specified version as the package version no matter what the upstream version is.

Personally I prefer the first option.

Redsandro:
@AdmiringWorm said:
> Set the version to 0.0 and commit with the [AU mendeley] in the title 

Done: https://github.com/chocolatey/chocolatey-coreteampackages/commit/b9c88d9e7ae43efa8daa419a0b7bc8c7ef81abe4

> @gep13 it didn't make it through moderation, it fails with error code 1223.

Yes, hence the hotfix. Some NSIS packages do that sometimes, but they install correctly.

Redsandro:
The package still fails. I don't get it.

```
ERROR: Path doesn't exist: C:\Program Files\Mendeley Desktop\mendeley.exe
The install of mendeley was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\mendeley\tools\chocolateyInstall.ps1'.
 See log for details.
```

Note that `mendeley` is installed and works fine, but the package fails.

AdmiringWorm:
@Redsandro that's no wonder, the actual path would be `C:\Program Files (x86)\Mendeley Desktop\MendeleyDesktop.exe`

Redsandro:
@AdmiringWorm no that's fine. I'm running x86 (so `"Program Files"` is correct, on your machine it will be different.)

The problem is that `chocolatey-core\Register-Application.ps1` is looking for `mendeley.exe`, which doesn't exist. What activates Register-Application? What is it for?

```
ERROR: Path doesn't exist: C:\Program Files\Mendeley Desktop\mendeley.exe
 at Register-Application, C:\ProgramData\chocolatey\extensions\chocolatey-core\Register-Application.ps1: line 40
at <ScriptBlock>, C:\ProgramData\chocolatey\lib\mendeley\tools\chocolateyInstall.ps1: line 25
at <ScriptBlock>, C:\ProgramData\chocolatey\helpers\chocolateyScriptRunner.ps1: line 48
at <ScriptBlock>, <No file>: line 1
```

Redsandro:
I'm an idiot. Stand by.

AdmiringWorm:
@Redsandro 
Sorry, I did not mean the `Program Files` part of the path, but the name of the executable which is `MendeleyDesktop.exe` not `mendeley.exe`.

From my understanding, `Register-Application` main job is to create a registry key to make the executable available in the Windows Run dialog.

Redsandro:
I removed the registration and let the installer handle it. It adds Mendeley to start.
https://github.com/chocolatey/chocolatey-coreteampackages/commit/c2d5e4807f111afe03740b1a34a8af9ecbf81ad8

Sorry for crying wolf.

AdmiringWorm:
@Redsandro adding to start, and adding to run dialog is not the same.
The run dialog I'm talking about is what you get when you press `Win+R`.

AdmiringWorm:
I would rather change the `Register-Application` call to `Register-Application "$installLocation\MendeleyDesktop.exe" -Name $packageName` instead.

Redsandro:
Ah - Personally I don't care for that (adding non-default features without opt-in). Run dialog is more for admin stuff that isn't in the start menu. And I can't imagine any of its millions of academic users care, since they've always entered it in the start menu, not the run dialog.

I just want it to install equal to a manual install, without errors, with the least amount of customizations. Feel free to add extra features. Know that a hardcoded .exe might break when they decide to change it again (e.g. `MendeleyClient.exe`)

Redsandro:
Package works, finally.

PR #581, opened 2017-02-13
(git-disable-gcm) Add package for disabling Git Credential Manager

pascalberger:
Since the Git installer doesn't provide an option for not installing the GCM, I created a separate package which would disable GCM.

Disabling GCM while installing Git was originally requested in https://github.com/ferventcoder/chocolatey-packages/issues/149.

### DIFF
--- a/manual/git-disable-gcm/tools/chocolateyInstall.ps1
+++ b/manual/git-disable-gcm/tools/chocolateyInstall.ps1
@@ -1,0 +1,5 @@
+﻿$ErrorActionPreference = 'Stop'
+
+Update-SessionEnvironment
+
+git config --system --unset credential.helper
--- a/manual/git-disable-gcm/tools/chocolateyInstall.ps1
+++ b/manual/git-disable-gcm/tools/chocolateyInstall.ps1
@@ -1,5 +1,7 @@
 ﻿$ErrorActionPreference = 'Stop'
 
-Update-SessionEnvironment
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+. $toolsPath\helpers.ps1
 
-git config --system --unset credential.helper
+$gitExecutable = Get-GitExecutable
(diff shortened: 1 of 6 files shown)


pascalberger:
@ferventcoder You mean by adding the functionality to the `git.install` package and make it available through a package parameter? Yes, this would be possible, but same points as discussed on https://github.com/chocolatey/chocolatey-coreteampackages/pull/515. Since this is not a feature of the installer IMHO there shouldn't be put additional logic into the chocolatey script. Also an additional packages works way better for dependency management (otherwise it would require https://github.com/chocolatey/choco/issues/488 to be implemented)

manual/git-disable-gcm/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,5 @@
+﻿$ErrorActionPreference = 'Stop'
+
+Update-SessionEnvironment
+
+git config --system --unset credential.helper
```
AdmiringWorm: should probably parse the registry or something for the git executable, just in case the user doesn't it available in PATH.

also, as per [2.2 Formatting](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#22-formatting) you're missing an empty line at the end of the file

manual/git-disable-gcm/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,5 @@
+﻿$ErrorActionPreference = 'Stop'
+
+Update-SessionEnvironment
+
+git config --system credential.helper manager
```
AdmiringWorm: same as my other comment.

pascalberger:
@AdmiringWorm I added path detection from registry and added empty lines at the end of the file. Should be ready for another review

PR #583, opened 2017-02-13
Allow numbers in flag names

petemounce:
Tested locally:
```
C:\src\github\chocolatey\chocolatey-coreteampackages\extensions\chocolatey-core.extension\extensions [fix_get_pp +0 ~1 -0 !]
λ  import-module .\chocolatey-core.psm1
C:\src\github\chocolatey\chocolatey-coreteampackages\extensions\chocolatey-core.extension\extensions [fix_get_pp +0 ~1 -0 !]
λ  get-packageparameters -Parameters "/foo:bar"
p:  /foo:bar
System.Collections.DictionaryEntry

Name                           Value
----                           -----
foo                            bar


C:\src\github\chocolatey\chocolatey-coreteampackages\extensions\chocolatey-core.extension\extensions [fix_get_pp +0 ~1 -0 !]
λ  get-packageparameters -Parameters "/foo2:bar"
p:  /foo2:bar
System.Collections.DictionaryEntry

Name                           Value
----                           -----
foo2                           bar
```

### DIFF
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -1,7 +1,10 @@
 # CHANGELOG
 
+## 1.0.7
+- Bugfix in `Get-PackageParameters`: flags can now have numbers in their names, whereas before, everything past the number would be truncated and the flag would turn into a boolean.
+
 ## 1.0.6
-- Bufix in `Get-AppInstallLocation`: Powershell 2 can not replace on null value.
+- Bugfix in `Get-AppInstallLocation`: Powershell 2 can not replace on null value.
 
 ## 1.0.5
 
--- a/extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
+++ b/extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>chocolatey-core.extension</id>
-    <version>1.0.6</version>
+    <version>1.0.7</version>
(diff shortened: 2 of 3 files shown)


AdmiringWorm:
this looks good to me as well, with the exception of perhaps saying parameter name instead of flags (could maybe be easier understood by maintainers without any prior programming knowledge).
I'm not sure if this should be considered a bugfix though, IMO adding the ability to also use numbers in the flags/parameters should perhaps be considered a feature instead of a bug.

Thoughts? 
/cc @gep13 @pascalberger

pascalberger:
@AdmiringWorm I can see both arguments regarding bug vs feature. My vote would go with bug though :)

gep13:
I guess it comes down to intended functionality.  I think the intent of the package parameters was to correctly parse all parameter names, and their associated values.  In my mind, this would have included a parameter that included a number in it's name.  As a result, I would suggest that this would be a bug fix.

petemounce:
Can I leave it to one of you to update the changelog for bugfix/feature, please? I don't think it should be me - but I do want to have access to the fix as soon as possible since I have a dependency on it now :)

extensions/chocolatey-core.extension/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # CHANGELOG
 
+## 1.0.7
+- Bugfix in `Get-PackageParameters`: flags can now have numbers in their names, whereas before, everything past the number would be truncated and the flag would turn into a boolean.
+
 ## 1.0.6
-- Bufix in `Get-AppInstallLocation`: Powershell 2 can not replace on null value.
+- Bugfix in `Get-AppInstallLocation`: Powershell 2 can not replace on null value.
```
AdmiringWorm: ah, thanks for fixing my typo for the previous release 😄

PR #597, opened 2017-02-20
(chocolatey-core.extension) New Helper 'Get-AvailableDriveLetter'

mwallner:
should return the next free drive letter

very useful for mounting iso-files from cocolateyinstall.ps1

### DIFF
--- a/extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
+++ b/extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
@@ -22,7 +22,7 @@
   http://stackoverflow.com/questions/12488030/getting-a-free-drive-letter/29373301#29373301
 #>
 function Get-AvailableDriveLetter {
-  param ([char[]]$ExcludedLetter)
+  param ([char[]]$ExcludedLetters)
 
   $Letter = [int][char]'C'
   $i = @()
@@ -31,11 +31,11 @@ function Get-AvailableDriveLetter {
   $(Get-PSDrive -PSProvider filesystem) | %{$i += $_.name}
   
   #Adding the excluded letter
-  $i+=$ExcludedLetter
+  $i+=$ExcludedLetters
   
   while($i -contains $([char]$Letter)){$Letter++}
 
-  if ($Letter -gt [int][char]'Z') {
+  if ($Letter -gt [char]'Z') {
     throw "error: no drive letter available!"
   }
   Write-Verbose "available drive letter: '$([char]$Letter)'"
--- a/extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
+++ b/extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
@@ -13,7 +13,7 @@
   (do not return X, even if it'd be the next choice)
 
 .INPUTS
-  specific drive letter 
+  specific drive letter(s) that will be excluded as potential candidates
 
 .OUTPUTS
   System.String (single drive-letter character)
@@ -22,7 +22,7 @@
   http://stackoverflow.com/questions/12488030/getting-a-free-drive-letter/29373301#29373301
 #>
 function Get-AvailableDriveLetter {
-  param ([char]$ExcludedLetter)
+  param ([char[]]$ExcludedLetter)
 
   $Letter = [int][char]'C'
   $i = @()
@@ -35,6 +35,10 @@ function Get-AvailableDriveLetter {
   
   while($i -contains $([char]$Letter)){$Letter++}
 
+  if ($Letter -gt [int][char]'Z') {
+    throw "error: no drive letter available!"
+  }
+  Write-Verbose "available drive letter: '$([char]$Letter)'"
   Return $([char]$Letter)
 }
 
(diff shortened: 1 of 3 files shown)


extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
```diff
@@ -0,0 +1,40 @@
...
+  specific drive letter 
+
+.OUTPUTS
+  System.String (single drive-letter character)
+
+.LINK
+  http://stackoverflow.com/questions/12488030/getting-a-free-drive-letter/29373301#29373301
+#>
+function Get-AvailableDriveLetter {
+  param ([char]$ExcludedLetter)
```
AdmiringWorm: perhaps allow a user to pass several drive letters for this

extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
```diff
@@ -0,0 +1,40 @@
...
+  
+  #getting all the used Drive letters reported by the Operating System
+  $(Get-PSDrive -PSProvider filesystem) | %{$i += $_.name}
+  
+  #Adding the excluded letter
+  $i+=$ExcludedLetter
+  
+  while($i -contains $([char]$Letter)){$Letter++}
+
+  Return $([char]$Letter)
```
AdmiringWorm: please add either `Write-Debug` or `Write-Verbose` that mentions which drive letter is returned

extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
```diff
@@ -0,0 +1,40 @@
...
+
+.EXAMPLE
+  Get-AvailableDriveLetter
+
+.EXAMPLE
+  Get-AvailableDriveLetter 'X'
+  (do not return X, even if it'd be the next choice)
+
+.INPUTS
+  specific drive letter
```
AdmiringWorm: Even though the parameter is self explanatory, this should probably say that it's the drive letter(s) that its not to use.

extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
```diff
@@ -0,0 +1,44 @@
...
+  
+  #getting all the used Drive letters reported by the Operating System
+  $(Get-PSDrive -PSProvider filesystem) | %{$i += $_.name}
+  
+  #Adding the excluded letter
+  $i+=$ExcludedLetter
+  
+  while($i -contains $([char]$Letter)){$Letter++}
+
+  if ($Letter -gt [int][char]'Z') {
```
ferventcoder: I'm assuming that $Letter also doesn't need to be converted to a letter?

Also, `c` and `C` are different chars. Are you planning to handle when the Excluded letter provided is not in uppercase?

extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
```diff
@@ -0,0 +1,44 @@
...
+  specific drive letter(s) that will be excluded as potential candidates
+
+.OUTPUTS
+  System.String (single drive-letter character)
+
+.LINK
+  http://stackoverflow.com/questions/12488030/getting-a-free-drive-letter/29373301#29373301
+#>
+function Get-AvailableDriveLetter {
+  param ([char[]]$ExcludedLetter)
```
ferventcoder: Is there a better name for this? Also, this could use some PARAMETER docs.

mwallner: open for suggestions.. ?
* Exclude
* ExcludeThisLetter(s)
* ExcludeList

.. can't think of anything that doesn't sound totally awkward without 'Exclude' in the parameter name
* IgnoreLetter(s) 
* KeepDriveLetter(s)

ferventcoder: StartAfterLetter ?

mwallner: hm .. possible but I don't like it - as it isn't quite right .. 
I could specify -StartAfterLetter 'V' .. but maybe 'G' is possible - so 'G' would be returned even though I specified 'StartAfter V'

ferventcoder: Ah fair, I read the code wrong. I thought it was starting with the exclude letter. I agree with @AdmiringWorm that it would be beneficial to make this an array.

mwallner: :confused: it is an array .. I'll just leave it as it is then? :smile:

extensions/chocolatey-core.extension/extensions/Get-AvailableDriveLetter.ps1
mwallner: the `-contains` cmdlet is case-insensitive - I'm starting with big 'C' and stopping with big 'Z' - therefore it should be safe ..
(am I missing something?)

AdmiringWorm: I tested this with lowercase letters used in the exclude letters parameters, which worked as expected.
So I think this would be fine as it is now.

mwallner:
**ping** @gep13 @pascalberger

AdmiringWorm:
I've updated the changelog in for this PR, and bumped the version.
So whoever ends merging this can now just merge it with `[PUSH chocolatey-core.extension]` in the merge title.

PR #602, opened 2017-02-28
PDF Creator Download Link Update

tkriviradev:
Hello I just post the updated link to the pdf creator


AdmiringWorm:
thank you for your contribution.
However, we are using the urls that are listed [here](http://download.pdfforge.org/download/pdfcreator/list) which still uses the same url format that the current package use.

Even if the url had changed, the [update script](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/PDFCreator/update.ps1) would probably need to be changed (the url is automatically inject in the install script during each appveyor update run)

as such I'm going to close this, feel free to reopen it if you disagree.

PR #605, opened 2017-02-28
(googlechrome) Added revision padding

### LINKED ISSUE #604: Google Chrome binaries have changed for 56.0.2924.87
The version is the same, however the binaries have changed.

https://chocolatey.org/packages/GoogleChrome#comment-3171878283

When the package was created for version 56.0.2924.87, the checksums were:

~~~powershell
 url                    = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise.msi'
  url64bit               = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
  checksum               = '9a5e2810a164654d9a4a42f0597d8668110134ee7d1aca0faf8cc48d203fc019'
  checksum64             = 'c47ff551ff8b251268905cd87fb299959e6bcb3640a3e60085a3a7dbf176845f'
~~~

However without a new version the checksums are now:

~~~powershell
checksum    = '8B83B51566E02A349BF4F94B753D198883994C713D8DE001B0ABF7E650BDC9FB'
checksum64  = '1450DE87F0289CE725DF1C138BBA2666304FB94BF4476B03C06D058C999D4805'
~~~

AdmiringWorm:
This will add padding to the revision number whenever the version is less than `56.0.2925`

**IMPORTANT merge using `[AU googlechrome]`, otherwise the padding is applied but the revision number is not incremented (but still be pushed since `56.0.2925.87 != 56.0.2925.87000`**

fixes #604

### DIFF
--- a/automatic/GoogleChrome/update.ps1
+++ b/automatic/GoogleChrome/update.ps1
@@ -1,6 +1,8 @@
 import-module au
+. "$PSScriptRoot\..\..\scripts\Get-Padded-Version.ps1"
 
 $releases = 'http://omahaproxy.appspot.com/all?os=win&amp;channel=stable'
+$paddedUnderVersion = '56.0.2925'
 
 function global:au_SearchReplace {
    @{
@@ -15,10 +17,12 @@ function global:au_SearchReplace {
 
 function global:au_GetLatest {
     $release_info = Invoke-WebRequest -Uri $releases -UseBasicParsing
+    $version = $release_info | % Content | ConvertFrom-Csv | % current_version
+
     @{
         URL32 = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise.msi'
         URL64 = 'https://dl.google.com/tag/s/dl/chrome/install/googlechromestandaloneenterprise64.msi'
-        Version = $release_info | % Content | ConvertFrom-Csv | % current_version
+        Version = Get-Padded-Version -Version $version -OnlyBelowVersion $paddedUnderVersion -RevisionLength 5
     }
 }
 


automatic/GoogleChrome/update.ps1
```diff
@@ -1,6 +1,8 @@
 import-module au
+. "$PSScriptRoot\..\..\scripts\Get-Padded-Version.ps1"
 
 $releases = 'http://omahaproxy.appspot.com/all?os=win&amp;channel=stable'
+$paddedUnderVersion = '56.0.2925'
```
pascalberger: Wouldn't it be better to have this also for future versions (assuming that Google won't change its practice)?

AdmiringWorm: I'm not sure, which is why I opened the PR instead of committing directly

ferventcoder:
Why four `0000` after the number? Wouldn't two or three be sufficient? If you go 1,000 package revisions, I think there is a problem.

AdmiringWorm:
@ferventcoder 
The reason I've been telling myself is to try showing a difference between the official version and the package fix iteration. Considering the next released revision version would be 1xx I don't think it would be too many zeroes.

AdmiringWorm:
@pascalberger I've updated to add padding for all versions
@ferventcoder I've changed the revision length to be a maximum length of 5 (making it 1-2 zeroes between actual version and fix incrementation)

automatic/GoogleChrome/update.ps1
ferventcoder: I like staying on the side of safety for now. If we need to do this later it is easy enough to add back in.

AdmiringWorm: 👍 I'll update again to add it back

ferventcoder:
I actually liked not having the padding until we need it, so I liked erring on the side of safety.

AdmiringWorm:
@ferventcoder I just readded the `paddedUnderVersion` variable.

AdmiringWorm:
looks like google chrome is failing all checks.
I'm guessing because of aws outage?
Guess we need to repush when the outage is over.

PR #609, opened 2017-03-02
(visualstudiocode-disableautoupdate) Update to latest Visual Studio Code version

pascalberger:
Current version of Visual Studio Code no longer seems to support `storage.json` file. Setting is therefore written to `settings.json`. Also default channel is now called `default` instead of `stable`.

Needs to be merged with `[PUSH visualstudiocode-disableautoupdate]` in message

### DIFF
--- a/manual/visualstudiocode-disableautoupdate/tools/chocolateyUninstall.ps1
+++ b/manual/visualstudiocode-disableautoupdate/tools/chocolateyUninstall.ps1
@@ -3,4 +3,4 @@
 $toolsPath = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
 . $toolsPath\helpers.ps1
 
-Set-UpdateChannel("stable")
+Set-UpdateChannel("default")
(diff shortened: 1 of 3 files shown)


pascalberger:
@AdmiringWorm Was myself undecided about that :) I decided to use fix notation to avoid confusion with Visual Studio Code version. But if you think increasing patch version is better I have absolutely no problem with that.

AdmiringWorm:
@pascalberger 
nah, just keep it that way. I agree it could avoid confusion with vscode version.

Just another thing, this is a required change for only the recently submitted version of vscode right?
If so, shouldn't the dependency version be incremented, since it won't be compatible with older versions of vscode?

pascalberger:
@AdmiringWorm I'm not entirely sure in which version they actually introduced it. The change from `storage.json` to `settings.json` actually happend some time ago, as seem the change from `stable` to `default`. So I think this solution should also work for older versions.

AdmiringWorm:
👍 
I see.
Well nothing left than merging and pushing the new version then.

PR #611, opened 2017-03-05
(piriform-packages) Fixed update hanging issue

AdmiringWorm:
All piriform packages was hanging on appveyor because they was missing the `-UseBasicParsing` flag when calling iwr.


majkinetor:
This wont work once h3 tag crosses multiple lines. But we can go with that and get notified if AU complains about version format.

AdmiringWorm:
@majkinetor that's true, it won't.
I didn't even take that into consideration, but that should probably be easy to fix if it occurs sometime in the future.

PR #612, opened 2017-03-05
(chocolatey-core.extension) Added simple wrapper around Install-ChocolateyInstallPackage

AdmiringWorm:
This wrapper makes it easier to embedd both 32-bit and 64-bit
installers inside a package, extracting the needed check
to decide which installer is meant to be run.

I plan to make the same for `Get-ChocolateyUnzip`, but want to see if I'm moving in the right direction first.

### DIFF
--- a/extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
+++ b/extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
@@ -1,0 +1,179 @@
+. "$env:ChocolateyInstall\helpers\functions\Install-ChocolateyInstallPackage.ps1"
+$installPackageFunc = Get-Item Function:\Install-ChocolateyInstallPackage | ? { $_.Parameters.ContainsKey('file64') }
+if ($installPackageFunc) {
+  return
+}
+
+Write-Debug "Loading Install-ChocolateyInstallPackage override"
+Rename-Item Function:Install-ChocolateyInstallPackage Install-ChocolateyInstallPackageOriginal
+
+<#
+.SYNOPSIS
+**NOTE:** Administrative Access Required.
+Installs software into "Programs and Features". Use
+Install-ChocolateyPackage when software must be downloaded first.
+.DESCRIPTION
+This will run an installer (local file) on your machine.
+.NOTES
+This command will assert UAC/Admin privileges on the machine.
+If you are embedding files into a package, ensure that you have the
+rights to redistribute those files if you are sharing this package
+publicly (like on the community feed). Otherwise, please use
+Install-ChocolateyPackage to download those resources from their
+official distribution points.
+This is a native installer wrapper function. A "true" package will
+contain all the run time files and not an installer. That could come
+pre-zipped and require unzipping in a PowerShell script. Chocolatey
+works best when the packages contain the software it is managing. Most
+software in the Windows world comes as installers and Chocolatey
+understands how to work with that, hence this wrapper function.
+.INPUTS
+None
+.OUTPUTS
+None
+.PARAMETER PackageName
+The name of the package - while this is an arbitrary value, it's
+recommended that it matches the package id.
+.PARAMETER FileType
+This is the extension of the file. This can be 'exe', 'msi', or 'msu'.
+Licensed editions of Chocolatey use this to automatically determine
+silent arguments. If this is not provided, Chocolatey will
+automatically determine this using the downloaded file's extension.
+.PARAMETER SilentArgs
+OPTIONAL - These are the parameters to pass to the native installer,
+including any arguments to make the installer silent/unattended.
+Pro/Business Editions of Chocolatey will automatically determine the
+installer type and merge the arguments with what is provided here.
+Try any of the to get the silent installer -
+`/s /S /q /Q /quiet /silent /SILENT /VERYSILENT`. With msi it is always
+`/quiet`. Please pass it in still but it will be overridden by
+Chocolatey to `/quiet`. If you don't pass anything it could invoke the
+installer with out any arguments. That means a nonsilent installer.
+Please include the `notSilent` tag in your Chocolatey package if you
+are not setting up a silent/unattended package. Please note that if you
+are submitting to the community repository, it is nearly a requirement
+for the package to be completely unattended.
+.PARAMETER File
+Full file path to native installer to run. If embedding in the package,
+you can get it to the path with
+`"$(Split-Path -parent $MyInvocation.MyCommand.Definition)\\INSTALLER_FILE"`
+In 0.10.1+, `FileFullPath` is an alias for File.
+This can be a 32-bit or 64-bit file. This is mandatory in earlier versions
+of Chocolatey, but optional if File64 has been provided.
+.PARAMETER File64
+Full file path to a 64-bit native installer to run. Available in 0.10.4+.
+If embedding in the package, you can get it to the path with
+`"$(Split-Path -parent $MyInvocation.MyCommand.Definition)\\INSTALLER_FILE"`
+Provide this when you want to provide both 32-bit and 64-bit
+installers or explicitly only a 64-bit installer (which will cause a package
+install failure on 32-bit systems).
+.PARAMETER ValidExitCodes
+Array of exit codes indicating success. Defaults to `@(0)`.
+.PARAMETER UseOnlyPackageSilentArguments
+Do not allow choco to provide/merge additional silent arguments and
+only use the ones available with the package. Available in 0.9.10+.
+.PARAMETER IgnoredArguments
+Allows splatting with arguments that do not apply. Do not use directly.
+.EXAMPLE
+>
+$packageName= 'bob'
+$toolsDir   = "$(Split-Path -Parent $MyInvocation.MyCommand.Definition)"
+$fileLocation = Join-Path $toolsDir 'INSTALLER_EMBEDDED_IN_PACKAGE'
+$packageArgs = @{
+  packageName   = $packageName
+  fileType      = 'msi'
+  file          = $fileLocation
+  silentArgs    = "/qn /norestart"
+  validExitCodes= @(0, 3010, 1641)
+  softwareName  = 'Bob*'
+}
+Install-ChocolateyInstallPackage @packageArgs
+.EXAMPLE
+$packageName= 'bob'
+$toolsDir   = "$(Split-Path -Parent $MyInvocation.MyCommand.Definition)"
+$fileLocation = Join-Path $toolsDir '32BIT_INSTALLER_EMBEDDED_IN_PACKAGE'
+$fileLocation64 = Join-Path $toolsDir '64BIT_INSTALLER_EMBEDDED_IN_PACKAGE'
+
+$packageArgs = @{
+  packageName   = $packageName
+  fileType      = 'msi'
+  file          = $fileLocation
+  file64        = $fileLocation64
+  silentArgs    = "/qn /norestart"
+  validExitCodes= @(0, 3010, 1641)
+  softwareName  =
+}
+
+Install-ChocolateyInstallPackageEx @packageArgs
+.EXAMPLE
+>
+$packageArgs = @{
+  packageName   = 'bob'
+  fileType      = 'exe'
+  file          = '\\SHARE_LOCATION\to\INSTALLER_FILE'
+  silentArgs    = "/S"
+  validExitCodes= @(0)
+  softwareName  = 'Bob*'
+}
+Install-ChocolateyInstallPackage @packageArgs
+.EXAMPLE
+Install-ChocolateyInstallPackage 'bob' 'exe' '/S' "$(Split-Path -Parent $MyInvocation.MyCommand.Definition)\bob.exe"
+.EXAMPLE
+>
+Install-ChocolateyInstallPackage -PackageName 'bob' -FileType 'exe' `
+  -SilentArgs '/S' `
+  -File "$(Split-Path -Parent $MyInvocation.MyCommand.Definition)\bob.exe" `
+  -ValidExitCodes = @(0)
+.LINK
+Install-ChocolateyPackage
+.LINK
+Uninstall-ChocolateyPackage
+.LINK
+Get-UninstallRegistryKey
+.LINK
+Start-ChocolateyProcessAsAdmin
+#>
+function Install-ChocolateyInstallPackage {
+  param(
+    [parameter(Mandatory=$true, Position=0)][string] $packageName,
+    [parameter(Mandatory=$false, Position=1)]
+    [alias("installerType","installType")][string] $fileType = 'exe',
+    [parameter(Mandatory=$false, Position=2)][string[]] $silentArgs = '',
+    [alias("fileFullPath")][parameter(Mandatory=$false, Position=3)][string] $file,
+    [alias("fileFullPath64")][Parameter(Mandatory=$false)][string] $file64,
+    [parameter(Mandatory=$false)] $validExitCodes = @(0),
+    [parameter(Mandatory=$false)]
+    [alias("useOnlyPackageSilentArgs")][switch] $useOnlyPackageSilentArguments = $false,
+    [parameter(ValueFromRemainingArguments = $true)][Object[]] $ignoredArguments
+  )
+  $bitnessMessage = ''
+  $fileFullPath = $file
+  if ((Get-ProcessorBits 32) -or $env:ChocolateyForceX86 -eq 'true') {
+    if (!$file) { throw "32-bit installation is not supported for $packageName" }
+    if ($file64) { $bitnessMessage = '32-bit ' }
+  } elseif( $file64) {
+    $fileFullPath = $file64
+    $bitnessMessage = '64-bit '
+  }
+
+  if ($fileFullPath -eq '' -or $fileFullPath -eq $null) {
+    throw "Package parameters incorrect, either File or File64 must be specified."
+  }
+
+  Write-Host "Installing $bitnessMessage$packageName..."
+
+  $packageArgs = @{
+    packageName    = $packageName
+    fileType       = $fileType
+    silentArgs     = $silentArgs
+    file           = $fileFullPath
+    validExitCodes = $validExitCodes
+    useOnlyPackageSilentArgs = $useOnlyPackageSilentArguments
+    ignoredArguments = $ignoredArguments
+  }
+
+  Install-ChocolateyInstallPackageOriginal @packageArgs
+}
+
+# We need to force the export of Install-ChocolateyInstall function
+Export-ModuleMember -Function Install-ChocolateyInstallPackage
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 1.4.0
+- `Install-ChocolateyInstallPackage`: Overrides chocos own function and provides option to use 64-bit files
+
 ## 1.3.0
 
 - `Get-EffectiveProxy`: Get the current proxy using several methods
--- a/extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
+++ b/extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>chocolatey-core.extension</id>
-    <version>1.3.0</version>
+    <version>1.4.0</version>
     <title>Chocolatey Core Extensions</title>
     <summary>Helper functions extending core choco functionality</summary>
     <authors>chocolatey</authors>


majkinetor:
LGTM

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackageEx.ps1
```diff
@@ -0,0 +1,94 @@
...
+    if (!$file64) {
+      $fileFullPath = $file
+    } else {
+      Write-Host "Installing $packageName 64-bit"
+    }
+    $fileFullPath = $file64
+  }
+
+  $packageArgs = @{
+    packageName    = $packageName
```
pascalberger: Remove whitespace before `=` or intend all of them to the same level

AdmiringWorm: I've indented them this way on purpose, if I were to align them all to the same place as `useOnlyPackageSilentArgs` then that would be too much IMO.
Therefore I indented only the first 5 arguments.

ferventcoder: It looks fine to me.

ferventcoder:
Would it be simply better to just add this to 0.10.4?

ferventcoder:
I'm just thinking about all of the things, like the package-validator and those similar that would now need to account for this. Not that excited about changing all of those systems so they can see this as well when we can simply fix Chocolatey.

ferventcoder:
And yes, I do realize that means it would be a few months before we could account for it on the community repository (about 6 months to allow the community to fully upgrade). As we've seen, folks are still upgrading from 0.9.8.x (over 3 years old now!!!!) so that things will work with latest changes on the community repository.

AdmiringWorm:
@ferventcoder Yeah it would, but that would require the user to run choco v 0.10.4+.
> I'm just thinking about all of the things, like the package-validator and those similar that would now need to account for this. Not that excited about changing all of those systems so they can see this as well when we can simply fix Chocolatey.

I understand that, the reason I added this script is to have an option when users are running an older version of choco.
If there is a way to substitute the function when an older version is used, then I'm all ears.

ferventcoder:
> If there is a way to substitute the function when an older version is used, then I'm all ears.

@AdmiringWorm there actually is. It's actually somewhat simple and [something we used a long time ago when we were building mocking for Pester](https://github.com/chocolatey/chocolatey/blob/6bcc3971e501cb8e7bd9ece2d44a6252a7b498a6/tests/_FunctionRenameActuals.ps1).

You can look at the version of Chocolatey `$CHOCOLATEY_VERSION` - https://github.com/chocolatey/choco/wiki/HelpersReference#environment-variables

> CHOCOLATEY_VERSION - The version of Choco you normally see. Use if you are 'lighting' things up based on choco version. (0.9.9+) - Otherwise take a dependency on the specific version you need.

If that version is less than version 0.10.4 (flip it to use System.Version I believe and compare the two).

Then use that to decide whether to load this or not - if you decide to load the function, capture the original `Install-ChocolateyInstallPackage`:

~~~powershell

rename-item function:Install-ChocolateyInstallPackage Install-ChocolateyInstallPackageActual
~~~

And rename the `Install-ChocolateyInstallPackageEx` function to `Install-ChocolateyInstallPackage`. So it would look something like the following:

~~~powershell
$currentChocoVersion = $env:CHOCOLATEY_VERSION
if ($currentChocoVersion -and [Version]$currentChocoVersion -ge [Version]'0.10.4'){
  return
}

Write-Debug "Loading Install-ChocolateyInstallPackage override"
Rename-Item function:Install-ChocolateyInstallPackage Install-ChocolateyInstallPackage-Actual

# docs here
function Install-ChocolateyInstallPackage {
  # code here
  Install-ChocolateyInstallPackage-Actual @packageArgs  
}
~~~

AdmiringWorm:
@ferventcoder awesome, I'll definitely take a look at changing this function to something like that.
I'm not really a fan of just testing for a certain version though, but I did find a way to test if a certain parameter exists in a function.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackageEx.ps1
```diff
@@ -0,0 +1,94 @@
...
+#>
+
+function Install-ChocolateyInstallPackageEx {
+  param(
+    [parameter(Mandatory=$true, Position=0)][string] $packageName,
+    [parameter(Mandatory=$false, Position=1)]
+    [alias("installerType","installType")][string] $fileType = 'exe',
+    [parameter(Mandatory=$false, Position=2)][string[]] $silentArgs = '',
+    [alias("fileFullPath")][parameter(Mandatory=$false, Position=3)][string] $file,
+    [alias("fileFullPath64","file64Bit")][Parameter(Mandatory=$false, Position=4)][string] $file64,
```
ferventcoder: Remove position. We are moving away from this. So never add new items with a position.

AdmiringWorm: will do

ferventcoder: Also not a fan of $file64bit, can it be removed? I want this to have parity with function updates that will roll back into Chocolatey.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackageEx.ps1
```diff
@@ -0,0 +1,94 @@
...
+Install-ChocolateyInstallPackageEx @packageArgs
+#>
+
+function Install-ChocolateyInstallPackageEx {
+  param(
+    [parameter(Mandatory=$true, Position=0)][string] $packageName,
+    [parameter(Mandatory=$false, Position=1)]
+    [alias("installerType","installType")][string] $fileType = 'exe',
+    [parameter(Mandatory=$false, Position=2)][string[]] $silentArgs = '',
+    [alias("fileFullPath")][parameter(Mandatory=$false, Position=3)][string] $file,
```
ferventcoder: Do you think we should leave this mandatory - This will either be a 64bit file or 32bit if more than one is sent.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackageEx.ps1
AdmiringWorm: sure, it can be removed. I actually added it because of the use of `-Url64bit ` in the `Install-ChocolateyPackage`

ferventcoder:
> I'm not really a fan of just testing for a certain version though, but I did find a way to test if a certain parameter exists in a function.

That's fine as well. And likely better.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackageEx.ps1
AdmiringWorm: I don't really think this should be mandatory, if a package only supports 64-bit the IMO the helper function should throw an exception when trying to install the 32-bit version

ferventcoder: My point here is making this as well transitioned as possible.

ferventcoder: The documentation leads me to believe this is 32-bit, when in fact I can pass a 64bit installer through $file. Think of how this would be used over time and transition.

ferventcoder: Although I see the idea of making it optional so we can fail early if the file is only supported in 64bit scenarios.

ferventcoder:
Filed https://github.com/chocolatey/choco/issues/1187

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackageEx.ps1
AdmiringWorm: The idea was to have it behave somewhat similiar to the `Install-ChocolateyPackage` urls parameters (which also gives the impression that `Url` parameter is only for 32bit).
I'll definitely need to improve the documentation though.
I don't really see the problem with the transition though, since it will still fail if both file and file64 is not set.

AdmiringWorm:
@ferventcoder I've added the check and renaming of the existing function.
I haven't yet tested it in POSH v2 though, but that as well as the other comments must unfortunately have to wait until tomorrow.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
```diff
@@ -0,0 +1,104 @@
...
+    validExitCodes = $validExitCodes
+    useOnlyPackageSilentArgs = $useOnlyPackageSilentArguments
+    ignoredArguments = $ignoredArguments
+  }
+
+  Install-ChocolateyInstallPackageOriginal @packageArgs
+}
+
+# We need to force the export of Install-ChocolateyInstall function
+Export-ModuleMember -Function Install-ChocolateyInstallPackage
```
ferventcoder: Will that work from here?

ferventcoder: And this is why I had a particular recommendation on letting each function decide whether to load up versus the universal one in the PowerShell module file. Sometimes I have a good reason that I can't always explain at the proper moment. Or it doesn't had those needs like this would have now.

AdmiringWorm: I believe so yes (I did try it on my own computer, although only through importing `chocolatey-core.psm1` manually).
Since the `chocolatey-core.psm1` is sourcing all of the files, it just don't export it since it had found an existing function with the same name.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
```diff
@@ -0,0 +1,104 @@
...
+}
+
+Install-ChocolateyInstallPackageEx @packageArgs
+#>
+$have64BitFileParam = Get-Item function:\Install-ChocolateyInstallPackage | ? { $_.Parameters.ContainsKey('file64') }
+if ($have64BitFileParam) {
+  return
+}
+
+Write-Debug "Loading Install-ChocolateyInstallPackage override"
```
ferventcoder: And these debugging messages are necessary for determining issues - it may not be a bad idea to also have a debug message in the function itself.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
AdmiringWorm: Before importing the chocolatey-core.psm1 file manually the following functions was found:
```powershell
> Get-ChildItem function:\Install*

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Function        Install-BinFile                                    0.0        chocolateyInstaller
Function        Install-ChocolateyDesktopLink                      0.0        chocolateyInstaller
Function        Install-ChocolateyEnvironmentVariable              0.0        chocolateyInstaller
Function        Install-ChocolateyExplorerMenuItem                 0.0        chocolateyInstaller
Function        Install-ChocolateyFileAssociation                  0.0        chocolateyInstaller
Function        Install-ChocolateyInstallPackage                   0.0        chocolateyInstaller
Function        Install-ChocolateyPackage                          0.0        chocolateyInstaller
Function        Install-ChocolateyPath                             0.0        chocolateyInstaller
Function        Install-ChocolateyPinnedTaskBarItem                0.0        chocolateyInstaller
Function        Install-ChocolateyPowershellCommand                0.0        chocolateyInstaller
Function        Install-ChocolateyShortcut                         0.0        chocolateyInstaller
Function        Install-ChocolateyVsixPackage                      0.0        chocolateyInstaller
Function        Install-ChocolateyZipPackage                       0.0        chocolateyInstaller
Function        Install-VisualStudio                               0.0        chocolateyInstaller
Function        Install-Vsix                                       0.0        chocolateyInstaller
Function        Install-WindowsUpdate                              0.0        chocolateyInstaller
```

After importing:
```powershell
> Get-ChildItem function:\Install*

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Function        Install-BinFile                                    0.0        chocolateyInstaller
Function        Install-ChocolateyDesktopLink                      0.0        chocolateyInstaller
Function        Install-ChocolateyEnvironmentVariable              0.0        chocolateyInstaller
Function        Install-ChocolateyExplorerMenuItem                 0.0        chocolateyInstaller
Function        Install-ChocolateyFileAssociation                  0.0        chocolateyInstaller
Function        Install-ChocolateyInstallPackage                   0.0        chocolatey-core
Function        Install-ChocolateyPackage                          0.0        chocolateyInstaller
Function        Install-ChocolateyPath                             0.0        chocolateyInstaller
Function        Install-ChocolateyPinnedTaskBarItem                0.0        chocolateyInstaller
Function        Install-ChocolateyPowershellCommand                0.0        chocolateyInstaller
Function        Install-ChocolateyShortcut                         0.0        chocolateyInstaller
Function        Install-ChocolateyVsixPackage                      0.0        chocolateyInstaller
Function        Install-ChocolateyZipPackage                       0.0        chocolateyInstaller
Function        Install-VisualStudio                               0.0        chocolateyInstaller
Function        Install-Vsix                                       0.0        chocolateyInstaller
Function        Install-WindowsUpdate                              0.0        chocolateyInstaller
Function        Install-ChocolateyInstallPackageOriginal           0.0        chocolatey-core
```

ferventcoder: Ruhroh - looks like it is not picking up the original one.

`Function        Install-ChocolateyInstallPackageOriginal           0.0        chocolatey-core`

That should probably be

`Function        Install-ChocolateyInstallPackageOriginal           0.0        chocolateyInstaller`

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
```diff
@@ -0,0 +1,104 @@
...
+  fileType      = 'msi'
+  file          = $fileLocation
+  file64        = $fileLocation64
+  silentArgs    = "/qn /norestart"
+  validExitCodes= @(0, 3010, 1641)
+  softwareName  =
+}
+
+Install-ChocolateyInstallPackageEx @packageArgs
+#>
```
ferventcoder: Synopsis and docs should be right by `function Install-ChocolateyInstallPackage {`.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
AdmiringWorm: I think it's a drawback of the global import file, and the use of rename-item.
Could be wrong though.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
```diff
@@ -0,0 +1,104 @@
...
+      Write-Host "Installing $packageName 32-bit"
+    }
+    $fileFullPath = $file
+  }else {
+    if (!$file64) {
+      $fileFullPath = $file
+    } else {
+      Write-Host "Installing $packageName 64-bit"
+    }
+    $fileFullPath = $file64
```
ferventcoder: This looks incorrect.

ferventcoder: This should be captured inside of the else statement above it.

AdmiringWorm:
Looks like Rename-Item can't be used as it currently is.
The extensions are imported before chocos own extensions is imported, which results in the original `Install-ChocolateyInstallPackage` function isn't yet available 😢

ferventcoder:
IIRC - Chocolatey loads its functions first, but it may not export everything until later, which results in a much less deterministic way of knowing what is loaded.

ferventcoder:
However, I'm guessing you may be able to search for the function in `chocolateyInstaller` module namespace.

AdmiringWorm:
> Chocolatey loads its functions first, but it may not export everything until later

Sorry, that's what I meant.

> However, I'm guessing you may be able to search for the function in chocolateyInstaller module namespace.

Yeah it could be, I'll need to have a look at that.

ferventcoder:
@AdmiringWorm changed up the logic a bit - Let's ensure we match up on parameter names and aliases - https://github.com/chocolatey/choco/commit/4cf244184b18e1befcfe86ddca114cef3f4459a3

AdmiringWorm:
@ferventcoder I'll update this one with the same logic.
Still struggling with the renaming the original function though, currently the only way I've found that works is to first manually source the original `Install-ChocolateyInstallPackage` script.

extensions/chocolatey-core.extension/extensions/Install-ChocolateyInstallPackage.ps1
AdmiringWorm: When importing with the latest commits,
the results is now this:
```powershell
PS C:\Users\Administrator> Get-ChildItem function:\Install*

CommandType     Name                                               ModuleName
-----------     ----                                               ----------
Function        Install-BinFile                                    chocolateyInstaller
Function        Install-ChocolateyDesktopLink                      chocolateyInstaller
Function        Install-ChocolateyEnvironmentVariable              chocolateyInstaller
Function        Install-ChocolateyExplorerMenuItem                 chocolateyInstaller
Function        Install-ChocolateyFileAssociation                  chocolateyInstaller
Function        Install-ChocolateyInstallPackage                   chocolateyInstaller
Function        Install-ChocolateyInstallPackageOriginal           chocolateyInstaller
Function        Install-ChocolateyPackage                          chocolateyInstaller
Function        Install-ChocolateyPath                             chocolateyInstaller
Function        Install-ChocolateyPinnedTaskBarItem                chocolateyInstaller
Function        Install-ChocolateyPowershellCommand                chocolateyInstaller
Function        Install-ChocolateyShortcut                         chocolateyInstaller
Function        Install-ChocolateyVsixPackage                      chocolateyInstaller
Function        Install-ChocolateyZipPackage                       chocolateyInstaller
Function        Install-Vsix                                       chocolateyInstaller
Function        Install-WindowsUpdate                              chocolateyInstaller
```

ferventcoder: So now the opposite problem.

AdmiringWorm: well, yes and no.
The reason I got chocolatey-core and chocolateyInstaller is becuase previously I imported both chocolateyInstaller.psm1 and chocolatey-core.psm1 separately.

Now I only imported chocolateyInstaller.psm1

ferventcoder:
@AdmiringWorm note I added https://github.com/chocolatey/choco/issues/1200 to address the non-deterministic load.

AdmiringWorm:
> @ferventcoder said
> @AdmiringWorm note I added chocolatey/choco#1200 to address the non-deterministic load (not that it is going to help here)

yeah, that probably won't help in this case.
I honestly don't know how to proceed with this PR, as previously mentioned the only way that I found to work was to first source the `Install-ChocolateyInstallPackage` file which then made it discoverable.

AdmiringWorm:
@ferventcoder any chance of getting another review on this PR? so it hopefully can be merged....

AdmiringWorm:
@ferventcoder I know I'm nagging, but is this PR acceptable as it currently stands?
If not, this PR could probably just be closed as I'm not able to get it to work any other way.

AdmiringWorm:
@majkinetor while that is certainly possible, I believe there should be an option to keep backwards compatibility while still using new functions. Hence why this PR, as well as it was started before it was added to choco.

Like what is done with the `Get-UninstallRegistryKey` function.

majkinetor:
Why promote bad behavior (keeping the old choco version) ? This way you force users to transparently update the choco by putting it as dependency.

AdmiringWorm:
I'm closing this now, as there seem to be no interest in it.

AdmiringWorm:
> Getting the functionality working properly with the same name was my biggest concern.

I understand that, unfortunately I wasn't able to get it to work with the methods you suggested.
The only way I could make it to work was with sourcing the actual script manually (semi-manually?), and although I'm not entirely happy with it I could see no other way.

However, I've dropped trying to get this into the extension. I haven't removed the branch though, as I'm hoping someone will pick it up and continue the work...

AdmiringWorm:
> @majkinetor said...
> Why promote bad behavior (keeping the old choco version) ? This way you force users to transparently update the choco by putting it as dependency.

The idea behind it wasn't really meant as to *promote bad behavior*, but rather to provide an alternative to users that can't update chocolatey (be it because of a bug, or company policy or something else).
As well as I don't really like having a dependency on chocolatey itself (thinks of it like this, the user only have 1 package installed that depends on choco, he removes the package using `choco uninstall -x packageName`, it would then also remove the lib directory of chocolatey making it not updating itself anymore, possibly stop working (happened during testing for me for a package))

majkinetor:
> users that can't update chocolatey (be it because of a bug, or company policy or something else).
You need to keep perspective. There will be very very very low amount of those users. And they are getting lower lower lower every day.

> choco uninstall -x packageName, it would then also remove the lib directory of chocolatey 

If this happens @ferventcoder , it probably shouldn't. Did you try it ? Chocolatey is added as first package  during installation, it doesn't make sense it gets ever uninstalled.

PR #622, opened 2017-03-13
(GH-621) PowerShell functions - allow for ignored parameters and splatting

### LINKED ISSUE #621: PowerShell functions should leave room for future expansion
In Chocolatey, when we write functions, we want to ensure there is always room for future expansion. For functions that could ultimately be placed into Chocolatey core code or load up over core code (like Get-UninstallRegistryKey), we want to ensure that they follow the same rules.

ferventcoder:
It is hard to know how future expansion of a PowerShell function is
going to look. When a function is created, it is sometimes set with a
certain number of parameters. When not using `CmdletBinding` other
parameters can be passed and ignored. However, when `CmdletBinding` is
opted into explicitly or implicitly, it doesn't automatically allow for
additional ignored arguments to be passed through. In fact, it out right
errors when you pass more than the arguments that are present. It also
does not allow for splatting with more parameters than are used for a
particular function. This is suboptimal.

Fortunately, for functions with `CmdletBinding`, an additional catch all
parameter can be added to the end to see the other parameters and add
them as part of the function. This allows for future expansion and
allows for older versions of the function to gracefully adapt when
new parameters have been added to the function (they simply ignore
what they don't know about).

Add `IgnoredArguments` to each of the functions in the extensions to
allow for gracefully allowing future expansion and for splatting.

This was seen in Chocolatey's code base in https://github.com/chocolatey/choco/commit/4ee8818d505df99b550041b7dc2c82b3bd909f4f
when it added documentation and needed to enable CmdletBinding to
produce proper docs.

Fixes #621

### DIFF
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Use `$IgnoredArguments` in all functions to allow for future expansion and splatting ([#621](https://github.com/chocolatey/chocolatey-coreteampackages/issues/621))
 
 ## 1.1.0
-- `Get-AvailableDriveLetter` - Get the next unused drive letter
+- `Get-AvailableDriveLetter`: Get the next unused drive letter
 
 ## 1.0.7
 - Bugfix in `Get-PackageParameters`: flags can now have numbers in their names, whereas before, everything past the number would be truncated and the flag would turn into a boolean.
@@ -44,4 +44,4 @@
 
 ## Version 0.1.3
 
-- `Get-WebContent` -  Download file with choco internals.
+- `Get-WebContent`:  Download file with choco internals.
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 1.2.0
+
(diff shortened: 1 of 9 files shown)


ferventcoder:
FYI - I noticed there is some inconsistencies in the functions (formatting, documentation methods, etc). I think we should address making those things consistent as well. It's not a lot of work to do but there is still some things there. And add that to our wiki.

ferventcoder:
> I'm not sure if this should be pushed right away though.
> If it should, could you also update the version and changelog.md?

I was thinking that would be another PR independent of the changes for this ticket? No matter, I can update the changelog and version as an additional commit.

ferventcoder:
@chocolatey/chocolateycommunitypackages pushed up changes

ferventcoder:
@AdmiringWorm what is required for this to be pushed? https://chocolatey.org/packages/chocolatey-core.extension looks like it is currently on the old version. I fully admit it's only been an hour since it was merged.

AdmiringWorm:
@ferventcoder create an empty commit with the title `[PUSH chocolatey-core.extension]`.
That will then push the package as it is (without going through the update process for AU packages).

AdmiringWorm:
I hoped getting #612 into the same update though, if there isn't any changes to it I need to make.

ferventcoder:
Is #612 ready to go? Not that this is super high priority, I just am working against a bug on it right now.

AdmiringWorm:
I believe it is. If it isn't I'm not sure what more needs to be done....

PR #629, opened 2017-03-16
Flash Packages Update to Latest Version

RedBaron2:
Do to a 'error' in the processing of the update.ps1 file at this time. The following packages didn't get updated to the Latest Version(s). This will bring the packages up to date.

@AdmiringWorm


majkinetor:
This is automatic package and as such its AU updater should be fixed, packages should not get manually updated. See here:

https://gist.github.com/gep13/bd2eaa76f2a9ab739ca0544c502dca6e#errors

majkinetor:
I see that you already did some work on AU, what is wrong with submitting only update.ps1 changes ?

PR #633, opened 2017-03-17
[sysinternals] Sha2sum seems to be changed upstream

netdata:


AdmiringWorm:
@netdata thank you for your PR, unfortunately we do not accept PR's that only changes the checksums (these are auto calculated by when packages are updated using AU).

Am I correct to assumre that the zip file is the only thing that have changed, and not a new version that was released. (The former we can force an update, the later would imply our update script have broken).

I'm closing this though, as it won't be accepted with only a checksum change.

AdmiringWorm:
@netdata sysinternals package is in the process of getting updated now, the latest version also adds support for nano servers (if choco supports it).

PR #635, opened 2017-03-17
(sysinternals) Added support for nano server

AdmiringWorm:
Also removed duplicate urls/checksums that isn't needed.

Required to merge with `[AU sysinternals]` in the merge title

### DIFF
--- a/automatic/sysinternals/sysinternals.nuspec
+++ b/automatic/sysinternals/sysinternals.nuspec
@@ -4,7 +4,7 @@
   <metadata>
     <id>sysinternals</id>
     <title>Sysinternals</title>
-    <version>2017.02.17</version> 
+    <version>2017.2.17</version>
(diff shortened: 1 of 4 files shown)


AdmiringWorm:
> Looks awesome.

Thanks, I'm not sure if choco even runs on nano servers though, but figured it couldn't hurt to add support for it in the package.

majkinetor:
Choco will run on nano if it doesn't now.

gep13:
It is on the road map, but doesn't work currently.

AdmiringWorm:
@gep13 I see, well this is one package that don't need to be updated at least when choco supports nano servers

gep13:
>@AdmiringWorm said...
this is one package that don't need to be updated at least when choco supports nano servers

Nice!  In order to support Nano, Chocolatey needs to be ported to run on .Net Core.

Issue #146, opened 2016-02-13
Fix missing licenseUrls


gep13:
- [x] activepresenter\activepresenter.nuspec
- [x]  absolute-uninstaller\absolute-uninstaller.nuspec
- [x]  chromium\chromium.nuspec
- [ ]  feeddemon\feeddemon.nuspec
- [x]  h264tscutter\h264tscutter.nuspec
- [x]  haali-media-splitter\haali-media-splitter.nuspec
- [ ]  kingsoft-office-free\kingsoft-office-free.nuspec
- [x] lockhunter\lochunter.nuspec
- [x]  nlite\nlite.nuspec
- [ ]  no-ip-duc\no-ip-duc.nuspec
- [x]  python\python.nuspec
- [x]  python-x86_32\python-x86_32.nuspec
- [x]  python2\python2.nuspec
- [x]  python2-x86_32\python2-x86_32.nuspec
- [x]  python3\python3.nuspec
- [x]  python3-x86_32\python3-x86_32.nuspec
- [x]  rukerneltool\rukerneltool.nuspec
- [x]  speccy\speccy.nuspec
- [x] tuniac\tuniac.nuspec
- [x]  universal-extractor\universal-extractor.nuspec
- [x]  rdcman\rdcman.nuspec
- [x]  truecrypt-langfiles\truecrypt-langfiles.nuspec
- [x]  vcredist2015\vcredist2015.nuspec
- [ ] vscode-ember-frost\vscode-ember-frost.nuspec

gep13:
These were found using:

```powershell
Get-ChildItem C:\github\Organisations\chocolatey\chocolatey-coreteampackages -Recurse -Include *.nuspec | 
    ForEach-Object { 
        if(-not(Select-String -InputObject $_ -Pattern "<licenseurl>")){
            Write-Host "Found $_"; 
        }
    }
```

majkinetor:
Since this is almost done, I will close this in a month or two if there is no further work on remaining 2 packages.

AdmiringWorm:
@majkinetor it seems like several of the packages missing licenseUrl was incorrectly checked and a few was missing.
As such I don't believe this should be pending closure just yet, it's good to have a reference of which packages is missing some recommended metadata.

majkinetor:
Some are unlisted and hence no longer apply, restoring those.

majkinetor:
Lockhunter is checked now since it doesn't actually provide any license.

PR #637, opened 2017-03-19
(rukerneltool) Added missing licenseUrl

MeadMA:
See <https://github.com/chocolatey/chocolatey-coreteampackages/issues/146>.  Unfortunately, the license is just thrown into the Quick Start guide, which is available only in German.


pascalberger:
LGTM

PR #640, opened 2017-03-21
(flashplayer*) fix dot selection for XML object

octomike:
Sorry for the fixup.. I just noticed the dot notation doesn't work with Powershell v2. Is this an issue? The explicit index for the XML collection solves this.


AdmiringWorm:
@octomike since the update.ps1 script isn't provided with the package, but is only run on appveyor (and on our computers whenever we test the update script) this is not an issue.

AU isn't backwards compatible with POSH V2. I think it requires V4+, or V5+.

As such, this PR isn't really necessary. But I can merge it if you still want to.

octomike:
Thanks for the explanation! I think I'll just close this then.

PR #641, opened 2017-03-21
(qbittorrent) Add support for 64-bit installer

### LINKED ISSUE #587: (qbittorrent) Added support for 64bit
It was requested on disqus to add support for qbittorrent 64bit:

https://chocolatey.org/packages/qbittorrent#comment-3155134209

AdmiringWorm:
fixes #587

### DIFF
--- a/automatic/qbittorrent/qbittorrent.nuspec
+++ b/automatic/qbittorrent/qbittorrent.nuspec
@@ -6,7 +6,7 @@
     <title>qBittorrent</title>
     <version>3.3.11</version>
     <authors>Christophe Dumez</authors>
-    <owners>nconrads, chocolatey</owners>
+    <owners>chocolatey,nconrads</owners>
     <summary>qBittorrent is a free software cross-platform BitTorrent client GUI written with Qt4.</summary>
     <description>
 qBittorrent is a free software cross-platform BitTorrent client GUI written with Qt4.
@@ -45,6 +45,9 @@ it for the BitTorrent protocol itself. *(from Wikipedia, 2013-04-18)*
 - IPv6 compliant
 - Sequential downloading (aka "Download in order")
 - Available on most platforms: Linux, Mac OS X, Windows, OS/2, FreeBSD
+
+## Notes
+- This version includes the 64-bit version of qbittorrent, if you wish to continue using the 32-bit version you need to pass `--x86` when calling `choco install/update`
     </description>
     <projectUrl>http://www.qbittorrent.org/index.php</projectUrl>
     <projectSourceUrl>https://github.com/qbittorrent/qBittorrent</projectSourceUrl>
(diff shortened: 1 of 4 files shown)


majkinetor:
Any reason for not using `Get-RemoteFiles` ?

AdmiringWorm:
Don't work with sourceforge urls

AdmiringWorm:
@majkinetor  basically just get the following exception if trying to use the `Get-RemoteFiles` helper:
```
Exception calling "DownloadFile" with "2" argument(s): "An exception occurred during a WebClient request."
```
https://gist.github.com/anonymous/606af1600f8672c76c49e66ea133871f/5835d1742938bf26c6b23d28b2721ee560cf5509

majkinetor:
It works for me, see for example: 

* https://github.com/majkinetor/au-packages/blob/2b47f191dfc281e471a1f0ef59ab9a518015ab88/smplayer/update.ps1#L22 
* https://github.com/majkinetor/au-packages/blob/f5c41bc2bb862bf09c8e73020b6aef5c768b8ac7/plantuml/update.ps1#L17

AdmiringWorm:
not for me on the qbittorrent package, both with and without the `-FileNameSkip` parameter.

majkinetor:
Get-RemoteFiles tries to deduce *everything* by itself. You need to help it with funky sites like this, as it provides parameters to override everything it does.

In this case, it can't deduce extension so it fails. I should probably handle this a bit better as it takes extension from URL and not the final file name but just adding `FileType` in Latest solves this issue.

AdmiringWorm:
ah, I see. I didn't think setting the `FileType` property would matter since it was the `DownloadFile` that failed.
Thanks alot.

majkinetor:
Besides that, with suffix (_x32, _x64) you can reduce 6 more lines as you don't need to have replacement in chocolateyInstall.ps1. 

chocolateyInstall should be generic with Get-RemoteFiles, only silent args are changed basically.

majkinetor:
DownloadFile failed because of invalid file path to save to

AdmiringWorm:
> Besides that, with suffix (_x32, _x64) you can reduce 6 more lines as you don't need to have replacement in chocolateyInstall.ps1

I know, but I have a preference of being explicit with which files to use.

majkinetor:
I see, if you want to add 5 minutes to each package you make, fine by me.

This is my generic embeddable exe installer which always works without any changes except silentArgs.

```powershell
$toolsDir      = Split-Path $MyInvocation.MyCommand.Definition
$embedded_path = if ((Get-ProcessorBits 64) -and $env:chocolateyForceX86 -ne 'true') {
         Write-Host "Installing 64 bit version"; gi "$toolsDir\*_x64.exe"
} else { Write-Host "Installing 32 bit version"; gi "$toolsDir\*_x32.exe" }

$packageArgs = @{
  packageName    = ''
  fileType       = $fileType
  file           = $embedded_path
  silentArgs     = '/S'
  validExitCodes = @(0, 1223)
}
rm $toolsDir\*.$fileType -ea 0
```

PR #655, opened 2017-03-29
(VoiceBot) Moving to core packages

Boyan-Kostadinov:


Boyan-Kostadinov:
Will redo without the extra dropbox files.

PR #661, opened 2017-04-01
(7zip.install) improved and added shim

### LINKED ISSUE #549: (7zip) 7z shim creation
Can the 7zip.install package also install a shim for the `7z.exe` executable? That is, I would like to have `7z.exe` on the system `PATH` has `choco.exe`.

majkinetor:
- Simplified installer
- Added shim (closes #549)
- Improved nuspec
- Removed 0 prefix on version

**Notes**:
- I removed helper because that check seems unnecessary - explorer is either running or not.
- This package can now be used as 7zip dependency instead of 7zip.portable

### DIFF
--- a/automatic/7zip.install/7zip.install.nuspec
+++ b/automatic/7zip.install/7zip.install.nuspec
@@ -3,7 +3,7 @@
   <metadata>
     <id>7zip.install</id>
     <title>7-Zip (Install)</title>
-    <version>16.4.0</version>
+    <version>16.04</version>
     <authors>Igor Pavlov</authors>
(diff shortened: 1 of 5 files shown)


AdmiringWorm:
Only thing I ought to mention is that this one isn't a direct replacement of the 7zip.portable when it comes to shims.
I believe the 7zip.portable have a few other binaries it shims on installation.

majkinetor:
Yes,, but it is not something of importance in general case. I looked for the `7z` vs `7za` difference and it seems there is none functional level (different linking).

PR #662, opened 2017-04-02
cleaned unlisted packages and outdated ketarin files

majkinetor:


majkinetor:
That is not important - there will be no work on this package due to priorities (and AFAIK development stopped). It can stay listed, that is totally beside the point.

majkinetor:
You could also remove chocolatey from the list of maintainers and let it stay with Purity.

AdmiringWorm:
as long as it's listed it shouldn't be removed from the repo.
Even if none of us is planning to work on it, doesn't mean that no-one will at some time.

majkinetor:
Then unlist it, what is the problem in that ?

majkinetor:
Then remove chocolatey user from the maintainers, simple as that.

AdmiringWorm:
@majkinetor I rather think it should be moved to the manual directory and just remove the ketarin file.
Since the package is still working, there isn't any reason to remove it from the repo.

majkinetor:
Not exactly working - it requires checksum fix.

AdmiringWorm:
true, it's missing checksum.
but since the download link is redirected to https, checksums isn't really required on those just yet.
Hence it works.

majkinetor:
Ah, ok. OK then, I'll move it to the manual package.

PR #670, opened 2017-04-12
(build) Enabled running tests on pull requests

AdmiringWorm:
With this PR we can finally enable tests for updating, installing, uninstalling package(s) on appveyor when a PR is submitted.

The benefit of this is to notice early on when a package update/install/uninstall is failing (although, some positive failures may still occur).
One drawback though, is that the install/uninstall is only subjective and should still also be tested in a choco test environment.

Feel free to test these changes out by opening a PR at https://github.com/AdmiringWorm/chocolatey-coreteampackages/pulls that is based of the test-pkgs-on-pr branch.

@gep13 this is what I suggested to you on sunday, it also requires the Pull Request web hook to be enabled here on github (as I mentioned on gitter).

### DIFF
--- a/scripts/Test-RepoPackage.ps1
+++ b/scripts/Test-RepoPackage.ps1
@@ -1,0 +1,633 @@
+﻿<#
+.SYNOPSIS
+  Test the au updating of changed packages, as well as the install and uninstall of those packages.
+
+.DESCRIPTION
+  This script uses git to get all the packages that have changed since it diverged from the
+  master branch and runs those packages through au updating (only if they are automatic packages),
+  tries to install them and also tries to uninstall those packages.
+  It also accepts an optional packageName that overrides the check for changed packages.
+
+.PARAMETER packageName
+  An optional value accepting a single package name.
+  If this variable is not specified, or if it's set to $null it gets the package
+  name from changed files.
+
+.PARAMETER type
+  An optional variable that can be used to specify which kind of tests that should
+  be run.
+  If the value is set to 'all' (the default) all tests will be run.
+
+.PARAMETER CleanFiles
+  If this variable have been specified the script will clean all
+  chocolatey logs, au output files and the nupkg files already existing
+  in the repository before running the tests.
+
+.PARAMETER TakeScreenshots
+  If this variable have been provided, the script will take a screenshot
+  of the user desktop after install/uninstall as well as when the message
+  'Failures' have been outputted by the choco executable.
+
+.PARAMETER chocoCommandTimeout
+  The timeout that should be passed to choco when installing/uninstalling
+  packages (defaults to 600 seconds, or 10 minutes)
+
+.PARAMETER timeoutBeforeScreenshot
+  The amount of seconds to sleep after installing/uninstalling a package
+  and before taking the screenshot.
+
+.PARAMETER screenShotDir
+  The directory to where the scrip should save all taken screenshots.
+  (The directory does not need to exist before running the script)
+
+.EXAMPLE
+  .\Test-RepoPackage.ps1
+#>
+param(
+  [string]$packageName = $null,
+  [ValidateSet('all','update','install','uninstall','none')]
+  [string]$type = 'all',
+  [switch]$CleanFiles,
+  [switch]$TakeScreenshots,
+  [int]$chocoCommandTimeout = 600,
+  [int]$timeoutBeforeScreenshot = 1,
+  [string]$screenShotDir = "$env:TEMP\screenshots"
+)
+
+function WriteOutput() {
+<#
+.SYNOPSIS
+  Simple helper function to simplify
+  the different output methods we use.
+#>
+  param(
+    [Parameter(ValueFromPipeline = $true)]
+    [string]$text,
+    [ValidateSet('Error','Warning','ChocoError','ChocoWarning','ChocoInfo','Normal')]
+    [string]$type = 'Normal'
+  )
+   switch -Exact ($type) {
+    'Error' { Write-Error $text }
+    'Warning' { Write-Warning $text }
+    'ChocoError' { Write-Host $text -ForegroundColor Black -BackgroundColor Red }
+    'ChocoWarning' { Write-Host $text -ForegroundColor Black -BackgroundColor Yellow }
+    'ChocoInfo' { Write-Host $text -ForegroundColor White -BackgroundColor Black }
+    Default { Write-Host $text }
+  }
+}
+
+function WriteChocoOutput() {
+<#
+.SYNOPSIS
+  A simple helper function for processing the output of choco.
+  Only meant to handle the colorization of text.
+#>
+  param(
+    [Parameter(ValueFromPipeline = $true)]
+    [string]$text
+  )
+  begin {}
+  process {
+    switch -Regex ($text) {
+      '^\s*ERROR|\s*Failures|^Uninstall may not be silent' { WriteOutput $text -type ChocoError }
+      '^\s*WARNING' { WriteOutput $text -type ChocoWarning }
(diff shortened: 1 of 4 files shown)


AdmiringWorm:
@chocolatey/community-maintainers anyone up for review this PR?
any feedback would be greatly appreciated, as well as testing towards the previously mentioned fork for scenarios I forgot/didn't think about.

Of course, non-team members are also welcome to review and test this.

scripts/Test-RepoPackage.ps1
```diff
@@ -0,0 +1,633 @@
...
+  $pathsToClean = @(
+    "$env:ChocolateyInstall\logs\*"
+    "$env:TEMP\chocolatey\au\*"
+    "$PSScriptRoot\..\Update-Force-Test*.md"
+  )
+  if ($TakeScreenshots) { $pathsToClean += @("$screenShotDir\*") }
+  $pathsToClean += Get-ChildItem -Recurse "$PSScriptRoot\.." -Filter "*.nupkg" | select -Expand FullName
+
+  $pathsToClean | % {
+    if (Test-Path $_) { rm $_ -Recurse }
```
majkinetor: Better: `rm $pathsToClean -ea 0 -Recurse`

scripts/Test-RepoPackage.ps1
```diff
@@ -0,0 +1,633 @@
...
+#>
+  param(
+    [Parameter(Mandatory = $true)]
+    [string]$packageDirectory
+  )
+  $packageName = $PackageDirectory -split '[\/\\]' | select -last 1
+  $nuspecPath = Get-Item "$PackageDirectory\*.nuspec"
+  $content = Get-Content -Encoding UTF8 $nuspecPath
+  $Matches = $null
+  $content | ? { $_ -match "\<dependency.*id=`"(${packageName}.(install|portable|app|commandline))" } | select -First 1 | WriteOutput
```
majkinetor: Better to use `[xml]` here, dep may be on multiple lines.

scripts/Test-RepoPackage.ps1
```diff
@@ -0,0 +1,633 @@
...
+if ($TakeScreenshots) { . "$PSScriptRoot\Take-ScreenShot.ps1" }
+$arguments = @{
+  packages = $packages
+  chocoCommandTimeout = $chocoCommandTimeout
+  takeScreenshot = $TakeScreenshots
+  screenShotTimeout = $timeoutBeforeScreenshot
+  screenShotDir = $screenShotDir
+}
+
+switch -Exact ($type) {
```
majkinetor: This better not be switch, the code is not DRY because of it.
Better:

```
if ('update','all' -contains $type) { ... }
if ('install','all' -contains $type) { ... }
if ('unintsall', 'all' -contains $type) { }
```

scripts/Test-RepoPackage.ps1
```diff
@@ -0,0 +1,633 @@
...
+  } | ? { $_ -ne $null -and $_ -ne '' }
+}
+
+if ($packageName) {
+  $packages = GetPackagesFromName -packageName $packageName
+} else {
+  $packages = GetPackagesFromDiff -diffAgainst 'origin/master'
+}
+
+$packages = RemoveDependentPackages -packages $packages
```
majkinetor: explain ?

scripts/Test-RepoPackage.ps1
```diff
@@ -0,0 +1,633 @@
...
+
+  if (!(Test-Path env:\APPVEYOR)) {
+    return
+  }
+  if ($ExitCode -ne 0) {
+    $host.SetShouldExit($ExitCode)
+  }
+}
+
+function UploadAppveyorArtifact() {
```
majkinetor: I don't like this is tied to AppVeyor, means it can't be used on local machine.

scripts/Test-RepoPackage.ps1
```diff
@@ -0,0 +1,633 @@
...
+  try {
+    RunChocoPackProcess '' | WriteChocoOutput
+    $failureOccurred = $false
+    $previousPercentage = -1;
+    $progressRegex = 'Progress\:.*\s+([\d]+)\%'
+    . choco $args | % {
+      $matches = $null
+      if ($failureOccurred) { WriteOutput $_ -type ChocoError }
+       # We are only showing progress per 10th value
+      elseif([regex]::IsMatch($_, $progressRegex)) {
```
majkinetor: What is this progress thingie ?

majkinetor:
Its a good job nevertheless, the effort is evident.  🥇 🍨

scripts/Test-RepoPackage.ps1
AdmiringWorm: I did try that first, however it deleted a few files that it wasn't supposed to delete.
I still can't figure out why though, but worked as expected when I made it like this.

Also, I believe using -ea 0 would be incorrect, as we would probably want to error out if a file isn't possible to delete, like if it's locked or if it requires some permissions that the user don't have.

scripts/Test-RepoPackage.ps1
AdmiringWorm: true, It could be on multiple lines. I've yet to see I package that have split the dependency element on multiple lines though, so didn't bother using `[xml]` for this as it seemed unnecessary.

scripts/Test-RepoPackage.ps1
AdmiringWorm: when a package is a meta only package, we add the dependent package to its `DependentPackage` attribute. If the dependent package also have been changed, it would also end up in the `$packages` array, meaning it would be tested twice (and could fail especially on uninstall because of this).
So we remove any packages set as a `DependentPackage` from the `$packages` array to prevent that.

scripts/Test-RepoPackage.ps1
AdmiringWorm: The script is meant to be run on appveyor, with that said though.
This function doesn't do anything if the `$env:APPVEYOR` haven't been set.
So it's possible to run it on a local machine as well.

scripts/Test-RepoPackage.ps1
AdmiringWorm: The chocolatey download progress.
I found during testing that it outputted way to much, so added changes to only output per 10th percentile.
ie `10% 20% 30% 40% 50%` etc. and only once per those percentages.

AdmiringWorm:
> I think refactoring should be done - its not that easy to follow things in a single file of 1K Posh lines.

Then be glad I continued working on it after I had something that already worked. Back then it had around double the amount of lines.

> Would definitely make this available outside the AV - on my local branch I should be able to run this test.

You can run this outside of AV.
The code related to AV would basically not do anything if the `APPVEYOR` environment variable isn't set.

AdmiringWorm:
since no changes was requested (only suggestions) I'm merging this now,
it won't be available for PR's until either @gep13 or @ferventcoder enable `Pull Request` on the appveyor webhook though.

gep13:
@AdmiringWorm I have enabled the webhook as requested.

Great work on this! 😄

scripts/Test-RepoPackage.ps1
majkinetor: Actually -ea 0 is more correct because its atomic. With test-path file can get deleted in between.

scripts/Test-RepoPackage.ps1
majkinetor: Your observation is irrelevant even if true. The fact is that it can happen. Also, its more readable IMO. Just sayin.

scripts/Test-RepoPackage.ps1
majkinetor: Good idea.

PR #672, opened 2017-04-13
(googlechrome) Fixed the detection logic for 32bit installs

### LINKED ISSUE #667: (googlechrome) Switching from 32-bit to 64-bit
Hi,

I recently ran into an issue with the `GoogleChrome` package. I noticed that the package was switching Chrome from 32-bit to 64-bit on 64-bit systems, but didn't think it would be a problem. Since it ended up causing problems for me I had to look into it further.

I tested the registry key check that the package is using and it does not return a value to indicate that Chrome is 32-bit on example systems where Chrome is in fact 32-bit.

I found [this group post](https://productforums.google.com/forum/#!topic/chrome/w93eUkzgkEo) where the poster mentions that he noticed a file that was only present for 32-bit installs. This is part of the script I ended up using to check for 32-bit vs. 64-bit and it seems to be working so far FWIW. Horrible, I know, but its what I had to do to solve our problem.

```
$os = Get-WmiObject -Class 'Win32_OperatingSystem'

if ($os.OSArchitecture -eq '64-bit') {
	$appDir = "$(${env:ProgramFiles(x86)})\Google\Chrome\Application\$($chromeVersion)"

	if (-not(Test-Path $appDir)) {
		Write-Error "Path '$($appDir)' doesn't exist."
		return
	}

	if ((Test-Path "$($appDir)\nacl64.exe") -and (Test-Path "$($appDir)\nacl_irt_x86_32.nexe")) {
		return $chromeVersion
	}

	return "$($chromeVersion) (64-bit)"
}

return $chromeVersion

```

Anyway, I thought you might want to know that this could potentially be causing issues for others. Feel free to adapt the code above if you want, or look for a better solution. Or, if not maybe put a disclaimer in the package description?

Thanks,
Bryan

AdmiringWorm:
The registry key we previously used for detecting an already installed 32bit installation have changed.
This PR fixes that, while also providing a fallback to check if the `nacl_irt_x86_32.nexe` do exist (this only exists on 32bit installations).

The check for the currently installed version had also broken, so this is also fixed.

fixes #667 

*MERGE with `[AU googlechrome]` in the title*

/cc @majkinetor @mattheyan

### DIFF
--- a/automatic/GoogleChrome/GoogleChrome.nuspec
+++ b/automatic/GoogleChrome/GoogleChrome.nuspec
@@ -22,6 +22,9 @@ Chrome is a fast, simple, and secure web browser, built for the modern web.
     <releaseNotes></releaseNotes>
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/GoogleChrome</packageSourceUrl>
     <tags>google chrome web internet browser admin</tags>
+    <dependencies>
+      <dependency id="chocolatey-core.extension" version="1.2.0" />
+    </dependencies>
   </metadata>
   <files>
     <file src="tools\**" target="tools" />
--- a/automatic/GoogleChrome/tools/chocolateyInstall.ps1
+++ b/automatic/GoogleChrome/tools/chocolateyInstall.ps1
@@ -1,7 +1,7 @@
 ﻿$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
 . $toolsPath\helpers.ps1
 
-$version = $Env:ChocolateyPackageVersion
+$version = '57.0.2987.133'
 if ($version -eq (Get-ChromeVersion)) {
     Write-Host "Google Chrome $version is already installed."
     return
--- a/automatic/GoogleChrome/tools/chocolateyInstall.ps1
+++ b/automatic/GoogleChrome/tools/chocolateyInstall.ps1
@@ -3,8 +3,8 @@
 
 $version = '57.0.2987.133'
 if ($version -eq (Get-ChromeVersion)) {
-    Write-Host "Google Chrome $version is already installed."
-    return
+  Write-Host "Google Chrome $version is already installed."
+  return
 }
 
 $packageArgs = @{
--- a/automatic/GoogleChrome/tools/helpers.ps1
+++ b/automatic/GoogleChrome/tools/helpers.ps1
@@ -1,29 +1,29 @@
-function Get-Chrome32bitInstalled {
-    $registryPath = 'HKLM:\SOFTWARE\WOW6432Node\Google\Update\ClientState\*'
-    # We also return nothing if the user forces 32bit installation
-    # as we don't need to make any checks in that case.
-    if (!(Test-Path $registryPath) -or $env:ChocolateyForceX86 -eq $true) { return }
+﻿function Get-Chrome32bitInstalled {
+  $registryPath = 'HKLM:\SOFTWARE\WOW6432Node\Google\Update\ClientState\*'
+  # We also return nothing if the user forces 32bit installation
+  # as we don't need to make any checks in that case.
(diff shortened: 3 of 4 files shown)


AdmiringWorm:
@majkinetor underpaded version?
I'm sorry, I think I understand what you're talking about.

majkinetor:
Sorry, paddedUnderVersion ? When does it kick ?

AdmiringWorm:
ah, I see.
The `$paddedUnderVersion` variable is just passed to the `Get-Padded-Version` script, basically telling it to just add additional zeroes to the revision part of the version when the version is less than the version specified in `$paddedUnderVersion`

majkinetor:
Yeah, I know how to read param descriptions, I just wonder when does it happen in nature for new versions ?

mattheyan:
Yeah, I'm wondering the same thing. Isn't the padding optional? I could see it being useful if the vendor's download URL or version recorded in installed programs (or something else) used zero padding and you had to deal with that. Does it need to be added for some other reason in this case? And why the change to stop using the `chocolateyPackageVersion` environment variable? Also, does the ` -ChecksumFor none` bit have something to do with the fact that Chrome doesn't provide historical downloads (just a wild guess)? Sorry for so many questions, but I'd love to learn more about this stuff so I can improve my own package authoring skills.  ;)

AdmiringWorm:
> Yeah, I'm wondering the same thing. Isn't the padding optional? I could see it being useful if the vendor's download URL or version recorded in installed programs (or something else) used zero padding and you had to deal with that

The reason for adding a zero padding is to allow adding fix versions of the currently available version on chocolatey.
Since we can't update an already approved version of a package we'll need to add padding to the revision part of the version.
I.e.
Version `57.0.2987.133` exist on chocolatey, We pad the revision when we need to release a fix for the package so we end up with the version: `57.0.2987.13301`. If then Google updates only the revision on their next software release (although unlikely) to: `57.0.2987.134`. Then this version would be considered lower than the padded fix version, so we also needs to pad that version to `57.0.2987.13400` so it would be considered the latest version. But when Google updates the patch part of the version so the version ends up being `57.0.2988.140`, we don't need to pad the revision part of the version number as it would still be considered newer than the fixed version.

> And why the change to stop using the chocolateyPackageVersion environment variable?

this is done because we compare it to the installed version of chrome, so if we publish a fix version of the package ie `57.0.2987.13301` it wouldn't be the same version as the version gotten from the registry, hence why we can't continue to use `chocolateyPackageVersion` environment variable.

> Also, does the -ChecksumFor none bit have something to do with the fact that Chrome doesn't provide historical downloads (just a wild guess)?

No it doesn't have anything to do with that, it is done in case of the same version that we are trying to update to exist already on appveyor, or it exist on the computer of whoever tries to debug the update script. We still acquire the checksum in the `au_BeforeUpdate` function instead of running the chocolatey install script.

mattheyan:
@AdmiringWorm Excellent! Thanks for that info. I wondered how you would handle an update to the package when the software version used all 4 components and hadn't been updated, and now I know. Do you know if this is documented somewhere, or written up in a blog post (not necessarily the `-paddedUnderVersion` param, just the general approach of padding the version)? I see mention of the problem in the [Contributing Guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#133-specify-correct-nuspecversion), but no mention of the way to "fix" it - at least nothing specific.

In terms of @majkinetor's question, I'm guessing the padding hasn't kicked in because you're going to update the `paddedUnderVersion` when the changes are reviewed and ready to go?

Thanks again for the explanation. :-)

AdmiringWorm:
>  I'm guessing the padding hasn't kicked in because you're going to update the paddedUnderVersion when the changes are reviewed and ready to go?

The padding actually will already kick in (it won't get changed though, unless `[AU googlechrome]` is used as the same version is already available on chocolatey).
With the current `paddedUnderVersion` value, everything less than version `57.0.2988` will get padded

PR #675, opened 2017-04-15
(Waterfox) Regex Change

RedBaron2:
@AdmiringWorm @gep13 @pascalberger 
The current Regex  will fail for the newest version. This Regex will help to also future proof the update script in the future.

### DIFF
--- a/automatic/waterfox/update.ps1
+++ b/automatic/waterfox/update.ps1
@@ -31,7 +31,7 @@ function global:au_SearchReplace {
 function global:au_GetLatest {
   $download_page = Invoke-WebRequest -Uri $releases -UseBasicParsing
 
-  $re    = 'Setup\.exe$'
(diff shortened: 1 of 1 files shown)


automatic/waterfox/update.ps1
```diff
@@ -31,7 +31,7 @@ function global:au_SearchReplace {
 function global:au_GetLatest {
   $download_page = Invoke-WebRequest -Uri $releases -UseBasicParsing
 
-  $re    = 'Setup\.exe$'
+  $re    = '(?:Waterfox)%20([\d]{0,2}[\.][\d]{0,2}[\.][\d]{0,2})(%20Setup\-\d+\.[ex]+)|(?:Waterfox)%20([\d]{0,2}[\.][\d]{0,2}[\.][\d]{0,2})(%20Setup)(?!\-\d+)(\.[ex]+)'
```
AdmiringWorm: Looks a little more complicated than it needs to be.
But it works for now, so no matter

PR #682, opened 2017-04-17
(nuget.commandline) AU updater

### LINKED ISSUE #681: (nuget.commandline) new package

majkinetor:
For @ferventcoder 

closes #681 

`au_GetLatest` is generic, can work with any package, just change the `$packageName`.


automatic/nuget.commandline/nuget.commandline.nuspec
```diff
@@ -0,0 +1,27 @@
...
...Â¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡Ãƒâ€šÃ‚Â¬ÃƒÆ’Ã¢â‚¬Â¦Ãƒâ€šÃ‚Â¡ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬Ãƒâ€¦Ã‚Â¡ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â doesnÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã¢â‚¬Â ÃƒÂ¢Ã¢â€šÂ¬Ã¢â€žÂ¢ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã‚Â ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬ÃƒÂ¢Ã¢â‚¬Å¾Ã‚Â¢ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬Ãƒâ€¦Ã‚Â¡ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã¢â‚¬Â ÃƒÂ¢Ã¢â€šÂ¬Ã¢â€žÂ¢ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã†â€™Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬Ãƒâ€¦Ã‚Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¬ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã‚Â¦ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¡ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬Ãƒâ€¦Ã‚Â¡ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¬ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã¢â‚¬Â ÃƒÂ¢Ã¢â€šÂ¬Ã¢â€žÂ¢ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã†â€™Ãƒâ€šÃ‚Â¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬Ãƒâ€¦Ã‚Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¬ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã‚Â¦ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¾ÃƒÆ’Ã†â€™Ãƒâ€ Ã¢â‚¬â„¢ÃƒÆ’Ã‚Â¢ÃƒÂ¢Ã¢â‚¬Å¡Ã‚Â¬Ãƒâ€¦Ã‚Â¡ÃƒÆ’Ã†â€™ÃƒÂ¢Ã¢â€šÂ¬Ã…Â¡ÃƒÆ’Ã¢â‚¬Å¡Ãƒâ€šÃ‚Â¢t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
```
AdmiringWorm: you got alot of bogus characters here.
Please replace them with `“Ω”`

automatic/nuget.commandline/nuget.commandline.nuspec
```diff
@@ -0,0 +1,27 @@
...
+    <title>Nuget.CommandLine</title>
+    <version>0</version>
+    <authors>Microsoft</authors>
+    <owners>chocolatey, nuget-mirror</owners>
+    <summary>NuGet is the package manager for the Microsoft development platforms</summary>
+    <description>NuGet is the package manager for the Microsoft development platforms including .NET. NuGet gives you access to thousands of packages from other developers on [nuget.org](http://nuget.org), and the NuGet tools let you create, share, and host packages of your own.
+</description>
+    <projectUrl>https://github.com/NuGet/NuGet.Client</projectUrl>
+    <tags>nuget dot.net microsoft foss cross-platform</tags>
+    <copyright>Â© .NET Foundation. All rights reserved.</copyright>
```
AdmiringWorm: Did you really mean to have the `Â` character here?

majkinetor:
AU does that. I think after https://github.com/majkinetor/au/pull/60

I dont understand why. Inever got to the point where encoding behaves correctly,  there is always someone who complains...

automatic/nuget.commandline/nuget.commandline.nuspec
majkinetor: They pile up after few forced updates

AdmiringWorm:
@majkinetor yes that's when it started to happen, but it shouldn't anymore (as long as the bogus characters are fixed before the next update).

Basically it happened because the character was read as an ANSI character (although it's utf-8) and then converted to UTF-8 BOM.
In that case it didn't know before it converted it what kind of character it was.

automatic/nuget.commandline/nuget.commandline.nuspec
AdmiringWorm: If you haven't fixed the character after I submitted https://github.com/majkinetor/au/pull/63, then that would be expected.
If you remove them and replace it with the omega character, then it shouldn't continue to pile up like that.

majkinetor:
> but it shouldn't anymore

I am not sure I can confirm that. But will take another look.

AdmiringWorm:
> I am not sure I can confirm that. But will take another look.

If you try to run `test_all.ps1` again for this package now, you could confirm or deny my claim.
I've seen though, after I submitted the fix it never again happened to me.

majkinetor:
Looks like appveyor check works great BTW.

One thing I noticed, screenshots are black.

AdmiringWorm:
> One thing I noticed, screenshots are black.

Hmm, I'll need to look into that.
I know it worked great when an Image wasn't specified in appveyor.yml.

Seems like there is some incompatibilities between the vs2015 and WMF 5 image.

majkinetor:
Not that it is terribly important IMO. Its probably equally effective to just return desktop content as an artifact - will show up links and anything new added.

Screenshots might come in handy with a more robust check procedure - i.e. starting the GUI programs after install to see if they show up window. This could be added too , perhaps as as a directive to AV or something. Perhaps as a new PS1 script (`verify_install.ps1`) that branch would run after choco install ?

AdmiringWorm:
@majkinetor although not terribly important, the purpose of the screenshots is to see if a program is started after the install, or is hanging on the uninstall, etc.

AdmiringWorm:
I've disabled the screenshots until I can find a working way with the WMF 5 image.

majkinetor:
Something is wrong now that this got merged. During branch testing everything was fine.

There is nothing in au temp folder which is very strange.

Only this in build log:

```
Unable to find icon url for nuget.commandline
    + CategoryInfo          : OperationStopped: (Unable to find ...get.commandline:String) [], RuntimeException
    + FullyQualifiedErrorId : Unable to find icon url for nuget.commandline
   ```
And this in info:

```powershell
PS> Import-Clixml .\update_info.xml | % result | % errors
Path          : C:\projects\chocolatey-coreteampackages\automatic\nuget.commandline
Name          : nuget.commandline
Updated       : False
Pushed        : False
RemoteVersion :
NuspecVersion : 0
Result        :
Error         : Job returned no object, Vector smash ?
NuspecPath    : C:\projects\chocolatey-coreteampackages\automatic\nuget.commandline\nuget.commandline.nuspec
NuspecXml     : #document
```

Do you spot anything unusual ?

AdmiringWorm:
@majkinetor ah damn, I missed that.
It's because the Update-IconUrl script was updated to allow skipping packages (and thus changed to throw an error when one isn't found).
I didn't notice you didn't drop the nuget.commandline icon in the icons directory.
Just drop the icon in that directory, or add a comment in the nuspec file with the text `Icon: Skip` or `IconUrl: Skip check`

AdmiringWorm:
I'll update the script mentioning how to skip the icon check for a package in the error message

majkinetor:
OK, so it was responsible for this behavior ? I'll add a comment since there is no icon.

BTW
```

pdating 1 automatic packages at 2017-04-17 14:57:39 (forced)
Push is disabled
FORCE IS ENABLED. All packages will be updated
Invalid nuspec file Version '0' - using 0.0
   nuget.commandline is updated to 3.5.0
Running Report
Saving markdown report: C:\projects\chocolatey-coreteampackages\Update-Force-Test-.md
Running Gist
  ERROR:  
    {"message":"Not Found","documentation_url":"https://developer.github.com/v3/gists/#edit-a-gist"}
```

Maybe it would not be bad to actually see report on branch au test ? We could create a new gist for that and set a link in main report via `UserMessage`, so we can quickly see full AU results... Just food for thought.

majkinetor:
Should it be a single comment anywhwere, or inside existing comment or what ?

AdmiringWorm:
it doesn't matter if it's in a single comment or not. It actually don't even need to be inside a comment.
I'm just running each line in the nuspec file through the following regex: `Icon(Url)?:\s*Skip( check)?`

AdmiringWorm:
I was sure that nuget had an icon though.
If it doesn't probably iconUrl should be removed from the nuspec file

AdmiringWorm:
perhaps reuse the icon that was used when we tested out using `nuget.portable` package.
https://github.com/pascalberger/chocolatey-coreteampackages/blob/302a1ba6bf175cc37fd23b13a131a034385fca88/icons/nuget.png

majkinetor:
Perhaps you can add an error to info object on particular package when you fail iconurl check, then it will be visible in the report.

AdmiringWorm:
Sure, I'll look into doing that.

majkinetor:
I think if you return AUPackage object and set pkg.Error it will show. I am not sure if you can actually create an object, perhaps I need to expose it as a public builder function, you will have to try it.

```
$pkg = [AUPackage]::new(  'packagename' )
$pkg.Error = 'failed icon check'
```

majkinetor:
Hm... not even that will work, the routine will return that package only if it goes to the end.

I will see how can I change AU to allow for before and after each scripts to to this.

So postpone your work until https://github.com/majkinetor/au/issues/73 gets implemented.

AdmiringWorm:
👍 will do, hadn't started yet as the food was just ready.

but what about, perhaps catching thrown errors in AU?

majkinetor:
Yeah will see to do that, but perhaps access to the AUPackage is still desirable because as scripts could add arbitrary text in results array that will show in report

AdmiringWorm:
yeah, I think that would be fine as well.

Issue #683, opened 2017-04-17
(nomacs) Add proxy during installation


majkinetor:
The guys answered me [here](https://nomacs.org/redmine/issues/765)

So we could add `--proxy` in silent args and it should work. However that is problematic as proxy can come from several places.

I think this is good moment to add to core.extension function to return current proxy. With it this would become:

```
  silentArgs  = '--script "{0}\install-script.js --proxy {1}"' -f $toolsPath, (Get-EffectiveProxy)
```

`Get-EffectiveProxy` would look into
- IE proxy
- $env:http_proxy and friends
- $env:chocolateyExplicitProxy
etc..

and return something in specific order. 

I think that we can safely ignore username and password for now, I have never actually seen people use it. Furthermore, I am not aware of any method to get IE username and password.

---
Ref: #395

PR #684, opened 2017-04-17
(core.extension) Get-EffectiveProxy

majkinetor:
ref  #683

### DIFF
--- a/extensions/chocolatey-core.extension/extensions/Get-EffectiveProxy.ps1
+++ b/extensions/chocolatey-core.extension/extensions/Get-EffectiveProxy.ps1
@@ -40,22 +40,21 @@ function Get-EffectiveProxy(){
     }
     
     # Try chocolatey config file
-    $p = 'y' | choco.exe config get proxy --limit-output
-    $p = ($p -join "`n") -replace '(?s).+Do you want to continue.+?\n'    #happens on non-administrative shell
-    $p = $p.Trim()
+    [xml] $cfg = gc $env:ChocolateyInstall\config\chocolatey.config
+    $p = $cfg.chocolatey.config | % { $_.add } | ? { $_.key -eq 'proxy' } | select -Expand value
     if ($p) {
         Write-Verbose "Using choco config proxy"
         return $p
     }
 
     # Try winhttp proxy
-     (netsh.exe winhttp show proxy) -match 'Proxy Server\(s\)' | set proxy 
-     $proxy = $proxy -split ' :' | select -Last 1
-     $proxy = $proxy.Trim()
-     if ($proxy) {
-         Write-Verbose "Using winhttp proxy server"
-         return "http://" + $proxy
-     }
+    (netsh.exe winhttp show proxy) -match 'Proxy Server\(s\)' | set proxy 
+    $proxy = $proxy -split ' :' | select -Last 1
+    $proxy = $proxy.Trim()
+    if ($proxy) {
+        Write-Verbose "Using winhttp proxy server"
+        return "http://" + $proxy
+    }
 
     # Try using WebClient
     $url = "http://chocolatey.org"
--- a/extensions/chocolatey-core.extension/extensions/Get-EffectiveProxy.ps1
+++ b/extensions/chocolatey-core.extension/extensions/Get-EffectiveProxy.ps1
@@ -1,0 +1,65 @@
+<#
+.SYNOPSIS
+    Get the current proxy using several methods
+
+.DESCRIPTION
+    Function tries to find the current proxy using several methods, in the given order:
+        - $env:chocolateyProxyLocation variable
+        - $env:http_proxy environment variable
+        - Chocolatey config
+        - IE proxy
+        - Winhttp proxy
+        - WebClient proxy
+
+    Use Verbose parameter to see which of the above locations was used for the result, if any.
+    The function currently doesn't handle the proxy username and password.
+
+.OUTPUTS
+    [String] in the form of http://<proxy>:<port>
+#>
+function Get-EffectiveProxy(){
+
+    # Try chocolatey proxy environment vars
+    if ($env:chocolateyProxyLocation) {
+        Write-Verbose "Using `$Env:chocolateyProxyLocation"
+        return $env:chocolateyProxyLocation
+    }
+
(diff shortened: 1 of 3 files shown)


majkinetor:
No username and password options. I guess it would cover 99.99% of cases this way. Although I don't know how to find it out from IE settings, AFAIK winhttp doesn't have one, http_proxy can already contain one, so it would affect only`$env:$Env:chocolateyProxyLocation` and `choco config` methods (so we can always improve this later, its a minor thing).

AdmiringWorm:
just curious though, why the name `Get-EffectiveProxy`, why not just `Get-Proxy`, or `Get-SystemProxy`?

majkinetor:
Because there are different ways to set proxy and in certain contexts one gets returned over another. There is no such thing as system proxy although IE is often considered to be the on on Windows.

> LGTM, with the exception of indentation.

I guess you missed out the talk about it. The guideline about it is more harmful then anything else and should be changed to something that would not tolerate obvious problems such as random indentation changes or styles, badly formatted code in general etc.

AdmiringWorm:
> I guess you missed out the talk about it

Yeah, I must have missed that.

> The guideline about it is more harmful then anything else and should be changed to something that would not tolerate obvious problems such as random indentation changes or styles, badly formatted code in general etc.

I disagree, unless you consider consistency harmful.
Considered it's quite easy to keep the indentation in check if you use the [editorconfig plugin](http://editorconfig.org/#download) for whatever editor you use (which is also mentioned in the guidelines: [2.2 Formatting](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#22-formatting).
The indentation of 2 spaces have also been in the `.editorconfig` file since way before both you and me started working on this repo (combined).
I do agree though that we don't need to require any changes to the indentation unless it causes obvious problems, but it still should be mentioned (which I did here, I didn't require you to change anything. I just mentioned it)

majkinetor:
> Considered it's quite easy to keep the indentation in check if you use the editorconfig plugin for whatever editor you use (which is also mentioned in the guidelines:

You must install plugins for whatever editor(s) you use, you must inform other people about editor config etc. Also, AU allows you to use Github as editor (I often do, even from phone for simple changes) - editor config plugin ?

> The indentation of 2 spaces have also been in the .editorconfig file since way before both you and me started working on this repo (combined).

That is irrelevant.  2 spaces indentation is also very strange for Powershell.

> I do agree though that we don't need to require any changes to the indentation unless it causes obvious problems, but it still should be mentioned (which I did here, I didn't require you to change anything. I just mentioned it)

Like I said, making it mandatory doesn't contribute anything to the project except negative - imagine somebody constantly beaching on you because of 2 spaces, no extra line etc. I took a chance and change this section now, and merged it with Comments section - [2.2 Code style](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#22-code-style). Programmers also often like a freedom to use their own formatting rules, taking that little pleasure from them kills the enjoyment and alienates people. We should simply not allow garbage formatting in, variations of the same thing should be allowed.

majkinetor:
On the other hand [2.1- Encoding](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#21-encoding) should probably be enforced via script. You might want to see how to add it in your branch test.

majkinetor:
One problem with `choco.exe config get proxy --limit-output`. It blocks in non admin shell with:

```
PS > choco.exe config get proxy --limit-output -y
Chocolatey detected you are not running from an elevated command shell
 (cmd/powershell).

 You may experience errors - many functions/packages
 require admin rights. Only advanced users should run choco w/out an
 elevated shell. When you open the command shell, you should ensure
 that you do so with "Run as Administrator" selected. If you are
 attempting to use Chocolatey in a non-administrator setting, you
 must select a different location other than the default install
 location. See
 https://chocolatey.org/install#non-administrative-install for details.
```

Any way to skip this ?

majkinetor:
Nvm, `"y" | config get` does it

extensions/chocolatey-core.extension/extensions/Get-EffectiveProxy.ps1
```diff
@@ -30,29 +30,31 @@ function Get-EffectiveProxy(){
...
     $key = "HKCU:\Software\Microsoft\Windows\CurrentVersion\Internet Settings"
     $r = Get-ItemProperty $key
     if ($r.ProxyEnable -and $r.ProxyServer) {
         Write-Verbose "Using IE proxy settings"
-        return $r.ProxyServer
+        return "http://" + $r.ProxyServer
+    }
+    
+    # Try chocolatey config file
+    $p = 'y' | choco.exe config get proxy --limit-output
```
AdmiringWorm: I'm sorry to say, but you know as well as I do that calls to choco will get blacklisted.
Hence I can no longer approve this PR

AdmiringWorm: perhaps use what's used in the `Get-WebFile` helper?
https://github.com/chocolatey/choco/blob/master/src/chocolatey.resources/helpers/functions/Get-WebFile.ps1#L122-L126

majkinetor: Mhm... forgot about it. I could look into XML config then instead of calling choco, what do you think ?

Your link is something different, already handled by the func as the first method.

majkinetor: Besides, anything black listed can be manually white listed I think, this is a valid reason to use choco. Isn't that so @ferventcoder ?

AdmiringWorm: > Mhm... forgot about it. I could look into XML config then instead of calling choco, what do you think ?

that's probably an option, but it's probably frowned upon (don't think it's blacklisted though).

> Your link is something different, already handled by the func as the first method.

Oh right, I just thought perhaps the `$env:chocolateyProxyLocation` would be set to the same value that is returned from `choco config get proxy`.

I think the best bet now though, is to wait on @ferventcoder's thoughts for this first.

PR #693, opened 2017-04-21
Update spotify.nuspec

westfall:
Fixes packageSourceUrl


gep13:
@AdmiringWorm @majkinetor where do we stand on the failing PR build?  The change made in this PR has no bearing on that.  Should it be another PR to fix that problem?

PR #701, opened 2017-04-24
(OpenSSL.Light) Migrated to AU

AdmiringWorm:


majkinetor:
What is the point of lastversion ?

BTW, it occurred to me that we can use similar mechanism to support multiple streams of tools (https://github.com/majkinetor/au/issues/74). I.E forcing the update if specific stream is updated. The side effect would be that last version in nuspec would change but that would not damage the procedure.

AdmiringWorm:
the point of lastversion is to store the actual version from the previous update.
Since the version uses a letter to differentiate between bug fix releases we can't rely the version we use when pushing to chocolatey.

majkinetor:
Yes you can. You convert letter to number and then you don't need all that mumbo jumbo:

`[byte][char]<letter> - 97`

AdmiringWorm:
@gep13 @pascalberger @ferventcoder 
Any thoughts on my previous comment?

automatic/OpenSSL.Light/update.ps1
```diff
@@ -0,0 +1,56 @@
...
+  $version32 = $url32 -split "$verRe" | select -last 1 -skip 1
+  $version64 = $url64 -split "$verRe" | select -last 1 -skip 1
+  if ($version32 -ne $version64) {
+    throw "32bit version do not match the 64bit version"
+  }
+
+  if (Test-Path "$PSScriptRoot\LastVersion.txt") {
+    $lastVersion = Get-Content "$PSScriptRoot\LastVersion.txt"
+    if ($lastVersion -ne $version32) { $global:au_Force = $true }
+    }
```
pascalberger: No indentation. Code is quite hard to read this way

automatic/OpenSSL.Light/OpenSSL.Light.nuspec
```diff
@@ -1,27 +1,32 @@
...
+    <version>1.1.0.20170216</version>
+    <packageSourceUrl>https://github.com/Chocolatey/chocolatey-coreteampackages/tree/master/automatic/OpenSSL.Light</packageSourceUrl>
     <owners>chocolatey</owners>
-    <version>{{PackageVersion}}</version>
+    <title>OpenSSL – The Open Source SSL and TLS toolkit</title>
     <authors>Shining Light Productions</authors>
+    <projectUrl>https://slproweb.com/products/Win32OpenSSL.html</projectUrl>
+    <iconUrl>https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/ff58a020fa163df8a14004b9802a3cb455a76eac/icons/openssl.png</iconUrl>
+    <copyright>Ⓒ 2000-2003 Shining Light Productions</copyright>
+    <licenseUrl>https://www.openssl.org/source/license.html</licenseUrl>
```
pascalberger: Is this the correct license for the Win32 distro?

automatic/OpenSSL.Light/OpenSSL.Light.nuspec
```diff
@@ -1,27 +1,32 @@
...
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <tags>openssl-light openssl ssl tls pfx pem key rsa admin</tags>
```
pascalberger: Add obligatory tags as per [§ 1.2.2](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md#122-obligatory-tags)

automatic/OpenSSL.Light/OpenSSL.Light.nuspec
AdmiringWorm: Honestly, I don't now. I just reused the license already specified.

AdmiringWorm:
> @pascalberger said...
> How is it guaranteed that license.txt is persisted between build runs?

license.txt? Or are you talking about the lastversion.txt file perhaps?
It's persisted when the package gets updated (unless an error occurs during choco push, in which case we need to check it in manually).

pascalberger:
@AdmiringWorm Yeah, sorry. I meant `lastversion.txt` of course. So it gets persisted in the AppVeyor cache?

AdmiringWorm:
@pascalberger no when the package gets update, AU calls git on the changed files in the package directory and pushes the changes when all updates is complete.

However, I'm changing this to pad the revision with '8' characters instead of using the date hack.

AdmiringWorm:
@majkinetor I updated the `update.ps1` script to rather pad the revision number, instead of the date hack used.

AdmiringWorm:
@pascalberger any chance of an updated review?

pascalberger:
@AdmiringWorm LGTM

PR #702, opened 2017-04-24
(ruby) migrated from ferventcoder

### LINKED ISSUE #700: (ruby) Migrate from ferventcoder
As one of the top downloaded packages, it should be here and updated to newest version.

[Current issues](https://github.com/ferventcoder/chocolatey-packages/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20ruby)

/cc @ferventcoder

majkinetor:
- Was manual package
- Created embedded package with AU

closes #700

Closes at @ferventcoder 
- https://github.com/ferventcoder/chocolatey-packages/pull/238
- ~~https://github.com/ferventcoder/chocolatey-packages/issues/242~~
- https://github.com/ferventcoder/chocolatey-packages/issues/57 (gem works with this version)

/cc @ferventcoder

### DIFF
--- a/automatic/ruby/README.md
+++ b/automatic/ruby/README.md
@@ -6,5 +6,5 @@ This package provides a self-contained [Windows-based installer](https://rubyins
 
 ## Package Parameters
 
-- `InstallDir` - Ruby installation directory, by default `c:\tools\RubyXY` where XY are major and minor version parts.
-- `NoPath`     - Do not add ruby bin folder to machine PATH.
+- `/InstallDir` - Ruby installation directory, by default `c:\tools\RubyXY` where XY are major and minor version parts.
+- `/NoPath`     - Do not add ruby bin folder to machine PATH.
--- a/automatic/ruby/README.md
+++ b/automatic/ruby/README.md
@@ -1,0 +1,10 @@
+# <img src="" width="48" height="48"/> [](https://chocolatey.org/packages/ruby)
+
+Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.
+
+This package provides a self-contained [Windows-based installer](https://rubyinstaller.org) that includes the Ruby language, an execution environment, important documentation, and more.
+
+## Package Parameters
+
+- `InstallDir` - Ruby installation directory, by default `c:\tools\RubyXY` where XY are major and minor version parts.
+- `NoPath`     - Do not add ruby bin folder to machine PATH.
--- a/automatic/ruby/update.ps1
+++ b/automatic/ruby/update.ps1
@@ -1,5 +1,5 @@
 import-module au
-. $PSScriptRoot\..\..\scripts\Set-DescriptionFromReadme.ps1
+Import-Module "$PSScriptRoot\..\..\scripts\au_extensions.psm1
 
 $releases = 'https://rubyinstaller.org/downloads'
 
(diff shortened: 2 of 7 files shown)


majkinetor:
This works for me, not sure why it fails on AV ...

automatic/ruby/update.ps1
```diff
@@ -0,0 +1,37 @@
+import-module au
+Import-Module "$PSScriptRoot\..\..\scripts\au_extensions.psm1
```
AdmiringWorm: missing `"` at the end
probably the reason why the appveyor run fails.

automatic/ruby/README.md
```diff
@@ -0,0 +1,10 @@
+# <img src="" width="48" height="48"/> [](https://chocolatey.org/packages/ruby)
+
+Ruby is a dynamic, open source programming language with a focus on simplicity and productivity. It has an elegant syntax that is natural to read and easy to write.
+
+This package provides a self-contained [Windows-based installer](https://rubyinstaller.org) that includes the Ruby language, an execution environment, important documentation, and more.
+
+## Package Parameters
+
+- `InstallDir` - Ruby installation directory, by default `c:\tools\RubyXY` where XY are major and minor version parts.
```
AdmiringWorm: you should add `/` to the parameter keys.
There have been several times were users have reported they couldn't use packages with parameters,
It usually turned out that they was missing the `/` because the package parameters didn't mention it.

automatic/ruby/update.ps1
majkinetor: Meh... that line i changed in speed didnt have it locally. Thx.

majkinetor:
Done, just add choco user

AdmiringWorm:
Sorry, can't add the choco user right now. The package page won't load for me, I'll have a look at it a little later.

AdmiringWorm:
@majkinetor choco have been added.

ferventcoder:
This is a great start - however there are multiple running supported versions of Ruby at the same time (like VirtualBox but more). Did those get added to core team yet?

ferventcoder:
Case in point, https://rubyinstaller.org/

* 2.2.6
* 2.3.3 (this one is being managed)
* upcoming 2.4.x

Also, where there plans to grab https://chocolatey.org/packages/ruby.portable as well?

ferventcoder:
@majkinetor issue 57 is about updating the certificates in the OLDER ruby versions, which means some additional functionality into a package version fix. HTH.

majkinetor:
@ferventcoder 

Something like this is planned: https://github.com/majkinetor/au/issues/74

ferventcoder:
I thought this was already supported in the way that we are doing it with VirtualBox?

majkinetor:
Yes, but it is a [separate package](https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/virtualbox_5.0.x). With above implementation all would go into the single package, stream would have 1 additional `au_GetLatest` function and reuse all other package data. I.E things are not dry and new package name is introduced which is not intuitive. We could go without a new package name, different AU folders could push to the same package id, but there would still need to be multiple almost identical package folders.

majkinetor:
Actually, I am a bit mistaken about VBox, I thought it pushes to the different VBox package id. AdmiringWorm did that so I wasn't totally aware how it functions up I looked at it now. 

Still, it requires a separate folder and is treated as a separate AU package:

```
	virtualbox	        False	False	5.1.22	5.1.22
	virtualbox_5.0.x	False	False	5.0.40	5.0.32
```

so I guess it does make a sense to make it easier.

AdmiringWorm:
@majkinetor although it looks like a separate package, it reuses the original virtualbox scripts, and overrides the packagename variable so it pushes to the same chocolatey package.
Of course, when/if https://github.com/majkinetor/au/issues/74 then the separate directory could be removed though. (Same thing with the php directories)

PR #708, opened 2017-04-28
(scite4autohotkey) Modernized package

AdmiringWorm:


AdmiringWorm:
@majkinetor I don't suppose you have any suggestions to the autohotkey uninstall script.
I have no knowledge regarding autohotkey, so it's the best I could manage without some feedback.

AdmiringWorm:
I know it works, but I think it could be better

majkinetor:
Yeah, it could be better. That sleep can cause it to fail in some cases because something else can hijack active window. 

But, I don't think this package will see that many users so its best not to waste time perfecting its installer for the moment (unless somebody complains).

AdmiringWorm:
Yeah, I figured as much. Hence why I already merged it.
Regarding the sleep though, it was the only way I knew how to do.
Since all the pop up windows have the same information (as picked up by autohotkey), I had to wait for a small amount to let the dialog/window show, otherwise it would just try running it all on the same dialog window.

majkinetor:
> Since all the pop up windows have the same information (as picked up by autohotkey), 

You could look for:

- all text
- number of controls

etc.

The problem is not that, its that Send is non window specific. It just sends the keys. Its better to use `ControlSend` if it works, to send to specific window. 

With sending you might want to use BlockInput to make it more bulletproof although it doesn't guarantee 100% - 3td party window may programatically steal focus.

AdmiringWorm:
@majkinetor thanks, I'll keep that in mind for the next time I need to look into using autohotkey.

majkinetor:
In this case its probably the best to change original install.ahk instead of driving it via another ahk

PR #720, opened 2017-05-09
(kb2999226) fix KB id in install script

jberezanski:
Also add missing release notes for previous version.


jberezanski:
Do not merge yet, please

jberezanski:
I just realized it not only affects info messages, but actual install logic (applicability determination), so I'll reword the descriptions.

jberezanski:
Now it's correct.

PR #731, opened 2017-05-18
(sysinternals) fixed checksum

smithyuk:
Fixed checksum to prevent install failure + updated update.ps1 script

### DIFF
--- a/automatic/sysinternals/sysinternals.nuspec
+++ b/automatic/sysinternals/sysinternals.nuspec
@@ -4,7 +4,7 @@
   <metadata>
     <id>sysinternals</id>
     <title>Sysinternals</title>
-    <version>2017.05.16</version>
+    <version>2017.5.16.20170518</version>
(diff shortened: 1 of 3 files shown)


AdmiringWorm:
I'm afraid I'll have to deny this pull request.
Checksums are automatically generated by AU (the only thing we need to do is adding a empty commit with `[AU sysinternals]`.
There is also a change in this PR that shouldn't be here, there is a reason we have the `-ChecksumFor` set to `none`, the reason being a problem with one of the functions we use in the install script, so we need to generate the checksum using a powershell function instead (see the `global:au_BeforeUpdate` function).

I'll queue the update of sysinternals shortly though (just need to test that the checksum actually have changed).

smithyuk:
No worries, thanks!

AdmiringWorm:
I still appreciate you trying to help out though.

AdmiringWorm:
The sysinternals package is going through the update process now, and should have passed the automated checks on chocolatey.org in a few hours.

Issue #715, opened 2017-05-06
Package changelogs


AdmiringWorm:
Opening this for discussion for whether we should provide changelogs for when bugs/feautres have been fixed/added to the package, and if we should start adding changelogs then which format should we use.

Personally I think it would be helpful to add changelogs, and use a dates when these changes was implemented.
I.E.
```markdown
# Package Changelog

## 2015-05-01
### Bugs
- Fixed a problem that the wrong install directory was used (Issue \#457)

### Features
- Implemented ability to enable symlinks (Issue \#578)
```

Probably shouldn't require it though, not at first. But just recommend to add one.

Thoughts @gep13 @pascalberger @majkinetor

AdmiringWorm:
Another option could perhaps be to just use a generic `## Upcoming` subheader and have it replaced with the version when it have been pushed from appveyor (using a script).

majkinetor:
Changelog.md in package root might be good to have. Should probably be included in the package as it can be viewed from the gallery.

AdmiringWorm:
Yeah, it was in the package root I was thinking as well.
Not sure about including it in the package though, think it would be better to link to the changelog in the release notes element IMO.
Even though the changelog would be viewable on chocolatey.org, if using markdown it would just be displayed as ordinary text.

AdmiringWorm:
@majkinetor I've added a PR as an example to one way I think the changelogs could be implemented.
see PR #716

majkinetor:
I think its hard to read with all that subheaders. Peraps just use Bugfix prefex ?

majkinetor:
> Not sure about including it in the package though, think it would be better to link to the changelog in the release notes element IMO.

Not sure either. One rationale to include it was that you don't have to depend on external sources in order to see it which may be important for some people, especially those that cache it.

AdmiringWorm:
> Peraps just use Bugfix prefex and just version header ?

You mean something similiar to the [changelog used in chocolatey-core.extension](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/extensions/chocolatey-core.extension/CHANGELOG.md)?

Personally I like to separate bugfixes and features/improvements to their own section (under the version), as well as having the date it was added, and I don't really find the [7zip.install changelog](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/7zip.install/Changelog.md) I added hard to read.
With that said though, formatting could definitely be improved.

[...]

PR #744, opened 2017-05-29
(ruby) Ensure unattended

### LINKED ISSUE #742: (ruby) Ruby 2.4.1.1 is not unattended
This window comes up (but it does not block the install from finishing):
image

https://github.com/oneclick/rubyinstaller2/issues/43

ferventcoder:
Ruby now uses RubyInstaller2 as of 2.4, which means a new
way of installing and setting up the DevKit, which is built into
Ruby instead of having to install separately. Unfortunately 
that also means some new changes, as a window pops up 
after the install (thankfully in a non-blocking way) to ask the
user to select how they would like to have the devkit install
proceed. The switch `noridkinstall` prevents that from 
happening. As of any version of Ruby past 2.4.1-1, this will
be the default as well, but it's better to be explicit about 
specifying this behavior.

https://github.com/oneclick/rubyinstaller2/issues/43
https://github.com/oneclick/rubyinstaller2/wiki/FAQ#user-content-silent-install

Fixes #742

### DIFF
--- a/automatic/ruby/tools/chocolateyInstall.ps1
+++ b/automatic/ruby/tools/chocolateyInstall.ps1
@@ -13,7 +13,7 @@ $packageArgs = @{
   fileType       = 'exe'
   file           = gi "$toolsDir\*_x32.exe"
   file64         = gi "$toolsDir\*_x64.exe"
-  silentArgs     = '/verysilent /dir="{0}" /tasks="assocfiles,modpath"' -f $installDir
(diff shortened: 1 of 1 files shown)


majkinetor:
Changelog is not yet in the guidelines since we just recently adopted it, but I think its good to have it here to explain what is actually going on:

https://github.com/chocolatey/chocolatey-coreteampackages/issues/715

ferventcoder:
Keep getting folks asking questions about this...

PR #747, opened 2017-05-31
(apache-httpd) migration from external repo

The-Running-Dev:


automatic/apache-httpd/apache-httpd.nuspec
```diff
@@ -0,0 +1,30 @@
...
+    <releaseNotes>http://httpd.apache.org/docs/2.4/new_features_2_4.html</releaseNotes >
+    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/apache-httpd</packageSourceUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <summary>The Apache HTTP Server Project is a collaborative software development effort aimed at creating a robust, commercial-grade, featureful, and freely-available source code implementation of an HTTP (Web) server.</summary>
+    <description>
+    </description>
+    <tags>apache httpd webserver admin</tags>
+    <dependencies>
+       <dependency id="vcredist2015" version="14.0" />
+       <dependency id="chocolatey-core.extension" version="1.0.5" />
```
majkinetor: Use latest version

automatic/apache-httpd/readme.md
```diff
@@ -0,0 +1,26 @@
+# [<img src="https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/2bdf6f7e33ec1a8126829fbbc87b83e4473b3634/icons/apache-httpd.png" height="48" width="48" /> apache-httpd](https://chocolatey.org/packages/apache-httpd)
+
+Apache HTTP Web Server for Windows.
+
+### Note
```
majkinetor: Its `## Notes` that has a bluet list for each topic

automatic/apache-httpd/readme.md
```diff
@@ -0,0 +1,26 @@
...
+This package will unzip the latest windows binary provided at Apache Haus (http://www.apachehaus.com/) with OpenSSL on Visual Studio 2012 (VC11):
+with OpenSSL 1.0.2g, nghttp 1.9.2, Zlib 1.2.8 (mod_deflate), PCRE 8.38, APR 1.5.2, APR-Util 1.5.4, IPv6 and TLS SNI enabled
+
+The complete path of the package will be ''%unzipLocation%\Apache24'' (default: C:\tools\Apache\httpd-2.4.25\Apache24).
+
+A service with a given name is created with the default config location.
+
+This package needs either a Java Runtime to run. The package has no dependencies to a java runtime to give a choice to install JRE, JDK or Server JRE.
+
+### Parameters
```
majkinetor: Its `## Package parameters`

automatic/apache-httpd/readme.md
```diff
@@ -0,0 +1,26 @@
...
+
+The following package parameters can be set:
+
+ * `/unzipLocation` - Unzip to a different destination folder. Default: C:\tools\Apache\httpd-2.4.25
+ * `/serviceName` - The name of the windows service which will be create. Default: Apache
+
+These parameters can be passed to the installer with the use of `-packageParameters `.
+For example: choco install apache-httpd `-packageParameters '"/unzipLocation:C:\tools\Server /serviceName:Apache"'`.
+
+### Install
```
majkinetor: ?

automatic/apache-httpd/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,60 @@
+$PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
```
majkinetor: The package should be embedded because it is GNU and less then 150MB

automatic/apache-httpd/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,60 @@
...
+
+$apacheDir = Join-Path $packageArgs.installLocation $apacheDir.BaseName
+$httpdConfPath = Join-Path $apacheDir 'conf\httpd.conf'
+$httpdPath = Join-Path $apacheDir 'bin\httpd.exe'
+
+# Set the server root and port number
+$httpConf = Get-Content $httpdConfPath
+$httpConf = $httpConf -replace 'Define SRVROOT.*', "Define SRVROOT ""$($apacheDir -replace '\\', '/')"""
+$httpConf = $httpConf -replace 'Listen 80', "Listen $($packageArgs.port)"
+Set-Content -Path $httpdConfPath -Value $httpConf
```
majkinetor: Should probably be ASCII encoding

automatic/apache-httpd/apache-httpd.nuspec
photodude: @majkinetor Isn't the correct way to specify the minimum for chocolatey-core.extension?

automatic/apache-httpd/update.ps1
```diff
@@ -0,0 +1,37 @@
+Import-Module au
+Import-module "$PSScriptRoot\..\..\scripts\au_extensions.psm1"
+
+function global:au_AfterUpdate { Set-DescriptionFromReadme -SkipFirst 1 }
+
+function global:au_GetLatest {
+    $downloadPageUrl = 'http://www.apachehaus.com/cgi-bin/download.plx'
+    $versionRegEx = 'httpd\-([0-9\.]+)\-x\d+\-(.*)\.zip'
+    $url32 = 'https://www.apachehaus.com/downloads/httpd-$version-x86-$vcNumber.zip'
```
majkinetor: Using `$vcNumber` before defining it.

automatic/apache-httpd/update.ps1
```diff
@@ -0,0 +1,37 @@
...
+    $matches = [regex]::match($downloadPage.Content, $versionRegEx)
+    $version = [version]$matches.Groups[1].Value
+    $vcNumber = $matches.Groups[2].Value
+
+    $url32 = $ExecutionContext.InvokeCommand.ExpandString($url32)
+    $url64 = $ExecutionContext.InvokeCommand.ExpandString($url64)
+
+    $info = @{ Url32 = $url32; Url64 = $url64; Version = $version }
+
+    write-host $($info | out-string)
```
majkinetor: There is no need to write out Latest. You see it in the output and you can see it in shell with `$Latest`.

automatic/apache-httpd/readme.md
```diff
@@ -0,0 +1,26 @@
...
+
+The complete path of the package will be ''%unzipLocation%\Apache24'' (default: C:\tools\Apache\httpd-2.4.25\Apache24).
+
+A service with a given name is created with the default config location.
+
+This package needs either a Java Runtime to run. The package has no dependencies to a java runtime to give a choice to install JRE, JDK or Server JRE.
+
+### Parameters
+
+The following package parameters can be set:
```
majkinetor: This sentence is redundant.

automatic/apache-httpd/apache-httpd.nuspec
majkinetor: No, the newer versions contain bug-fixes, its always better to use newest.

automatic/apache-httpd/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,60 @@
...
+    checksumType    = 'sha256'
+    checksumType64  = 'sha256'
+    unzipLocation   = $env:chocolateyPackageFolder
+    installLocation = $env:APPDATA
+    port            = 80
+    serviceName     = 'Apache'
+}
+
+$packageParameters = Get-PackageParameters
+if ($packageParameters.installLocation) {
```
majkinetor: This entire section can be moved in packageArgs definition i.e.
```
$pp = Get-PackageParameters
$packageArgs = @{
   ...
   port = if ($pp.Port) { $pp.Port } else { 80 }
   ....
}
```

automatic/apache-httpd/readme.md
photodude: This sentence is in many packages.

majkinetor: That still makes it redudant :) The section is about package parameters, so those can obviously be set :)

The-Running-Dev:
Alright, I'll handle the changes.

automatic/apache-httpd/readme.md
```diff
@@ -0,0 +1,26 @@
+# [<img src="https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/2bdf6f7e33ec1a8126829fbbc87b83e4473b3634/icons/apache-httpd.png" height="48" width="48" /> apache-httpd](https://chocolatey.org/packages/apache-httpd)
+
+Apache HTTP Web Server for Windows.
+
+### Note
+
+This package will unzip the latest windows binary provided at Apache Haus (http://www.apachehaus.com/) with OpenSSL on Visual Studio 2012 (VC11):
+with OpenSSL 1.0.2g, nghttp 1.9.2, Zlib 1.2.8 (mod_deflate), PCRE 8.38, APR 1.5.2, APR-Util 1.5.4, IPv6 and TLS SNI enabled
+
+The complete path of the package will be ''%unzipLocation%\Apache24'' (default: C:\tools\Apache\httpd-2.4.25\Apache24).
```
photodude: Is the install path in `AppData` or in `C:\tools\` as stated here?
The existing package installs to `C:\tools\` might be best to be consistent and maintain that install location.

majkinetor:
For this type of package you also need `chocolateyBeforeModify` script - it fires only on update and is used to stop the service, do any uninstalls etc.

See for example [virtualbox](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/virtualbox/tools/chocolateyBeforeModify.ps1) or [gitlab-runner](https://github.com/majkinetor/au-packages/tree/master/gitlab-runner/tools).

automatic/apache-httpd/update.ps1
The-Running-Dev: This doesn't need to be changed, it's not being used before it's defined. The string gets expanded after the vsNumber variable has a value.

automatic/apache-httpd/update.ps1
The-Running-Dev: Yup, just left over from debugging.

automatic/apache-httpd/readme.md
The-Running-Dev: The default is %AppData%, I've updated the readme.

The-Running-Dev:
I made most of the requested changes, I still need to embed the software and add the modify script.

automatic/apache-httpd/update.ps1
majkinetor: I meant, there is no need to do so while debugging  :) Just type $Latest in shell after the ./update.ps1 run

The-Running-Dev:
I've made all requested changes. I think the build is failing due to missing chocolatey-core extensions.

automatic/apache-httpd/update.ps1
majkinetor: Then why not define it afterwards and skip expand string?

The-Running-Dev: I can, but as it is right now, the strings are defined in one place, on top of the function, in my opinion having them in the same place makes the code more readable.

automatic/apache-httpd/tools/chocolateyBeforeModify.ps1
```diff
@@ -0,0 +1,7 @@
+$configFile = Join-Path $env:chocolateyPackageFolder 'config.json'
+$config = Get-Content $configFile -Raw | Out-String | ConvertFrom-Json
```
AdmiringWorm: You can't use ConvertFrom-Json, scripts are meant to be PSv2 compatible and ConvertFrom-Json was added in PSv3

automatic/apache-httpd/tools/chocolateyBeforeModify.ps1
```diff
@@ -0,0 +1,7 @@
+$configFile = Join-Path $env:chocolateyPackageFolder 'config.json'
```
AdmiringWorm: I'm guessing this should actually be a config.json file located in the tools directory, not the root package directory.

automatic/apache-httpd/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,46 @@
...
+
+& $httpdPath -k install -n "$($packageArgs.serviceName)"
+
+$config = @{
+    destination = $apacheDir
+    httpdPath   = $httpdPath
+    serviceName = $packageArgs.serviceName
+}
+$configFile = Join-Path $toolsDir 'config.json'
+Set-Content $configFile -Value ($config | ConvertTo-Json)
```
AdmiringWorm: You can't use ConvertTo-Json, this was added in PSv2.

automatic/apache-httpd/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,6 @@
+$configFile = Join-Path $env:chocolateyPackageFolder 'config.json'
```
AdmiringWorm: I'm guessing this should actually be a config.json file located in the tools directory, not the root package directory (this would also be the reason why appveyor fails).

automatic/apache-httpd/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,6 @@
...
+$config = Get-Content $configFile -Raw | Out-String | ConvertFrom-Json
```
AdmiringWorm: You can't use ConvertFrom-Json, scripts are meant to be PSv2 compatible and ConvertFrom-Json was added in PSv3

automatic/apache-httpd/tools/chocolateyInstall.ps1
majkinetor: Its possible to depend on `jq` package but its probably the best to just -replace it.

majkinetor:
This build output might be a problem:
```
Testing httpd.conf....
Errors reported here must be corrected before the service can be started.
```

The-Running-Dev:
This should be all set on my end.

automatic/apache-httpd/apache-httpd.nuspec
```diff
@@ -37,7 +37,7 @@ Apache will be installed as a service under the default name 'Apache'
 </description>
     <tags>apache httpd webserver admin</tags>
     <dependencies>
-       <dependency id="vcredist2015" />
+       <dependency id="vcredist2012" />
```
photodude: Apache httpd-2.4.25 has vc14, vc11, and vc9 versions. Since vc9 is eol in August 2017 I suggest moving this to be dependent on VC14.

best case scenario would be to be dependent on VC14 with the option to use VC11 (if multiple dependency options is possible)

The-Running-Dev: Sure, done.

photodude:
thank you @The-Running-Dev @majkinetor @AdmiringWorm

PR #772, opened 2017-06-26
(git) modernization and new parameter

### LINKED ISSUE #606: (git.installer) Add option to disable GCM
It seems to be possible to disable GCM on installation.

See: https://github.com/git-for-windows/build-extra/commit/8764c72d83cd05a709fcb21700e42081d226037b and https://github.com/git-for-windows/build-extra/commit/612999af6eeade875c0659c0cad1ea019a9b07d1

Therefore we should ad a parameter to disable GCM on installation to the `git.install` package.

### LINKED ISSUE #608: (git-disable-gcm) GCM is reenabled after Git update
See https://chocolatey.org/packages/git-disable-gcm#comment-3171152490.

Git installer will use the settings set from the previous installation during update. As the `git-disable-gcm`package currently sets `--system` configuration, the setting will be overwritten on the next update of the `git.install` package.

One option would be to set the setting on a `--global` level, but this would not work in scenarios where Git is used by different users than who has run the installation.

There is also another configuration place in Git for Windows under `%ProgramData%,` see [here](https://github.com/git-for-windows/git/commit/c56a3fe8dd61db332db0d2e4f65814ee2eb5112e#diff-3cae93989ea2a52edcbd6d90803b0c6eR252), but needs to be checked if `--system` or `%ProgramData%` takes precedence.

majkinetor:
Closes #606 
Closes #608

### DIFF
--- a/automatic/git.install/README.md
+++ b/automatic/git.install/README.md
@@ -1,0 +1,21 @@
+# [<img src="" height="48" width="48" /> git.install](https://chocolatey.org/packages/git.install)
+
+Git for Windows focuses on offering a lightweight, native set of tools that bring the full feature set of the Git SCM to Windows while providing appropriate user interfaces for experienced Git users and novices alike.
+
+## Features
+
+* **Git BASH**: Git for Windows provides a BASH emulation used to run Git from the command line. *NIX users should feel right at home, as the BASH emulation behaves just like the "git" command in LINUX and UNIX environments.
+* **Git GUI**: As Windows users commonly expect graphical user interfaces, Git for Windows also provides the Git GUI, a powerful alternative to Git BASH, offering a graphical version of just about every Git command line function, as well as comprehensive visual diff tools.
+* **Shell Integration**: Simply right-click on a folder in Windows Explorer to access the BASH or GUI.
+
+## Package parameters
+
+- `/GitOnlyOnPath` - Puts gitinstall\cmd on path. This is also done by default if no package parameters are set.
+- `/GitAndUnixToolsOnPath` - Puts gitinstall\bin on path. This setting will override `/GitOnlyOnPath`.
+- `/NoAutoCrlf` - Ensure _'Checkout as is, commit as is'_. This setting **only affects new installs**, it will not override an existing `.gitconfig`.
+- `/WindowsTerminal` - Makes `vim` use the regular Windows terminal instead of MinTTY terminal.
+- `/NoShellIntegration` - Disables shell integration ( _"Git GUI Here"_ and _"Git Bash Here"_ entries in context menus).
+
+## Notes
+
+- The package uses default install options minus cheetah integration and desktop icons. Cheetah prevents a good upgrade scenario, so it has been removed.
--- a/automatic/git.install/README.md
+++ b/automatic/git.install/README.md
@@ -15,6 +15,7 @@ Git for Windows focuses on offering a lightweight, native set of tools that brin
 - `/NoAutoCrlf` - Ensure _'Checkout as is, commit as is'_. This setting **only affects new installs**, it will not override an existing `.gitconfig`.
 - `/WindowsTerminal` - Makes `vim` use the regular Windows terminal instead of MinTTY terminal.
 - `/NoShellIntegration` - Disables shell integration ( _"Git GUI Here"_ and _"Git Bash Here"_ entries in context menus).
+- `/NoCredentialManager` - Disable git credential manager by adding `$Env:GCM_VALIDATE='false'` user environment variable.
 
 ## Notes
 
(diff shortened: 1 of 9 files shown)


automatic/git.install/README.md
```diff
@@ -0,0 +1,23 @@
...
+* **Shell Integration**: Simply right-click on a folder in Windows Explorer to access the BASH or GUI.
+
+## Package parameters
+
+- `/GitOnlyOnPath` - Puts gitinstall\cmd on path. This is also done by default if no package parameters are set.
+- `/GitAndUnixToolsOnPath` - Puts gitinstall\bin on path. This setting will override `/GitOnlyOnPath`.
+- `/NoAutoCrlf` - Ensure _'Checkout as is, commit as is'_. This setting **only affects new installs**, it will not override an existing `.gitconfig`.
+- `/WindowsTerminal` - Makes `vim` use the regular Windows terminal instead of MinTTY terminal.
+- `/NoShellIntegration` - Disables shell integration ( _"Git GUI Here"_ and _"Git Bash Here"_ entries in context menus).
+- `/NoCredentialManager` - Disable git credential manager by adding `$Env:GCM_VALIDATE='false'` user environment variable.
```
pascalberger: Git Credential Manager writes itself in camel case

automatic/git.install/git.install.nuspec
```diff
@@ -29,7 +29,7 @@ Git for Windows focuses on offering a lightweight, native set of tools that brin
 
 - The package uses default install options minus cheetah integration and desktop icons. Cheetah prevents a good upgrade scenario, so it has been removed.
 - For disabling the Git Credential Manager see the [Disable Git Credential Manager package](https://chocolatey.org/packages/git-disable-gcm).
```
pascalberger: We could add here that either the ` NoCredentialManager` flag or the `Disable Git Credential Manager package` could be used.

manual/git-disable-gcm/git-disable-gcm.nuspec
```diff
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
   <metadata>
     <id>git-disable-gcm</id>
-    <title>Disable Git Credential Manager</title>
+    <title>[DEPRECATED] Disable Git Credential Manager</title>
```
pascalberger: Why deprecating this? There are still valid use cases for a separate package which don't work with parameters.

majkinetor: Because it is re-enabled after git update so its not very useful.

majkinetor:
Failing build has nothing to do with the package as far as I can see.

manual/git-disable-gcm/git-disable-gcm.nuspec
pascalberger: I still hoped to get this fixed in the Git installer. But maybe its the best to deprecate it and if it once is supported fully by the Git installer we still can support it again.

automatic/git.install/git.install.nuspec
pascalberger: If we deprecate the `git-disable-gcm` package this sentence should be removed.

manual/git-disable-gcm/git-disable-gcm.nuspec
majkinetor: Even if it works, I believe it promotes bad pattern.

pascalberger: I know, we had this discussion already. I'll put similar packages in the future in another repository.

automatic/git.install/git.install.nuspec
AdmiringWorm: @pascalberger ignore the description inside the nuspec file, have a look at the Readme.md instead.
You see there that this sentence is removed (and if you look at the update.ps1 file, the description gets updated from the Readme.md file).

automatic/git.portable/README.md
```diff
@@ -0,0 +1,9 @@
+# [<img src="" height="48" width="48" /> git.portable](https://chocolatey.org/packages/git.portable)
```
AdmiringWorm: @majkinetor you may wish to update the img src url here manually, if I remember correctly the icon url in the nuspec file is up to date, so this one won't be updated.

automatic/git.install/README.md
```diff
@@ -0,0 +1,23 @@
+# [<img src="" height="48" width="48" /> git.install](https://chocolatey.org/packages/git.install)
```
AdmiringWorm: @majkinetor you may wish to update the img src url here manually, if I remember correctly the icon url in the nuspec file is up to date, so this one won't be updated.

Issue #777, opened 2017-06-28
(winscp) migrate from dtgm


majkinetor:
Current package outdated. No open issues at @dtgm

Intention is to:

- embed winscp itself.
- modernize it

majkinetor:
I decided to leave out language stuff as it always require network connection due to large number of localisations. This can eventually be added later as a separate package.

PR #778, opened 2017-06-29
(winscp) GH-#777

majkinetor:
#777

PR @ dtgm: https://github.com/dtgm/chocolatey-packages/pull/309


majkinetor:
For some reason the build fail because link is unreachable from AV (in `au_BeforeUpdate`) although it works locally for me. I logged in on AV and saw that the page returned is different, it might be some regional settings, but didn't have time to investigate fully the solution.

@AdmiringWorm , could you perhaps run the updater at your location to see if the error reaperars ?

AdmiringWorm:
@majkinetor I don't have the time right now to check, but I'll do it as soon as possible

majkinetor:
@AdmiringWorm , nvm I switched downloads to use sourceforge, didn't notice the links before. Now it works. Just add choco user when you have time. Thx.

Issue #782, opened 2017-07-05
(notepadplusplus.install) Error with using software name wildcard pattern in match


michaelray:
(notepadplusplus.install)  When attempting to install the notepadplusplus.install Chocolatey package, I receive an error "ERROR: Bad argument to operator '-match': parsing "Notepad++*" - Nested quantifier +..".  This package was previously installing/updating on the same computer without issue up until relatively recently.

In a relatively recent commit ( 2d0768c80f73defe862f770322a474aaf4fbde52 ), a change was made to use "Notepad++*" (wildcard pattern-based) instead of "notepad\+\+" (regex pattern-based) as this new "softwareName" key on the packageArgs.

My thought is that this softwareName should have stayed as the regex pattern-based string ("notepad\+\+") instead of the wildcard pattern-based ("Notepad++*") because the called function that uses the parameter is Get-AppInstallLocation ... and the description for the first $AppNamePattern argument to that Get-AppInstallLocation function is "Regular expression pattern".  The Install-ChocolateyPackage function doesn't use the "softwareName" argument key as far as I can tell, so it is only used for getting the apps install location after installation has completed.

I'll submit a PR for this momentarily and reference this issue.

* Side note:  A separate issue appears to exist with Get-AppInstallLocation, where functions/operators that use wildcard patterns and regex patterns are being passed the same argument value interchangeably.. and I intend to submit a separate issue for that.  While it is related to this issue, again my thought is that since Get-AppInstallLocation says that the first parameter should be a regex, this should be adjusted to reflect how it is currently designed to function (and use a regex instead of wildcard).

* Package Version used: 7.4.2
* Operating System and version: Windows 7 Pro (Windows v6.1.7601.65536)
* Chocolatey version: v0.10.7
* Install/uninstall gist:
https://gist.github.com/michaelray/fa3ad48d81026988810923226d86e7bf

PR #783, opened 2017-07-05
(notepadplusplus.install) Fix error pattern match

michaelray:
Fix an error where a wildcard pattern is being used instead of a regex
pattern when calling Get-AppInstallLocation.

softwareName should have stayed as the regex pattern-based string ("notepad++") instead of the wildcard pattern-based ("Notepad++*") because the called function that uses the parameter is Get-AppInstallLocation ... and the description for the first $AppNamePattern argument to that Get-AppInstallLocation function is "Regular expression pattern". The Install-ChocolateyPackage function doesn't use the "softwareName" argument key as far as I can tell, so it is only used for getting the apps install location after installation has completed.  (More details in the submitted issue related to this.)

## Types of changes

## Checklist:

- [ ] All files is up to date with the latest [Contributing Guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)

Issue submitted for this:
https://github.com/chocolatey/chocolatey-coreteampackages/issues/782

### DIFF
--- a/automatic/notepadplusplus.install/tools/chocolateyInstall.ps1
+++ b/automatic/notepadplusplus.install/tools/chocolateyInstall.ps1
@@ -17,7 +17,7 @@ $packageArgs = @{
   checksumType64 = 'sha256'
   silentArgs     = '/S'
   validExitCodes = @(0)
-  softwareName   = 'Notepad++*'
+  softwareName   = 'Notepad\+\+'
 }
 Install-ChocolateyPackage @packageArgs
 


majkinetor:
This is better to fix upstream in the ` Get-AppInstallLocation` since when it happens this will stop working again.

michaelray:
I disagree because the function that this notepadplusplus.install package is calling (Get-AppInstallLocation) specifies a "Regular expression pattern" for the $AppNamePattern argument, which is the function that this notepadplusplus.install package is calling, except the notepadplusplus.install package is not providing a "regular expression pattern" as the argument and is instead passing a wildcard expression.

A fix upstream in Get-AppInstallLocation would break on the -match operation used if it is changed to take a wildcard pattern and, further, other callers (in other packages) would already be passing regular expression patterns.  The easiest fix that produces the least amount of potential breakage in other packages using it already would likely be to use a separate argument for a wildcard pattern if Get-AppInstallLocation is to be changed to support that.

In the meantime, I believe the packages (like notepadplusplus.install) should be passing arguments to the called functions (like Get-AppInstallLocation) based on what is actually specified by that function being called.

When testing this change on my local machines, I saw success on the machines where I was previously receiving error & the machines that were previously succeeding using notepadplusplus.install continued to succeed with this change.

Using a wildcard pattern results in Get-AppInstallLocation producing an error when it is not able find the install path from the registry first.

Using a regular expression pattern (as this notepadplusplus.install package previously was but was recently changed to use the wildcard pattern) causes no errors with calling Get-AppInstallLocation.. the registry is initially checked and it results in no registry keys returned (because the function it is calling expects a wildcard pattern), but it then checks different locations without error.

PR #794, opened 2017-07-18
(texmaker) Updated to support the new msi installation

### LINKED ISSUE #793: (texmaker) New version only supports 64bit, and uses msi installer
As the title says, the latest version (5.5.0.0) of the texmaker available on their site seem to only support 64bit and not 32bit anymore, not completely sure what would be the best way to support this hence the opening of this issue.
The installer have also been changed to be a msi based installer instead of an exe based one.

AdmiringWorm:
other notable changes:
- Icon have been updated to the new icon used by texmaker

That's about it I guess.

fixes #793


AdmiringWorm:
Seems like the appveyor build failed (did I forgot to update the uninstall script completely perhaps 😕).
Need to look into that.

AdmiringWorm:
@majkinetor any chance you could have a look at this PR?
Now that the appveyor checks are successful.

automatic/texmaker/tools/chocolateyInstall.ps1
```diff
@@ -2,12 +2,16 @@
...
+
 $packageArgs = @{
-  packageName    = 'texmaker'
-  fileType       = 'exe'
+  packageName    = $env:ChocolateyPackageName
+  fileType       = 'msi'
   file           = "$toolsPath\texmakerwin32_install.exe"
   softwareName   = 'Texmaker'
-  silentArgs     = '/S'
+  silentArgs     = "/qn /norestart /l*v `"$($env:TEMP)\$($env:chocolateyPackageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
```
majkinetor: Perhaps better to use `$(Get-PackageCacheLocation)\MsiInstall.log`.

majkinetor:
Actually, noticed few things:

1. Currently core extension is not used at all, so it should be removed from deps if it remains so
2. Last [Remove-Item removes   exe, not an msi](https://github.com/chocolatey/chocolatey-coreteampackages/blob/fc7e378fe2fb55d56be7493fba9613a920ca6c34/automatic/texmaker/tools/chocolateyInstall.ps1#L20)

AdmiringWorm:
yeah, the core extension was meant to be removed, just completely forgot about it.

> Last Remove-Item removes exe, not an msi.

Oops, how did I mess that up.

PR #800, opened 2017-07-24
(ffmpeg) added README.md; remove older install dirs

### LINKED ISSUE #799: (ffmpeg) Use a static install location
REQUEST: Can this package be changed to always install the latest version of ffmpeg to the same directory please? As it's currently packaged it always installs to programdata\chocolatey\lib\ffmpeg\tools\ffmpeg-VERSION-winBITS-static. Every time a new release comes out it makes a new directory under the package's toolsdir and extracts the files there. This causes hell with other programs (pyTiVo and streambaby are prime examples) that rely on ffmpeg via hard coded paths that keep failing after an upgrade. Of course, there's the issue that old versions never get deleted and just keep accumulating and taking up unnecessary disk space. If ffmpeg just got extracted to the toolsdir without creating a subdirectory, new versions would overwrite old versions and all of this would be solved. ALTERNATIVE#1 the location could be added to the path and updated with each version. Alternative#2 there's always the bin/tools directory (with no version specific subfolder please!). Would you please consider installing to a consistent generic folder?
 

Moving the files from the extracted dir to the toolsdir in the installation script would be perfectly acceptable. Or making a ffmpeg dir inside of toolsdir and then moving them from the extracted path into the generic ffmpeg directory would work as well. 
Any reliably consistent path will do.

majkinetor:
closes #799

### DIFF
--- a/automatic/ffmpeg/ffmpeg.nuspec
+++ b/automatic/ffmpeg/ffmpeg.nuspec
@@ -31,7 +31,7 @@ Security is a high priority and code review is always done with security in mind
     <projectSourceUrl>https://git.ffmpeg.org/ffmpeg.git</projectSourceUrl>
     <docsUrl>https://ffmpeg.org/ffmpeg.html</docsUrl>
     <dependencies>
-      <dependency id="chocolatey" version="0.10.7" />     
+      <dependency id="chocolatey" version="0.10.5" />     
     </dependencies>
   </metadata>
   <files>
--- a/automatic/ffmpeg/ffmpeg.nuspec
+++ b/automatic/ffmpeg/ffmpeg.nuspec
@@ -30,6 +30,9 @@ Security is a high priority and code review is always done with security in mind
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/ffmpeg</packageSourceUrl>
     <projectSourceUrl>https://git.ffmpeg.org/ffmpeg.git</projectSourceUrl>
     <docsUrl>https://ffmpeg.org/ffmpeg.html</docsUrl>
+    <dependencies>
+      <dependency id="chocolatey" version="0.10.7" />     
+    </dependencies>
   </metadata>
   <files>
     <file src="tools\**" target="tools" />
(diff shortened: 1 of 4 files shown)


automatic/ffmpeg/ffmpeg.nuspec
```diff
@@ -30,6 +30,9 @@ Security is a high priority and code review is always done with security in mind
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/ffmpeg</packageSourceUrl>
     <projectSourceUrl>https://git.ffmpeg.org/ffmpeg.git</projectSourceUrl>
     <docsUrl>https://ffmpeg.org/ffmpeg.html</docsUrl>
+    <dependencies>
+      <dependency id="chocolatey" version="0.10.7" />
```
AdmiringWorm: Shouldn't the version be set to `0.10.5`?
From my understanding, dependency on chocolatey should be the minimum version is will work with (I'm saying 0.10.5 because 0.10.4 was broken).

automatic/ffmpeg/tools/chocolateyInstall.ps1
```diff
@@ -1,14 +1,16 @@
...
+    FileFullPath   = gi $toolsPath\*_x32.zip
+    FileFullPath64 = gi $toolsPath\*_x64.zip    
+    Destination    = $toolsPath
 }
+
+ls $toolsPath\* | ? { $_.PSISContainer } | rm -Recurse -Force #remove older package dirs
 Get-ChocolateyUnzip @packageArgs
 rm $toolsPath\*.zip -ea 0
+
+mv $toolsPath\ffmpeg-* $toolsPath\ffmpeg
```
AdmiringWorm: I think this could cause problem with users that have the ffmpeg package already installed, but meh guess we'll handle that then.

majkinetor: No it wont, because one of the previous lines removes all previous folders.

AdmiringWorm: Oh right, missed that.

bcurran3:
That was fast. Hot damn. Thanks guys.

majkinetor:
Note that this will be available with next official ffmpeg update. We didn't push now.

PR #802, opened 2017-07-24
(seafile-client) change URL from bintray to seadrive.org

octomike:
## Description

The bintray.com URL never worked for me.. don't why that was or is. Also I'm not sure how or if `update.ps1` would modify `tools/chocolateyInstall.ps1`. 
I checked the hash, it's still the same file - you never know.

## Motivation and Context

The package was broken.

## How Has this Been Tested?

`choco install seafile-client -s .`

### DIFF
--- a/automatic/seafile-client/tools/chocolateyInstall.ps1
+++ b/automatic/seafile-client/tools/chocolateyInstall.ps1
@@ -7,7 +7,7 @@ $packageArgs = @{
 
   checksum     = 'e76626742272ac04060d002b7f83f5f25d2f704723b6192751e6ac104140b5e1'
   checksumType = 'sha256'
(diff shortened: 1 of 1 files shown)


AdmiringWorm:
unfortunately the `update.ps1` file will overwrite the url specified in `chocolateyInstall.ps1`.
However, from the looks of it, it may be that the only thing we need to do is force an update of the `seafile-client` package.

I have a few things I need to take care of first, but I'll have a look into that shortly.

AdmiringWorm:
a updated package is in the process of getting pushed to chocolatey now, the next version will be an embedded package.
So no need to download anything during install anymore (except for the package itself of course).

/cc @octomike

PR #815, opened 2017-08-10
(visualstudiocode) Install 32-bit version on 64-bit systems if already a 32-bit version is installed

### LINKED ISSUE #813: (visualstudiocode) Upgrade from 32-bit to 64-bit
## Expected Behavior
Install 32-Bit version if already a 32-Bit version is installed.

## Current Behavior
Starting with 1.15.0 Visual Studio Code supports 64-Bit version. If already a 32-Bit version is installed upgrade fails though and asks to first uninstall the 32-Bit version.

## Possible Solution
Similar implementation to Firefox package: https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/firefox/tools/chocolateyInstall.ps1#L12

## Your Environment
* Package Version used: 1.15.0
* Operating System and version: Windows 10 64-Bit
* Chocolatey version: 0.10.7

pascalberger:
## Description
Install 32-bit version on 64-bit systems if already a 32-bit version is installed

## Motivation and Context
Fixes #813 

## How Has this Been Tested?
Tested updating from 1.14.2 to 1.15 on a 64-bit system and clean installation of 1.15 on a 64-bit system

## Types of changes

## Checklist:

### DIFF
--- a/automatic/visualstudiocode/Tools/ChocolateyInstall.ps1
+++ b/automatic/visualstudiocode/Tools/ChocolateyInstall.ps1
@@ -11,7 +11,6 @@ Write-Host "Merge Tasks: "
 Write-Host "$mergeTasks"
 
 $packageName = 'visualstudiocode'
-$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
 $url32 = 'https://az764295.vo.msecnd.net/stable/8b95971d8cccd3afd86b35d4a0e098c189294ff2/VSCodeSetup-ia32-1.15.0.exe'
 $url64 = 'https://az764295.vo.msecnd.net/stable/8b95971d8cccd3afd86b35d4a0e098c189294ff2/VSCodeSetup-x64-1.15.0.exe'
 $checksum32 = '9ee938c5f695fd54a06f9a48122d409f5a12d2039f63741c82d58314f1da89d6'
@@ -19,7 +18,6 @@ $checksum64 = '0c77cdb1c472a4dcbedc39d704ad79248997a66450083f3a1e48dd070b37f439'
 
 $packageArgs = @{
   packageName = $packageName
-  unzipLocation = $toolsDir
   fileType = 'EXE'
   url = $url32
   url64bit = $url64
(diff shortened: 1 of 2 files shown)


majkinetor:
Isn't this overkill now that choco remembers install parameters ?
Using [something like this](https://github.com/majkinetor/au-packages/blob/master/smplayer/tools/chocolateyInstall.ps1#L4) perhaps is way easier. The only way to install x32 version on x64 system is to force it.

pascalberger:
The issue is that pre 1.15 32-bit version was installed on 64-bit systems.
If you now update it will fail and ask the user to uninstall 32-bit version
first. This PR makes sure that upgrading from pre 1.15 to 1.15 is possible
on 64-bit systems, by staying on the 32-bit version in these cases.

Am 11.08.2017 10:55 vorm. schrieb "Miodrag Milić" <notifications@github.com
>:

Isn't this overkill now that choco remembers install parameters ?
Using something like this
<https://github.com/majkinetor/au-packages/blob/master/smplayer/tools/chocolateyInstall.ps1#L4>
perhaps is way easier. The only way to install x32 version on x64 system is
to force it.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/chocolatey/chocolatey-coreteampackages/pull/815#issuecomment-321761931>,
or mute the thread

majkinetor:
> If you now update it will fail and ask the user to uninstall 32-bit version first. 

Perhaps that is a better option here ? Why would anybody keep using x32 bit version on x64 bit Windows to only limit itself ? With this update the fact that there is x64 bit version could go unnoticed (I only glanced  what failed for example).

pascalberger:
@majkinetor Also OK to keep the current behavior and close this PR if there's consense that a failing upgrade is the best option.

AdmiringWorm:
I don't particularly have any opinion on the best option here.
But just wanted to say one thing, the code added in this PR would have been best to have been added before 64bit releases had been added, to keep upgrades as smooth as possible.
But now, with a release already released on choco that is already lost unfortunately.

> @majkinetor said...
> Isn't this overkill now that choco remembers install parameters ?

We shouldn't really rely completely on this, remember that feature is still in preview (I think), and it only remembers the last call to `install/upgrade` (again, I think).

majkinetor:
I think its the best for users to manually decide in this case especially because extensions do not need to be reinstalled. I just did so for my own case and without `cup` problem it would probably go unnoticed.

ferventcoder:
> With this update the fact that there is x64 bit version could go unnoticed (I only glanced what failed for example).

@majkinetor you will notice, vscode tells you every time you open it.

PR #832, opened 2017-08-29
(tor-browser) Set NTFS modify file permissions

### LINKED ISSUE #831: (tor-browser) Permission issues when installed to default location (lib\tor-browser\tools\tor-browser)
Post install of tor-browser, when trying to run it gives the error:
(title) Tor Browser profile Problem
(text) Tor browser does not have permission to access the profile. Please adjust your file system permissions and try again.

## Expected Behavior
It should have permissions to run for at least the user account that installed the package.

## Current Behavior
Package installer account does not have permission to run the program and files need to manually have permissions changed to allow running of the program.

## Possible Solution
add this to end of install script:

```powershell
# set NTFS modify file permissions to $toolsDir\tor-browser\ for user account that installed the package
$WhoAmI=whoami
$Acl = Get-Acl "$toolsDir\tor-browser"
$Ar = New-Object  system.security.accesscontrol.filesystemaccessrule($WhoAmI,"Modify","Allow")
$Acl.SetAccessRule($Ar)
Set-Acl "$toolsDir\tor-browser" $Acl
```

(I forked it but not sure how to submit/push it back)!

bcurran3:
Post install of tor-browser, when trying to run it gives the error:
(title) Tor Browser profile Problem
(text) Tor browser does not have permission to access the profile. Please adjust your file system permissions and try again.

Expected Behavior

It should have permissions to run for at least the user account that installed the package.

Current Behavior

Package installer account does not have permission to run the program and files need to manually have permissions changed to

Possible Solution

add this to end of install script:

set NTFS modify file permissions to $toolsDir\tor-browser\ for user account that installed the package

$WhoAmI=whoami
$Acl = Get-Acl "$toolsDir\tor-browser"
$Ar = New-Object system.security.accesscontrol.filesystemaccessrule($WhoAmI,"Modify","Allow")
$Acl.SetAccessRule($Ar)
Set-Acl "$toolsDir\tor-browser" $Acl

(I forked it but not sure how to submit/push it back)!

Fixes #831

### DIFF
--- a/automatic/tor-browser/tools/chocolateyInstall.ps1
+++ b/automatic/tor-browser/tools/chocolateyInstall.ps1
@@ -35,3 +35,10 @@ Install-ChocolateyShortcut `
   -ShortcutFilePath "$desktop\Tor Browser.lnk" `
   -TargetPath "$toolsDir\tor-browser\Browser\firefox.exe" `
   -WorkingDirectory "$toolsDir\tor-browser\Browser"
+
+# set NTFS modify file permissions to $toolsDir\tor-browser\ for user account that installed the package
+$WhoAmI=whoami
+$Acl = Get-Acl "$toolsDir\tor-browser"
+$Ar = New-Object  system.security.accesscontrol.filesystemaccessrule($WhoAmI,"Modify","Allow")
+$Acl.SetAccessRule($Ar)
+Set-Acl "$toolsDir\tor-browser" $Acl


automatic/tor-browser/tools/chocolateyInstall.ps1
```diff
@@ -35,3 +35,10 @@ Install-ChocolateyShortcut `
   -ShortcutFilePath "$desktop\Tor Browser.lnk" `
   -TargetPath "$toolsDir\tor-browser\Browser\firefox.exe" `
   -WorkingDirectory "$toolsDir\tor-browser\Browser"
+
+# set NTFS modify file permissions to $toolsDir\tor-browser\ for user account that installed the package
+$WhoAmI=whoami
```
AdmiringWorm: while this will work for the current user, what about any other users on the computer?

bcurran3: Can be changed easily to allow all users. I go the restrictive route and only give permission to the user who had the permission to put it there. That's just my choice. You're free to do otherwise of course!

majkinetor: One solution would be to add permission to everybody. This can probably be solved better way tho, by directing the tor to use. 

See also https://tor.stackexchange.com/questions/7318/where-is-the-profile where they mention each user should have its own profile now. I don't know why it doesn't happen but perhaps it could be made forced to do so via command line.

bcurran3: That link is OSX related and talking about Mac user profiles, not Tor profiles.
If I found the right place... looks like it's a feature request for Tor to support multiple profiles but is still open:
https://trac.torproject.org/projects/tor/ticket/14981

It's simply a file permission issue. Debatable who should be allowed access to it. When I package portable programs, I don't know who should have access to them since ProgramData is restricted by default. I decided to go the safe way in these situations and ONLY give access to the user account who had the permissions to install the package. That person could then decide to change the file permissions for everyone to have access or not. This could be put in as a package note. Better to give at least the package installer rights then no one but admins. Best case scenario would be to add a package parameter at install to give the option to set allow permissions for everyone and then default to a more restrictive installer account only.

majkinetor: Yeah, its bad design choice on many programs as they should write in rm `$env:APPDATA` by default. 

I think package parameter might be worth considering or we can just leave it this way.

bcurran3: I shouldn't really defend the developers, but keep in mind that it's a portable program not a program with an installer. Most users will download, extract, move to desktop, and use with no problems. Being portable, cross-platform, and "self contained," it makes sense the way it is.

> I think package parameter might be worth considering or we can just leave it this way.

I'm don't understand the statement. Leave it as the package currently exists where people can not run it without modifying permissions? Or add my suggestions so that the installing user can run the program? 

**I feel strongly that the person who installed the package should be able to easily run it.**

majkinetor: Leave it this way = as with your current mods.

bcurran3: I'm 💯  behind that. 👍

PR #842, opened 2017-09-19
(gpg4win*) migrated from dtgm; modernized; embedded

majkinetor:
## Description

Migrated from [dtgm](https://github.com/dtgm/chocolatey-packages)

## Motivation and Context

* Embedded
* Modernized

## How Has this Been Tested?
Local tests

## Types of changes

- [x] Migrated package (a package have been migrated from another repository)

## Checklist:

- [x] I have updated the documentation accordingly (this mean usually the notes in the description of a package).

## Original Location
- [Original Repository](https://github.com/dtgm/chocolatey-packages)
- No Open Issues
- [x] [PR](https://github.com/dtgm/chocolatey-packages/pull/322)
- [x] The [migration guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Package-migration-process) have been followed


majkinetor:
As a note, previous migrations @ dtgtm are [still not merged](https://github.com/dtgm/chocolatey-packages/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aopen%20coreteam) so I don't think situation will change for this one. 

Don't know why it doesn't install on AV, look like it popups something and installer blocks.

majkinetor:
Yeah, I didn't think about icons, just took what was there from dtgm and the millisecond thought was that there is convention about naming the icon the same as the package and that probably your script wont auto update it otherwise ...

AdmiringWorm:
@majkinetor I couldn't find any reason why it won't install on appveyor (other than perhaps a service is disabled, or a different program is conflicting with it, so I'm satisfied to ignore the build error for now).

Anyways, I also update the iconurl to work with only a single icon, instead of the duplicate ones as well as updated the license url to point to the gpg4win website one.

AdmiringWorm:
I've added the chocolatey user to the packages, and considering the unlikeliness of the PR's will get merged anytime soon, I'm thinking that we can just merge this PR whenever you feel it's ready (we should also merge with a forced update for all the packages IMO)

AdmiringWorm:
Missed this

> there is convention about naming the icon the same as the package and that probably your script wont auto update it otherwise ...

While we do have a convention for naming the icon, I think in certain cases we really should ignore that convention so we don't end up with a bunch of duplicate icons.

Also, my script will pick up the icon **IF** the correct name is used in the iconUrl for the package.
See https://github.com/chocolatey/chocolatey-coreteampackages/pull/842/commits/1a9678bba1b3a32faa2463d724cd093e7877022d for what was needed to change to make it work as expected.

The script uses the following matching to find an icon:
1. Use the file name of the existing icon url (excluding extension)
2. Use the name of the package
3. Use the name of the package without one of the following suffixes (`.install`, `.portable`, `.commandline`)

majkinetor:
The whole point for me is not to think about icon URL. Can I just put something like:

```xml
<iconUrl> gpg4win.png</iconUrl>
```

and be done wiht it ? 

Thx for the fixes you made.

AdmiringWorm:
@majkinetor if you prefix it with a `/`, then yes you can.
To grab the file name, the script first looks for that character, it's not necessary to fill in the iconUrl when the icon matches the package name though (or maybe you need to have something there, but doesn't matter what).

Issue #849, opened 2017-09-23
(putty.install) Iinstalls but Choco shows error code 1603


pauby:
## Expected Behavior

Package should install without any error being shown.

## Current Behavior

Package shows this error on install:

ERROR: Running ["C:\WINDOWS\System32\msiexec.exe" /i "C:\ProgramData\chocolatey\lib\putty.install\tools\putty-64bit-0.70
-installer.msi" /quiet] was not successful. Exit code was '1603'. See log for possible error messages.
The upgrade of putty.install was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\putty.install\tools\chocolateyInstall.ps1'.
 See log for details.

## Possible Solution

Add 1603 to valid exit codes.

## Steps to Reproduce (for bugs)

Install the package.

pauby:
PR closed so closing issue.

AdmiringWorm:
Reopened as it deserves to be looked into why the installer returns the error code.

pauby:
Thsi was on a new laptop I was setting up so when I get some time I will pull it out and get the Choco logs from it.

AdmiringWorm:
Adding my latest comment from the PR #850 so we don't have to jump back and forth:

> so I will have to dig it back out and go through the logs. Will also run the install again with the switches you asked for and post them here.

Unfortunately I don't believe the chocolatey logs will be useful in this case, but the logs from the msi installer would be useful (msi installers probably logs to a file by default, but I have no idea where 😢)

EDIT:
I plan to go through all of the packages here in this repository to enable logging to a specific file by default (including this one) (same with inno setup installers as well).

pauby:
> I plan to go through all of the packages here in this repository to enable logging to a specific file by default (including this one) (same with inno setup installers as well).

@AdmiringWorm That is a very good idea. And it's something I will look at doing with my own packages. Logging the install to the chocolatey package folder.

I have so far tried 5 times to force the putty.install package install, uninstall and also with rebooting in between and sods law it installs every time without error.

AdmiringWorm:
@pauby I was rather thinking of logging it to the package cache directory, similiar to how the silent args are used in the default template for msi installs.
```
silentArgs    = "/qn /norestart /l*v `"$($env:TEMP)\$($packageName).$($env:chocolateyPackageVersion).MsiInstall.log`""
```
[...]

PR #850, opened 2017-09-23
Add exit code 1603

pauby:
## Description
Added valid exit code 1603 as per #849 - 1603 is a common exit code used by the MSI installer but it still installs the package.

## Motivation and Context
Resolves issue #849 

## How Has this Been Tested?
It's not been tested in this installer but has in others.

## Types of changes

## Checklist:

- [ ] All files is up to date with the latest [Contributing Guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)

### DIFF
--- a/automatic/putty.install/tools/chocolateyInstall.ps1
+++ b/automatic/putty.install/tools/chocolateyInstall.ps1
@@ -9,7 +9,7 @@ $packageArgs = @{
     File           = "$toolsPath\putty-0.70-installer.msi"
     File64         = "$toolsPath\putty-64bit-0.70-installer.msi"
     SilentArgs     = "/quiet"
-    ValidExitCodes = @(0)
+    ValidExitCodes = @(0,1603)
 }
 Install-ChocolateyInstallPackage @packageArgs
 


AdmiringWorm:
I'm a little hesitant to accept a PR that adds a common error code as a valid exit code, especially considering I've yet been able to reproduce any instance where I actually get an error code of 1603.

pauby:
It's a common error code for MSI packages - another generic Microsoft error code that generally means little. If the package didn't install I'd be looking to see if there is an issue but whenever I get these error codes (and I get them on different machines but the package always installs)

If you're unhappy with the PR I'll close it.

AdmiringWorm:
@pauby it isn't that I'm unhappy with the PR or anything like that, but rather that there generally is a reason for the 1603 error code (no matter how unhelpful that error code is).
Most likely IMO, there is either some kind of conflict when running the installer that prevents all of the tasks to be completed successfully (even though the program gets installed), or a missing requirement or something along those lines.

Could you try installing the package with logging to file enabled?
`choco install putty.install --installargs="/l*v C:\putty.MsiInstall.log"` (Just change the path if you like).

Then have a look through the log file to see if there is any mention on why it fails.
In the meantime I'll open this PR again.

pascalberger:
I also don't think 1603 should be handled as successful, since it does have different reasons (see [here](https://support.microsoft.com/en-us/help/834484/you-receive-an-error-1603-a-fatal-error-occurred-during-installation-e)) and handling it as successful will result in failing installations not being recognized as this.

pauby:
@AdmiringWorm I updated the issue with a comment but just to reiterate it here too - this was on a new laptop install (fresh Windows 10 Pro build with all patches) so I will have to dig it back out and go through the logs. Will also run the install again with the switches you asked for and post them here.

@pascalberger I understand where both of you are coming from on this one. I've yet to get that error and the installation actually fail which was my initial motivation for raising this PR. There are a whole raft of posts on the error and how unhelpful it is. When I had this problem the very first time that Microsoft article was the first one that came up in the search and it's suggestions didn't apply.

AdmiringWorm:
> so I will have to dig it back out and go through the logs. Will also run the install again with the switches you asked for and post them here.

Unfortunately I don't believe the chocolatey logs will be useful in this case, but the logs from the msi installer would be useful (msi installers probably logs to a file by default, but I have no idea where 😢)

PR #858, opened 2017-10-01
(bulk-crap-uninstaller) embedded and modernized; updated readme

### LINKED ISSUE #857: (bullk-crap-uninstaller) Modernize and fix links
Should be embedded. Changelog and license links do not work

majkinetor:
Closes #857


automatic/bulk-crap-uninstaller/tools/chocolateyInstall.ps1
```diff
@@ -1,16 +1,14 @@
...
-  checksum64             = '589737f7b5e70558c8e9a13aeab0e654a8c73c4d35ec1d620371665e89b6433d'
-  checksumType           = 'sha256'
-  checksumType64         = 'sha256'
-  silentArgs             = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART'
-  validExitCodes         = @(0)
-  softwareName           = 'Bulk Crap Uninstaller'
+  packageName    = 'bulk-crap-uninstaller'
+  fileType       = 'exe'
+  file           = gi $toolsPath\*.exe
+  file64         = gi $toolsPath\*.exe
```
AdmiringWorm: Please remove the file64 property, it isn't needed.

automatic/winscp.install/winscp.install.nuspec
```diff
@@ -2,15 +2,15 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
```
AdmiringWorm: this you really intend to change the nuspec file for this package?
Please revert.

automatic/bulk-crap-uninstaller/tools/chocolateyInstall.ps1
majkinetor: Why not ?

AdmiringWorm: Because there is only 1 installer

majkinetor: We have that on number of packages. The point here is that correct installer message will be printed since installers contain both x32 and x64 versions.

majkinetor: Talking about  `"Installing 64-bit bulk-crap-uninstaller..."`

AdmiringWorm: So I take it you feel it's wrong when it just outputs `"Installing bulk-crap-uninstaller..."`?

The architecture is only outputted when both `file` and `file64` is specified, otherwise they are left out completely (same thing with `url` and `url64`)

AdmiringWorm: It also kinda conflicts with [1.1.6](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md#116-support-multiple-architectures) (although that specifically mentions urls, but it should be extended to use embedded files as well)

majkinetor: OK

PR #859, opened 2017-10-01
(opera) removed x32 bit check

### LINKED ISSUE #789: (opera) Upgrading 64-bit Opera actually forces 32-bit
## Expected Behavior
Upgrading 64-bit Opera upgrades to the new version of 64-bit Opera and other packages are not forced to 32-bit.

## Current Behavior
When upgrading 64-bit Opera, the [`Is32BitInstalled`](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/opera/tools/helpers.ps1#L6) check is triggering so the 32-bit Opera is downloaded and installed instead and also the [`ChocolateyForceX86`](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/opera/tools/chocolateyInstall.ps1#L5) environment variable is set so subsequent unrelated package upgrades in a `choco upgrade all` are forced to 32-bit as well. I'm guessing that at some point the 64-bit Opera installer changed to no longer explicitly put its uninstall key into the 64-bit registry and is now putting it into the 32-bit registry instead.

1. On a 64-bit version of Windows: `choco install opera -y --version=46.0.2597.39` will download and install the previous 64-bit version of Opera.
2. Then running `choco upgrade all -y` will download and install the latest 32-bit version. If another package needs to be upgraded after Opera, it will also be forced to 32-bit.

## Context
Just trying to upgrade packages.

## Your Environment
* Package Version used: 46.0.2597.46
* Operating System and version: Windows 10 Version 1703 (OS Build 15063.413)
* Chocolatey version: 0.10.7

majkinetor:
closes #789

### DIFF
--- a/automatic/opera/tools/chocolateyInstall.ps1
+++ b/automatic/opera/tools/chocolateyInstall.ps1
@@ -2,9 +2,6 @@
 $toolsPath = (Split-Path -Parent $MyInvocation.MyCommand.Definition)
 . "$toolsPath\helpers.ps1"
 
-if (Is32BitInstalled) { $env:ChocolateyForceX86 = $true }
-
-$version     = '48.0.2685.32'
 $packageArgs = @{
   packageName    = $env:ChocolateyPackageName
   fileType       = 'exe'
@@ -18,7 +15,7 @@ $packageArgs = @{
   validExitCodes = @(0)
 }
 
-if (IsVersionAlreadyInstalled $version) {
+if (!$Env:ChocolateyForce -and (IsVersionAlreadyInstalled $Env:ChocolateyPackageVersion)) {
   Write-Output "Opera $version is already installed. Skipping download and installation."
 } else {
   Install-ChocolateyPackage @packageArgs
--- a/automatic/opera/tools/chocolateyInstall.ps1
+++ b/automatic/opera/tools/chocolateyInstall.ps1
@@ -15,7 +15,8 @@ $packageArgs = @{
   validExitCodes = @(0)
 }
 
-if (!$Env:ChocolateyForce -and (IsVersionAlreadyInstalled $Env:ChocolateyPackageVersion)) {
+$version = ''
+if (!$Env:ChocolateyForce -and (IsVersionAlreadyInstalled $version)) {
   Write-Output "Opera $version is already installed. Skipping download and installation."
 } else {
   Install-ChocolateyPackage @packageArgs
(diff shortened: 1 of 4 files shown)


automatic/opera/tools/chocolateyInstall.ps1
```diff
@@ -18,7 +15,7 @@ $packageArgs = @{
   validExitCodes = @(0)
 }
 
-if (IsVersionAlreadyInstalled $version) {
+if (!$Env:ChocolateyForce -and (IsVersionAlreadyInstalled $Env:ChocolateyPackageVersion)) {
```
AdmiringWorm: using `$env:ChocolateyPackageVersion` won't work in the long run, the package version may be different than the software version.

automatic/opera/tools/helpers.ps1
```diff
@@ -1,13 +1,3 @@
-function Is32BitInstalled {
```
AdmiringWorm: Is there no longer any way to detect if the 32bit version is installed?

automatic/opera/tools/chocolateyInstall.ps1
AdmiringWorm: In cases where we need to provide a package fix.

automatic/opera/tools/helpers.ps1
majkinetor: There is no need to do so. `force86` is remembered now so it will be reused in subsequent install.  Also, this is not specific to this package.

AdmiringWorm: I was under the impression that it only remembered the previous install/update.
Maybe that was fixed though without me noticing.

majkinetor: :point_up: [September 21, 2017 8:53 PM](https://gitter.im/chocolatey/chocolatey-coreteampackages?at=59c40a8ebac826f054e5902f)

AdmiringWorm: Oh right, completely forgot about that one.
Ignore me on this one then...

automatic/opera/tools/chocolateyInstall.ps1
majkinetor: OK, returned

AdmiringWorm: thanks

majkinetor:
I will not push fix, Opera updates frequently.

AdmiringWorm:
yeah, usually it does.
Sometimes it's only an update for one of the architectures though, unfortunately we ignore those right now.

Issue #873, opened 2017-10-13
(openssl.light) installation error "Cannot bind argument to parameter 'Path' because it is null"


donestefan:
## Expected Behavior
`choco install openssl.light` should install without error message

Please note that chocolatey has been installed in non-default folder location. This might be be the cause of the trouble. Other packages work without problems.
```
C:\Windows\system32>echo %ChocolateyInstall%
D:\PortableApps\chocolatey
```
## Current Behavior
"choco install openssl.light" shows error `ERROR: Cannot bind argument to parameter 'Path' because it is null.`
Package will not be installed. Package files will end up in `D:\PortableApps\chocolatey\lib-bad`

relevant Log entries show:
```
2017-10-13 15:44:29,697 5164 [DEBUG] - Running Get-ProcessorBits -compare '64' 
2017-10-13 15:44:29,698 5164 [INFO ] - VERBOSE: Trying Program Files with 2 levels depth: C:\Program Files C:\Program Files\*\* C:\Program Files (x86) C:\Program Files (x86)\*\*
2017-10-13 15:44:30,078 5164 [INFO ] - VERBOSE: Trying native commands on PATH
2017-10-13 15:44:31,037 5164 [INFO ] - VERBOSE: Trying Registry: \SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths
2017-10-13 15:44:31,076 5164 [INFO ] - VERBOSE: No location found
2017-10-13 15:44:31,092 5164 [ERROR] - ERROR: Cannot bind argument to parameter 'Path' because it is null.
2017-10-13 15:44:31,114 5164 [DEBUG] - Built-in PowerShell host called with ['[System.Threading.Thread]::CurrentThread.CurrentCulture = '';[System.Threading.Thread]::CurrentThread.CurrentUICulture = ''; & import-module -name 'D:\PortableApps\chocolatey\helpers\chocolateyInstaller.psm1'; & 'D:\PortableApps\chocolatey\helpers\chocolateyScriptRunner.ps1' -packageScript 'D:\PortableApps\chocolatey\lib\OpenSSL.Light\tools\chocolateyInstall.ps1' -installArguments '' -packageParameters '''] exited with '-1'.
2017-10-13 15:44:31,120 5164 [DEBUG] - Calling command ['"C:\Windows\System32\shutdown.exe" /a']
[...]

PR #874, opened 2017-10-13
(openssl) GH #873

majkinetor:
Some refactoring and fix for the issue

### DIFF
--- a/automatic/openssl.light/tools/chocolateyInstall.ps1
+++ b/automatic/openssl.light/tools/chocolateyInstall.ps1
@@ -1,13 +1,10 @@
-﻿$ErrorActionPreference = 'Stop';
+﻿$ErrorActionPreference = 'Stop'
 
-$toolsPath = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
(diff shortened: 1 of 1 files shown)


automatic/openssl.light/tools/chocolateyInstall.ps1
```diff
@@ -20,15 +17,10 @@ $packageArgs = @{
...
-$binPath = "$($installLocation)\bin" 
-
-if(Test-Path $installLocation ) {
-  Install-ChocolateyPath -PathToInstall $binPath -PathType Machine 
-}
-else {
-  Write-Warning "Could not add install directory to path"
-}
+if (!$installLocation)  { Write-Warning "Can't find install location, PATH not updated"; return }
+Write-Host "Installed to '$installLocation'"
```
AdmiringWorm: shouldn't this be added in an `else` statement?
What is the point of outputting the install location if it isn't found?

majkinetor: Nah, it exists the script in if part with `return` since there is nothing to be done

AdmiringWorm: ah right, sorry missed the return statement

Issue #736, opened 2017-05-25
(chromium) follow chrome 'stable' releases, use WebRTC/sync


AeliusSaionji:
Hello!

The current usage of the `chromium` package is not what one would expect when installing chromium.

It currently follows "last known good build" and also lacks WebRTC (H.264), widevine (Netflix), and syncing functionality.

This is clearly a package meant for minimal testing- not a fully featured browser one would expect to get.

We need to switch over to a build that includes
- WebRTC
- Sync
- Widevine (unless that can be installed as a separate package)

Furthermore, chromium releases should also be taking advantage of chocolatey's prelease/beta functionality. This package should be kept in-line with the current release of `chrome`. All these nightly ("last known good build") builds should not be released as new package versions- they should be released as `${version}-beta`.

We could optionally provide builds without webrtc, sync, and widevine via packageargs.

I might take a stab at implementing this sometime soon

AdmiringWorm:
> This is clearly a package meant for minimal testing- not a fully featured browser one would expect to get.

While chromium should be a usable browsers, it is less featured than other browsers that are based on chromium (Say: Google Chrome, Opera, Vivaldi, etc...).

> It currently follows "last known good build"

Unfortunately, that is the only ones we can get using the API we currently use (from https://chromium.woolyss.com/, I'm considering a change to https://omahaproxy.appspot.com/ though which then would allow us to provide only stable releases)

> We need to switch over to a build that includes
>
> - WebRTC
> - Sync
> - Widevine (unless that can be installed as a separate package)

Do you know of any (provided by google?), the one we currently use is the one provided by google (I think)

> Furthermore, chromium releases should also be taking advantage of chocolatey's prelease/beta functionality. This package should be kept in-line with the current release of chrome. All these nightly ("last known good build") builds should not be released as new package versions- they should be released as ${version}-beta.

Using beta for this package would be wrong, using `-snapshot` would be more correct, or `-alpha` or something along those lines.

Link-Satonaka:
> While chromium should be a usable browsers, it is less featured than other browsers that are based on chromium (Say: Google Chrome, Opera, Vivaldi, etc...).

My understanding is that chromium is only the parts of chrome that are open source. Chromium under Linux has all of the components I've mentioned (widevine provided as a separate installable package). Under Linux, chromium is as fully capable as chrome- it's just that Google isn't building it that way for Windows.

[...]

PR #875, opened 2017-10-14
(chromium) Rework updater with two major changes:

AeliusSaionji:
1. Use builds from https://github.com/henrypp/chromium which offers
codecs, chrome-sync, and webrtc. With this, chromium is as capable as chrome

2. Specifically follow the "release" versions, mirroring google chrome's
current release version. As chrome is built from chromium, and they use
the same versioning schemes, it follows that the stable version of
chrome is the stable version of chromium. chromium has no official
builds other than a nightly build with reduced feature set.

https://github.com/chocolatey/chocolatey-coreteampackages/issues/736


automatic/chromium/update.ps1
```diff
@@ -19,37 +15,22 @@ function global:au_SearchReplace {
...
-    $hashtype32 = @{$true = $hashtype; $false = $chromium32.hashtype }[ $chromium32.hashtype -ne ""]
-    $hashtype64 = @{$true = $hashtype; $false = $chromium64.hashtype }[ $chromium64.hashtype -ne ""]
-
-    $version = $chromium64.version
-
-    $url32 = 'https://storage.googleapis.com/chromium-browser-snapshots/Win/<revision>/mini_installer.exe'
-    $url64 = 'https://storage.googleapis.com/chromium-browser-snapshots/Win_x64/<revision>/mini_installer.exe'
-    $url32 = $url32 -replace '<revision>', $chromium32.revision
-    $url64 = $url64 -replace '<revision>', $chromium64.revision
+function global:au_BeforeUpdate {
```
majkinetor: Isn't more simply `Get-RemoteFiles -Purge -NoSuffix` working ?

majkinetor: See https://github.com/majkinetor/au#embedding-binaries

AeliusSaionji:
I haven't embedded the installer- is that the desirable default these days?

AdmiringWorm:
@AeliusSaionji embedding packages is a general recommendation to do on chocolatey.org (and a required on in this repository) whenever it is possible.

We will need to look into doing exactly that, and finding the reason why the uninstall script fails (before we can merge this PR).

RedBaron2:
Well I found the uninstall error. It has to do with a few things, and to keep this short. I did a shorthand Chocolatey package bad 😈 . 
@AeliusSaionji compounded the error by removing 
`"(^\s*packageName\s*=\s*)('.*')"= "'$1'$($Latest.PackageName)'"`
This caused Chocolatey to install the package without a packageName. So when the uninstall script came to my shorthand of `$registry = Get-UninstallRegistryKey -SoftwareName $packageName`
you end up with the error of a failed uninstall. It is an easy fix. just replace `$packageName` with `"chromium*"`
This causes everything to work right at least on my systems.
Hope this helps 😄

Akipe:
Hello!
@AeliusSaionji I think it is missing checking if the build is stable, some are dev. It is display below the title of each releases here https://github.com/henrypp/chromium/releases

Look this comment for more information https://github.com/chocolatey/chocolatey-coreteampackages/issues/736#issuecomment-336615375

I created script few days ago to retrieve the latest stable version, you can use it if needed (it surely needs to be improved) : https://github.com/Akipe/chocolatey-packages/tree/master/automatic/chromium-stable

AeliusSaionji:
@Akipe thanks! At the time I wrote my script, the developer had been relyably using the "Latest Release" and "Pre-Release" tags. That no longer seems to be the case, and your solution definitely works better.

@RedBaron2 thanks for that. Yeah, I don't understand why so many AU packages want to update the package name. Package names are static, they shouldn't be updated, should they?

AdmiringWorm:
@AeliusSaionji thank you for updating, unfortunately it seems there is a conflict with the master branch.
Could you please rebase towards the upstream master branch?

Then we could get this change merged.

automatic/chromium/tools/chocolateyInstall.ps1
```diff
@@ -12,18 +12,17 @@ if (Test-Path $Chromium) {
...
   silentArgs    = $silentArgs
   validExitCodes= @(0)
   softwareName  = 'Chromium'
-  checksum      = 'a2df30145c98391f61e0d181252a5ba8'
-  checksumType  = 'md5'
-  checksum64    = '9bc789b8477ef3f5a38ca088ba69db9f'
-  checksumType64= 'md5'
 }
 
 Install-ChocolateyPackage @packageArgs
```
AdmiringWorm: I missed this initially, you can't use `Install-ChocolateyPackage` when embedding a package (although it sometimes work), should use instead `Install-ChocolateyInstallPackage`.

I can make that change when I merge it in though.

AdmiringWorm:
@majkinetor I made some changes to this PR, mind having a look?

automatic/chromium/chromium.nuspec
```diff
@@ -15,7 +15,7 @@
 
 ## Notes
 
-- This package installs by default to the `%programfiles(x86)%` instead of `%localappdata%\Chromium`.
+- This package installs the Chromium snapshots builds by default to the `%programfiles(x86)%` instead of `%localappdata%\Chromium`.
```
AeliusSaionji: Looks like you accidentally this little tweak to the nuspec

AeliusSaionji:
Thanks for the fixes + help!

automatic/chromium/chromium.nuspec
AdmiringWorm: ops, I'll go through my commits and figure out where I made that mistake

AdmiringWorm: Ah, I see what happened.
I always test the updating process before I push out a new commit, and since the `Readme.md` hadn't been updated (still contains the old message), it got replaced in the nuspec file.

AdmiringWorm:
@AeliusSaionji thank you for all your work on this package,
@Akipe and also thank you for the script to parse out the stable version.

It will take some time until the chromium package have cached up to the latest snapshot build though, so in the meantime it would probably be best to use @Akipe's package until it does.

BTW @Akipe, when this package have catched up with the snapshot builds, would you mind deprecating your `chromium-stable` package and have the normal `chromium` as a dependency then?
Would be nice not to have anymore duplicated packages than what is necessary.

Akipe:
@AdmiringWorm of course ! I will notify users in the comments and warn them of the future change.

PR #879, opened 2017-10-20
(bluegriffon) Changed to embedded package + added innosetup logging

AdmiringWorm:


automatic/bluegriffon/tools/chocolateyInstall.ps1
```diff
@@ -1,16 +1,16 @@
...
+  file           = "$toolsPath\"
+  softwareName   = 'bluegriffon*'
+  silentArgs     = "/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP- /LOG=`"$($env:TEMP)\$($packageName).$($env:chocolateyPackageVersion).InnoSetup.log`""
   validExitCodes = @(0)
 }
 
-Install-ChocolateyPackage @packageArgs
+Install-ChocolateyInstallPackage @packageArgs
+
+Remove-Item -Force -ea 0 "$toolsPath\*.exe","$toolsPath\*.ignore"
```
majkinetor:
```
ls $toolsPath\*.exe | % { rm $_ -ea 0; if (Test-Path $_) { sc "$_.ignore" } }
```

automatic/bluegriffon/tools/chocolateyInstall.ps1
```diff
@@ -1,16 +1,16 @@
 ﻿$ErrorActionPreference = 'Stop'
 
+$toolsPath = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
```
majkinetor: Always wondered why people put this in expression when this is perfectly correct:

```
$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
```

AdmiringWorm: Honestly, I don't know.
I got so used to seeing it that way that it became a habit (and also because I have it that way in my migration script).

Doesn't really matter though IMO.

automatic/bluegriffon/tools/chocolateyInstall.ps1
AdmiringWorm: I believe that the script doesn't continue after uninstall until after the installer have completely finished, which would make that redundant.

But I could change it anyhow.

automatic/bluegriffon/tools/chocolateyInstall.ps1
majkinetor: It matters due to '2.2 Code style'.

AdmiringWorm: There isn't anything that conflicts with [2.2 Code style](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md#22-code-style), while it isn't needed and may not correspond to your own coding style, there is still nothing wrong with having it this way.

And just in case you haven't noticed, that line is also the same as the one you get when you run `choco new`

majkinetor: More complicated code is less readable then less complicated code. 
Choco internals has nothing to do with standards on this repository.

But whatever really. :)

ferventcoder: Something to do with PowerShell v2 IIRC.

PR #887, opened 2017-10-29
Testing out streams

AdmiringWorm:
Not completely sure if this is how the streams are meant to work, but opening this PR to find out.


majkinetor:
That looks good IMO. This is the scenario with streams automatically generated which is very nice and IMO not that weird.

AdmiringWorm:
I guess we can give this one a try in the repo itself then.
Cross fingers that everything will work as expected 🤞

AdmiringWorm:
@majkinetor yeah I guess, when thinking of streams I was actually thinking more of dynamically created functions/scriptblocks that only ran when needed

PR #897, opened 2017-11-08
Python

majkinetor:
Embedding and streams


majkinetor:
What remains to be done is to test some or all of the older packages as only 3.6 is tested on AV. I don't have vm installed ATM. I think that silent args are the same for both msi and exe but that might be wrong.

automatic/python3/python3.nuspec
```diff
@@ -31,6 +31,9 @@ For example: `--params '"/InstallDir:C:\tools\python2"'`
     <licenseUrl>https://www.python.org/download/releases/3.4.0/license</licenseUrl>
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/python3</packageSourceUrl>
     <projectSourceUrl>https://www.python.org/downloads/source</projectSourceUrl>
+    <dependencies>
+      <dependency id="chocolatey" version="0.10.8" />
```
AdmiringWorm: why a dependency on chocolatey, instead of chocolatey-core.extension?

automatic/python3/python3.json
```diff
@@ -1,4 +0,0 @@
-﻿{
```
AdmiringWorm: should probably not remove the json file, otherwise it won't be commited when the package is updated.
Correct me if I'm wrong though.

majkinetor: I thought it would... not sure. I guess that we can find out this way :)

automatic/python3/python3.nuspec
majkinetor: Because of Get-PackageParameters

AdmiringWorm: which is available in the extension, which would be better to have as a dependency instead of dictating which choco versions users must use IMO.

automatic/python3/python3.json
AdmiringWorm: Unless you have changed it since I last had a look at the `git` AU plugin, you use the argument `-u` which have the following description
> -u, --update          update tracked files

This also **only** updates tracked files, and nothing else.

majkinetor:
I tested it locally and all packages seem to work fine.

Issue #901, opened 2017-11-11
(nomacs) installer is changed upstream


majkinetor:
Nomacs no longer works as the exe file is removed and new MSI installer is created along with some other changes.

majkinetor:
#901

majkinetor:
It makes sense to embed the package now as vendor doesn't keep older executables.

PR #902, opened 2017-11-11
(nomacs) fixed updater and msi installer

majkinetor:
#901


majkinetor:
Can't see how this conflict happened. Otherwise this is done

automatic/nomacs/nomacs.nuspec
```diff
@@ -3,63 +3,58 @@
...
-  - rotate images
-  - drag and drop images
-  - automatically save screenshots
-  - delete/rename images
-  - create mosaic image
-  - set wallpaper (windows only)
+- crop images
+- resize images
+- pseudo color function
+- color adjustments (brightness, contrast, saturation â€¦)
```
AdmiringWorm: Seems like you have somehow gotten an invalid character here somehow. Would need to be fixed

RedBaron2:
I kind of remember there being now x86 & x64 versions for this software. Also the developer listing `/passive` as the unattended switch on the FAQ page.

AdmiringWorm:
@RedBaron2 if you look through the scripts, you'll notice both x86 and x64 is being aquired for the installer (the portable one only supports x64).

The installer is a normal msi, those support both the `/passive` and the `/quiet` switches.

RedBaron2:
OK mobile view of site was limiting viewing. Carry on.

PR #918, opened 2017-11-19
(opera) don't create desktop shortcut or pin to taskbar

cpmcgrath:
https://chocolatey.org/packages/Opera

## Description
Added two command line arguments to the installer, one to stop it from creating the desktop shortcut, the other for pinning to the taskbar.

## Motivation and Context
By default, the opera installer pins itself to the Windows Taskbar. This goes against the principles of how the taskbar should be used - Programs shouldn't pin themselves to the taskbar, Users should pin Programs to the Taskbar.

A Desktop shortcut is common practice, however for it to be re-added when updating the program is overriding the user's preference - If it's not there, it's because I deleted it, so don't readd it. Not creating the desktop shortcut on initial install is a small price to pay (it's preferable to me, but that's just personal opinion)

## How Has this Been Tested?
I downloaded the 64bit installer and ran it with the silentargs.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/opera/tools/chocolateyInstall.ps1
+++ b/automatic/opera/tools/chocolateyInstall.ps1
@@ -11,7 +11,7 @@ $packageArgs = @{
   checksum64     = '44dcef541df90b3665a88091c504e56a3594a630534ea8d0f1f76f443dd6fcba'
   checksumType   = 'sha256'
   checksumType64 = 'sha256'
-  silentArgs     = '/install /silent /launchopera 0 /setdefaultbrowser 0'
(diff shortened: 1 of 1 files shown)


AdmiringWorm:
Thank you for your contribution, but could I ask you to change it a little.
Make these changes as an optional parameter instead of having them hard coded into the silent arguments.

It's easy enough to do with the use of the `chocolatey-core.extension`'s `Get-PackageParameters` helper (just search for it in this repo to see how it is used).

As such it would also need to be documented under a `Package Parameters` sub header `## Package Parameters`

gep13:
@cpmcgrath have to agree with @AdmiringWorm here.  This would be better suited as a Package Parameter, rather than directly into the installation arguments.

majkinetor:
Closing this due to no feedback.

PR #939, opened 2017-12-23
(Chromium) add the snapshots package back & add check for the stable builds

RedBaron2:
@gep13 @AdmiringWorm @majkinetor @pascalberger 
Snapshots and Stable build enhancements

## Description
**Describe your changes in detail**
_This PR will add the snapshots build back to the chocolatey community feed. It also makes the snapshots build a embedded package like the stable build. 
This PR will also add a check into the stable build that will prevent the error that occurs when the stable 32-bit build is late. The check will prevent the error from occurring_ 

## Motivation and Context
**Why is this change required? What problem does it solve?**
_The Community Feed of Chromium Snapshots was removed, and this will bring it back. It also fixes a error that occurred in the current stable build_

_This also introduces a version check that will fail gracefully if the current version of the package is already installed on the system. This will also prevent the possible shim being created for the installation package for both snapshots and stable builds._ 

## How Has this Been Tested?
**Please describe in detail how you tested your changes.**
_Update versions have been tested locally and on AppVeyor
Installs have been done locally through install and daily `cup` of package for the past few weeks._

## Types of changes

## Checklist:

- [X] All files is up to date with the latest [Contributing Guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)

## Original Location
  - [ ] Issue 1 link
  - [ ] Issue 2 Link


RedBaron2:
I knew it was good. Sometimes doing the right thing is not always the easy thing 😄

automatic/chromium/tools/chocolateyInstall.ps1
```diff
@@ -1,26 +1,30 @@
...
+
+if (Test-Path $Chromium) {
+  $silentArgs = '--do-not-launch-chrome'
+} else {
+  $silentArgs = '--system-level --do-not-launch-chrome'
+}
+
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$version = "63.0.3239.108"
+Get-CompareVersion -version $version -notation "-snapshots" -package "chromium"
```
AdmiringWorm: What is the purpose of this.
As far as I can see, the only purpose is to write out a message, but nothing else.
Did you mean to have this one behave a little differently?
Like skipping installation if it's already installed?

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+        "(?i)(\s*32\-Bit Software.*)\<.*\>" = "`${1}<$($Latest.URL32)>"
+        "(?i)(\s*64\-Bit Software.*)\<.*\>" = "`${1}<$($Latest.URL64)>"
+        "(?i)(^\s*checksum\s*type\:).*"     = "`${1} $($Latest.ChecksumType32)"
+        "(?i)(^\s*checksum32\:).*"          = "`${1} $($Latest.Checksum32)"
+        "(?i)(^\s*checksum64\:).*"          = "`${1} $($Latest.Checksum64)"
+        }
+    ".\tools\chocolateyInstall.ps1" = @{
+        '(^[$]version\s*=\s*)(".*")' = "`$1""$($Latest.Version)"""
+        '(?i)(^\s*file\s*=\s*)(".*")' = "`$1""$($Latest.PackageName32)"""
+        '(?i)(^\s*file64\s*=\s*)(".*")' = "`$1""$($Latest.PackageName64)"""
```
AdmiringWorm: Why are you using `PackageName32` and `Packagename64`?
The helper you are using in `global:au_BeforeUpdate` automatically sets `FileName32` and `FileName64` which you should use instead.

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+}
+
+function global:au_BeforeUpdate {
+  # Copy the original file from the chromium folder
+    Get-RemoteFiles -Purge -FileNameBase "$($Latest.PackageName)"
+}
+
+    $releases_stable = 'https://api.github.com/repos/henrypp/chromium/releases'
+    $releases_snapshots = 'https://chromium.woolyss.com/api/v2/?os=windows&bit=<bit>&out=string'
+    $PackageName = 'chromium'
```
AdmiringWorm: Don't set the `PackageName` when it isn't needed.

`PackageName` is automatically set by AU

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+    $ChecksumType = 'sha256'
+    $PackageName32 = ("`$toolsdir\"+($PackageName)+"_x32.exe");
+    $PackageName64 = ("`$toolsdir\"+($PackageName)+"_x64.exe");
```
AdmiringWorm: These two arguments won't be available to the `global:au_SearchReplace`.
You need to set them to `$Latest.PackageName32` and `$Latest.PackageName64` if you truly need them.

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+    $url32 = $Pre_Url+"Win/<revision>/mini_installer.exe"
+    $url64 = $Pre_Url+"Win_x64/<revision>/mini_installer.exe"
+    $url32 = $url32 -replace '<revision>', $chromium32.revision
+    $url64 = $url64 -replace '<revision>', $chromium64.revision
+    
+    @{
+        URL32 = $url32
+        URL64 = $url64
+        Version = $version
+        PackageName = $PackageName
```
AdmiringWorm: Don't set the package name wehn you don't need to.

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+        Title = $Title
+        PackageName32 = $PackageName32
+        PackageName64 = $PackageName64
```
AdmiringWorm: Using variables before them are available is always a bad idea.

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+  $32LatestSyncInstallUrl = ($32LatestVersion.assets | Where-Object name -match "-sync.exe").browser_download_url
+
+  $64LatestVersion = $anyArchLatestStablesVersions | Where-Object {$_.tag_name -match $latestStableVersionNumber -and $_.tag_name -match "win64"}
+  $64LatestSyncInstallUrl = ($64LatestVersion.assets | Where-Object name -match "-sync.exe").browser_download_url
+
+  @{
+    Version = $latestStableVersionNumber
+    PackageName = $PackageName
+    Title = $Title
+	PackageName32 = $PackageName32
```
AdmiringWorm: Using variables before they are available is always a bad idea

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+
+  $32LatestVersion = $anyArchLatestStablesVersions | Where-Object {$_.tag_name -match $latestStableVersionNumber -and $_.tag_name -match "win32"}
+  $32LatestSyncInstallUrl = ($32LatestVersion.assets | Where-Object name -match "-sync.exe").browser_download_url
+
+  $64LatestVersion = $anyArchLatestStablesVersions | Where-Object {$_.tag_name -match $latestStableVersionNumber -and $_.tag_name -match "win64"}
+  $64LatestSyncInstallUrl = ($64LatestVersion.assets | Where-Object name -match "-sync.exe").browser_download_url
+
+  @{
+    Version = $latestStableVersionNumber
+    PackageName = $PackageName
```
AdmiringWorm: Don't set the package name when not needed (they are automatically set by AU)

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,114 @@
...
+    Title = $Title
+	PackageName32 = $PackageName32
+    URL32 = $32LatestSyncInstallUrl
+    ChecksumType32 = $ChecksumType
+	PackageName64 = $PackageName64
```
AdmiringWorm: Using variables before they are available is always a bad idea

automatic/chromium/tools/chocolateyInstall.ps1
RedBaron2: Yes, that is the intention. I will change it to set a bool for the install as a catch.

RedBaron2:
@AdmiringWorm 
The requested changes have been made. I hope that everything is good now 🤞

automatic/chromium/update.ps1
```diff
@@ -1,47 +1,106 @@
...
+    ".\chromium.nuspec" = @{
+    "(?i)(^\s*\<title\>).*(\<\/title\>)" = "`${1}$($Latest.Title)`${2}"
+    }
+  }
+}
+
+function global:au_BeforeUpdate {
+    Get-RemoteFiles -Purge -FileNameBase "$($Latest.PackageName)"
+    $Latest.FileName32 = ("`$toolsdir\"+$($Latest.PackageName)+"_x32.exe");
+    $Latest.FileName64 = ("`$toolsdir\"+$($Latest.PackageName)+"_x64.exe");
```
AdmiringWorm: I think I'll be changing this up before I merge this PR.
Rather than updating the FileName32 like this, I would rather have it similiar to how the following package uses regex replacement instead:
https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/texmaker/update.ps1#L19

But anyways, I can take care of that, but won't happen until later this weekend.

RedBaron2: I had not thought about doing it that way. I can make the changes in a while.

AdmiringWorm: If you could, that would be great.
Do note though, it's not wrong with doing it this way though, I just feel the other way is a little better (probably isn't though 😄)

RedBaron2: I'll work on it. It is a elegant way of doing it. Just requires work.

RedBaron2:
@AdmiringWorm 
I have made those changes to the update.ps1 file. I tested it locally. Enjoy 😄

AdmiringWorm:
@RedBaron2 great, just had a look at it, and it's exactly as I planned it myself.
I'll merge it now (which is the last thing I'll do now before I head to bed).

RedBaron2:
@AdmiringWorm Thanks 👍

PR #951, opened 2018-01-19
Add icons for 8 packages

abejenaru:
Packages are the following:
etcd, Kitematic, Kubernetes, Minikube, OpenShift Origin, Microsoft R Open, nteract and Pencil

Packages will have their PRs submitted shortly once the icon commit reference is available.

## Original Location
- [Original Repository] https://github.com/abejenaru/chocolatey-packages/tree/master/icons


majkinetor:
We can't accept icons like this. Please commit icons along with the package.

[3.4 Pull request one package](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md#34-pull-request-one-package)

AdmiringWorm:
@abejenaru btw, you don't need to have the correct url to the iconUrl when submitting the package, any valid url can be used in the iconUrl.
We have a script running that updates the iconUrl to the one that is located in the icons folder (based on a naming convention, either named the same as the package, or the same as the package exluding `.install`, `.portable`, `.commandline` etc...)

abejenaru:
@AdmiringWorm indeed, that was the problem I was trying to avoid by committing the icons first. Thanks!

PR #996, opened 2018-03-27
(inkscape) removes msi before upgrade to prevent 1638 error, fixes #550

### LINKED ISSUE #550: (inkscape) Due to the bug in vendor installer, package should uninstall prior to update
Upgrading the inkscape package is currently impossible due to a bug in the installer making it return error code 1638 if a previous version of the software is already installed.

See:
http://disq.us/p/1f03dr0
https://bugs.launchpad.net/inkscape/+bug/1653574

jtcmedia:
## Description
Added a beforemodify script that checks to see if Inkscape is installed and uninstalls the msi.

## Motivation and Context
fixes #550

## How Has this Been Tested?
Installed 0.92.2 w/ beforemodify script, then did an upgrade to 0.92.3. Tested OK.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/inkscape/tools/chocolateybeforemodify.ps1
+++ b/automatic/inkscape/tools/chocolateybeforemodify.ps1
@@ -1,0 +1,7 @@
+﻿$app = Get-WmiObject -Class Win32_Product -Filter "Name Like 'Inkscape%'"
+
+if ($app) {
+  Write-Host "Inkscape already installed. Removing..."
+  $app.Uninstall()
+  Write-Host "Done. Continuing with update/uninstall..."
+}


automatic/inkscape/tools/chocolateybeforemodify.ps1
```diff
@@ -0,0 +1,7 @@
+﻿$app = Get-WmiObject -Class Win32_Product -Filter "Name Like 'Inkscape%'"
```
AdmiringWorm: Please don't use the `Get-WmiObject`, but rather use a combination of [Get-UninstallRegistryKey](https://chocolatey.org/docs/helpers-get-uninstall-registry-key) and [Uninstall-ChocolateyPackage](https://chocolatey.org/docs/helpers-uninstall-chocolatey-package) instead.

If you could update with that, I would be more than happy to merge this one in.

AdmiringWorm: also forgot, this must be done in the install script, not the before modify script (*Since the before modify script is run on both upgrade and uninstall, but doesn't fail the package if it doesn't work)*)

AdmiringWorm:
@HolisticDeveloper while indeed it is still subject to the potential slow downs that was mentioned, that wasn't the only reason I requested it to be changed.

Whenever the `Win32_Product` is used, it forces Windows Installer to reconfigure every application that have been installed through MSI (this is also the cause of the potention slow down though).

This is also why it's *suggested* to avoid when possible.
https://github.com/chocolatey/package-validator/wiki/DontUseGetWmiObjectForFindingInstalledPackages

automatic/inkscape/tools/chocolateybeforemodify.ps1
jtcmedia: @AdmiringWorm I tried that before but ran into problems. I've spent enough time on this so I encourage someone else to have a go at this.

AdmiringWorm:
Thank you for your contribution, unfortunately as it currently stands we can't use it as is.
However, the issue you wanted to fix with this PR have been fixed with an alternate edition and will be available when the next version of InkScape is released.

PR #1008, opened 2018-04-24
(sourcetree) Use MSI installer

### LINKED ISSUE #1001: (sourcetree) Use MSI installer
## Expected Behavior
The Sourcetree package currently uses a Squirrel based installer which, installes into user directory. This doesn't match well with a machine wide packaging tool like Chocolatey. Starting with 2.5.0 a MSI installer is provided.

## Current Behavior
Package installes into user directory

## Possible Solution
Package should use MSI installer. MSI installer should be [publicly available after full rollout](https://groups.google.com/d/msg/atlassian-sourcetree-beta/zbXM0MN8Fp8/QqlQzAw4BwAJ).

MSI installer will disable auto-update and and remove embedded Git/Hg.

The packe should provide an upgrade path if possible from 2.4 to 2.5 by uninstalling current Sourcetree version and migrating settings (e.g. repos) to new installation

pascalberger:
## Description
Use new enterprise version of Sourcetree which comes as an MSI

## Motivation and Context
The current Squirrel based installer is not really suitable for a deployment with Chocolatey as it installs in the user directory. Using the MSI fixes these issues.

Fixes #1001

## How Has this Been Tested?
Tested locally

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


pascalberger:
Some questions by myself:

The MSI installs side by side with the Squirrel based installer used in current versions of this package. Since using the Squirrel based installer no longer makes sense I don't think the MSI deserves it's own package, but instead the existing package should be updated. But currently there is no proper upgrade path from previous versions to this version. Should we try to uninstall existing Sourcetree installations in this package? What if a user really want to have both installed side by side?

Settings also want be migrated from existing installations. Should we try to do this migration or is this out of scope for a Chocolatey package?

I also should update the description in the nuspec to reflect the above points based on the direction we want to take.

AdmiringWorm:
@pascalberger could you rebase towards the latest of master? Currently there are merge conflicts.

I'll review the PR after that have been done.

automatic/sourcetree/tools/chocolateyInstall.ps1
```diff
@@ -1,17 +1,17 @@
...
+  validExitCodes= @(0,1641,3010)
+  url           = 'https://downloads.atlassian.com/software/sourcetree/windows/ga/SourcetreeEnterpriseSetup_2.5.5.msi'
+  checksum      = '2fb9c8676259f8d74acd50e6069bc93c2ca7c3e9305341fa6e0571df0d7d74e9'
+  checksumType  = 'sha256'
+  url64bit      = ''
+  checksum64    = ''
+  checksumType64= 'sha256'
 }
 
 Install-ChocolateyPackage @packageArgs
```
AdmiringWorm: what about updates, is the new MSI installer subjected to the same kind of error during update as quite a few MSIs have?

pascalberger:
@AdmiringWorm @gep13 I've now added a check if there's already an existing installation of Sourcetree installed using the Squirrel installer on the machine. If so it asks the user if it should be uninstalled. Settings will **not** be migrated to the new installation.

This is ready for a review from my side

pascalberger:
> You can detect previous existing package and throw with adequate message for a user. User should devise migration strategy if thats needed.

@majkinetor Currently the package checks for existing Squirrel installation and asks user if he wants to uninstall it. Does this fit with what you asked? Or do you prefer to only print out a note that another SourceTree installation is present and new package will be installed side-by-side?

pascalberger:
@majkinetor I removed the uninstall code and just output a warning that standard version of Sourcetree won't be uninstalled. Also added an additional note to the package description

majkinetor:
Is there anything else you need to do here ?
It would be good to have some proof that this works before merging. Could you `Test-Package` locally or inside VM and give output here (or something equivalent).

majkinetor:
Nvm merging this, I thought AV buid failed for different reasons.

majkinetor:
Hm... I see now that you didn't throw here. Please throw on older version found. Nobody will see warnings once they do `cup all` but they will see errors as they are listed at the end of all updates. Nobody expects this - installing the same named software and getting something parallel to existing one.

We have similar situation with VSCode x64 install, and package didn't hendled it, but during installl VSCode would complain, which was singnal to all users that intervention is required.

pascalberger:
@majkinetor I don't think throwing will lead to the best user experience. This forces the user to manually uninstall his prior Sourcetree installation before he can install the new version. By this he loses all settings. By only warning he at least can check the settings in the old version and apply them in the new version before uninstalling the old version.

majkinetor:
Well... ok...

PR #1025, opened 2018-05-05
(vagrant) add vagrant plugin repair

### LINKED ISSUE #1024: (vagrant) Vagrant update may break Vagrant plugins
Vagrant plugins are sometimes broken after an update of Vagrant.

## Expected Behavior

Vagrant plugins work after an update of Vagrant.

## Current Behavior

Vagrant plugins need a manual run of "vagrant plugin repair" to work properly after an update of Vagrant.

## Possible Solution

Maybe it is a solution to add "vagrant plugin repair" to the update process for the Vagrant package to fix the broken plugins during the update automatically.

1. Install an older version of vagrant: choco install vagrant --version 2.0.1 
2. Install some plugins: vagrant plugin install vagrant-vbguest
3. Update vagrant: choco update vagrant
4. Try to start a Vagrantbox which uses the installed plug-ins.
5. You will see something like the following output:

```bash
Vagrant failed to initialize at a very early stage:

The plugins failed to initialize correctly. This may be due to manual
modifications made within the Vagrant home directory. Vagrant can
attempt to automatically correct this issue by running:

  vagrant plugin repair

If Vagrant was recently updated, this error may be due to incompatible
versions of dependencies. To fix this problem please remove and re-install
all plugins. Vagrant can attempt to do this automatically by running:

  vagrant plugin expunge --reinstall

Or you may want to try updating the installed plugins to their latest
versions:

  vagrant plugin update

Error message given during initialization: Unable to resolve dependency: user requested 'vagrant-hostmanager (> 0)'
```

## Context

I saw this issue sometimes in the last weeks and just want to make it visible for other users. I think the error message of vagrant will help every user to fix this issue by himself, but it would be a nice improvement to add a function to the update process which fixes the problem automatically.

## Your Environment

* Package Version used: 2.0.1 and 2.1.0
* Operating System and version: Windows 10 Pro
* Chocolatey version: 0.10.11
* Install/uninstall gist: -

majkinetor:
closes #1024

### DIFF
--- a/automatic/vagrant/tools/chocolateyinstall.ps1
+++ b/automatic/vagrant/tools/chocolateyinstall.ps1
@@ -14,3 +14,6 @@ $packageArgs = @{
   softwareName   = 'vagrant'
 }
 Install-ChocolateyPackage @packageArgs
+
+Update-SessionEnvironment
+vagrant plugin repair  #https://github.com/chocolatey/chocolatey-coreteampackages/issues/1024


AdmiringWorm:
hmm, I see the failure from appveyor shows a drawback from the testing script.
Guess I'll need to look into grabbing valid exit codes.

Issue #866, opened 2017-10-03
Packages that should get embedded


majkinetor:
[1.1.4 - Embed a package if allowed](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md#114-embed-a-package-if-allowed) means that there are number of packages here that could be enhanced in this manner.

## Packages that should get embedded

- [ ] SparkleShare (GNU)
- [x] Python2 (GPL)
---

- [x] bluegriffon (Mozilla Public License)
- [x] brackets (MIT)
- [x] freevideoeditor
  > You may install and use the free version of the Product, released under the brand VSDC Free Video Editor, free of charge for personal and educational use. In this case, you are granted the right to use and to make an unlimited number of copies of this software.
- [x] gimp (GPLv3)
- [x] cdburnerxp
  > You may also make copies and distribute the software according to requirements of section 2 below.
- [x] cdrtfe (GNU)
- [x] classic-shell 
  > You are permitted to distribute the Classic Shell installer or offer it for download on your website, as long as:
- [x] clipboardfusion
  >  Free versions available for download from Binary Fortress Software's website may be freely distributed with written permission from Binary Fortress Software 
- [x] clipgrab (GNU)
- [x] displayfusion
  >  Free versions available for download from Binary Fortress Software's website may be freely distributed with written permission from Binary Fortress Software 
- [x] djv
  > Redistributions in binary form must reproduce the above copyright notice, this list of conditions, and the following disclaimer in the documentation and/or other materials provided with the distribution.
- [x] encfs4win (LGPL)
- [x] Ext2Fsd (LGPL)
- [x] hostsman
  > The Licensor grants you the right to make this freeware available on an
internet site for the purpose of downloading, as long as NO FEE IS CHARGED for
said download.
- [x] inkscape (GPL2)
- [x] jitsi (GPL2)
- [x] LibreCAD  (GPL2)
- [x] maxima (GPL)
- [x] mixx (GPL)
- [x] mkvtoolinix (
- [x] mpchc (GPL)
- [x] MuseScore (GPL2)
- [x] Mumble
  > Redistributions of source code must retain the above copyright notice,  this list of conditions and the following disclaimer. can be distributed through non-commercial channels,
- [x] nircmd ([personal communication](https://github.com/majkinetor/au-packages/blob/master/nirlauncher/legal/redistribution.pdf))
- [x] notepadplusplus (GPL)
[...]

PR #1039, opened 2018-05-21
(keeweb) Migrate package

wget:
Just port my KeeWeb package to coreteampackage.

## How Has this Been Tested?
* Windows 7 SP1
* AppVeyor WMF 5 image. However, as per Rob's request we removed the forcing of the TLS version in our chocolateyInstall.ps1. This change might make the build fail (the test of the package install process).

N/A

## Types of changes

## Checklist:

## Original Location
- [Original Repository](https://github.com/wget/chocolatey-package-keeweb)
- [x] The [migration guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Package-migration-process) have been followed


automatic/keeweb/keeweb.nuspec
```diff
@@ -0,0 +1,31 @@
...
+    <metadata>
+        <id>keeweb</id>
+        <title>KeeWeb</title>
+        <version>1.6.2</version>
+        <authors>antelle</authors>
+        <owners>wget</owners>
+        <summary>Free cross-platform password manager compatible with KeePass</summary>
+        <description>This webapp is a browser and desktop password manager compatible with KeePass databases. It doesn't require any server or additional resources. The app can run either in browser, or as a desktop app.</description>
+        <projectUrl>https://keeweb.info/</projectUrl>
+        <tags>keeweb keypass password admin</tags>
```
majkinetor: missing madatory tags: `foss` `cross-platform`

wget: @majkinetor You mean the final nupkg package should have both (32 and 64bits) installers bundled? That's 40Mio *2 for both :-/

majkinetor: Yes, because it makes package stable and the one you can depend on. 
There are number of other packages like that here, they are marked with special icon in report. There is also an [open issue](https://github.com/chocolatey/chocolatey-coreteampackages/issues/866) about this.

wget: Ok I updated the package according to the changes you requested.

wget:
Ok I updated the package according to the changes you requested.

majkinetor:
It requires some reafactoring

- Move legal stuff to legal folder, add  this folder in nuspec
- Your choco installer doesn't need to decided on x64/x32. Depend on latest choco and use File/File64 parameters (and remove readme note about this since its basic choco behavior).

After that this is read 4 merge.

Also,  your `get_latest` can be significantly reduced; you don't need to set checksums when using `Get-RemtoeFiles`, it will do that for you. This is not the same since comparing to what you did now - hash is calculated from the file wile you take from the site. I do not requrie you to change this but if this was not your original intention you could go this route too.

wget:
@majkinetor `Your choco installer doesn't need to decide on x64/x32 manually` this is what does QBittorrent thought:

https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/qbittorrent/tools/chocolateyinstall.ps1#L6

majkinetor:
That package was created back in time when choco didn't support this.
Examples: 

https://github.com/chocolatey/chocolatey-coreteampackages/blob/0259c93ffc63f34cc48861bbbfa6e4d04bba70f1/automatic/codelite/tools/chocolateyInstall.ps1

https://github.com/chocolatey/chocolatey-coreteampackages/blob/0b755454aab4e6bb5591a502fe76de7972f5159f/automatic/nomacs/tools/chocolateyInstall.ps1

majkinetor:
> And yes, wrt. the hashes, I'm aware Get-RemoteFiles already does such a job, but indeed, like you understood, I'm relying on upstream's hashes rather than the ones generated automatically :)

FYI, AU can still check it out for you:

>_From: AU README.md_
>
> If the ChecksumXX hash key is present, the AU will change to checksum verification mode - it will download the installer and verify that its checksum matches the one provided. If the key is not present, the AU will calculate hash with the given ChecksumTypeXX algorithm.

wget:
> If the ChecksumXX hash key is present, the AU will change to checksum verification mode - it will download the installer and verify that its checksum matches the one provided. If the key is not present, the AU will calculate hash with the given ChecksumTypeXX algorithm.

@majkinetor So do I need to change anything then?

automatic/keeweb/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,18 @@
...
+  file64         = "$toolsPath\KeeWeb-1.6.2.win.x64.exe"
+  softwareName   = 'keeweb*'
+  silentArgs     = '/S'
+  validExitCodes = @(0)
+}
+
+Install-ChocolateyInstallPackage @packageArgs
+
+# Lets remove the installer as there is no more need for it
+Remove-Item -Force -ea 0 "$toolsDir\*.exe","$toolsDir\*.exe.ignore"
```
majkinetor: This can fail. This is what we use, and you can remove ignore files

`ls $toolsPath\*.exe | % { rm $_ -ea 0; if (Test-Path $_) { sc "$_.ignore" '' } }`

majkinetor:
This looks OK (just change that one line).

Thanks for your patience. We do this kind of thing on purpose, so your next package wont require so much assistence.

majkinetor:
> @majkinetor So do I need to change anything then?

In theory, this can make invalid package as checksum can change during download etc. But probability is very low and perhaps this is how it should be.

Looking forward to MM package.

wget:
@majkinetor I'm already set as maintainer of KeeWeb and chocolatey has already been set as maintainer as well :)

And btw, I already have write commit rights to this core team packages repo :) But, right now, haveing a second admin helping me in reviewing things is great. Thanks!

PR #1053, opened 2018-06-06
(opera) Parameters for disabling desktop and taskbar shortcuts

jack1142:
## Description
I added support for two Opera installer arguments - one stops creating desktop shortcut for Opera and the other stops pinning Opera to Windows taskbar.

## Motivation and Context
This change provides better expierence for users, because you can disallow installer from creating desktop shortcut for Opera or pinning it to taskbar, so user won't need to remove shortcut by himself each time.

## How Has this Been Tested?
I packed package with `choco pack` and tested the package with `choco install opera -s . --params "/NoDesktopShortcut /NoTaskbarShortcut"` and with `choco install opera -s .` (to see, if it didn't destroy standard installation) and just in case I also tested uninstalling it.

Tested on Windows 10, version 1803

## Types of changes

## Checklist:

### DIFF
--- a/automatic/opera/README.md
+++ b/automatic/opera/README.md
@@ -3,3 +3,10 @@
 
 The Opera web browser makes the Web fast and fun, giving you a better web browser experience on any computer.
 
+
+## Parameters
+- `/NoDesktopShortcut` - Do not create desktop shortcut for Opera
+- `/NoTaskbarShortcut` - Do not pin Opera to taskbar
+
+These parameters can be passed to the installer with the use of `--params`.
+For example: `--params '"/NoDesktopShortcut /NoTaskbarShortcut"'`
--- a/automatic/opera/opera.nuspec
+++ b/automatic/opera/opera.nuspec
@@ -14,6 +14,12 @@
     <description><![CDATA[
 The Opera web browser makes the Web fast and fun, giving you a better web browser experience on any computer.
 
+## Parameters
+- `/NoDesktopShortcut` - Do not create desktop shortcut for Opera
+- `/NoTaskbarShortcut` - Do not pin Opera to taskbar
+These parameters can be passed to the installer with the use of `--params`.
+For example: `--params '"/NoDesktopShortcut /NoTaskbarShortcut"'`
+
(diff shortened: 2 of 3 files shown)


jack1142:
Ah, and to add to this, in my opinion it shouldn't create desktop and taskbar shortcut defaultly, but I'm not sure, if you would agree to change it, so for now I left it as it is.

jack1142:
And one more thing, I wanted to add date to version but Opera already uses 4 part segmentation in version number :/

AdmiringWorm:
> Ah, and to add to this, in my opinion it shouldn't create desktop and taskbar shortcut defaultly, but I'm not sure, if you would agree to change it, so for now I left it as it is.

while personally I also don't like to have desktop shortcuts by default, we should honor the defaults of the installer.

> And one more thing, I wanted to add date to version but Opera already uses 4 part segmentation in version number :/

Yeah, adding a date wouldn't work here. While there are ways we could push it out, it's not yet been added to the package (and probably won't be yet either). So basically, these changes won't be available until the next version of Opera hits.

jack1142:
> Yeah, adding a date wouldn't work here. While there are ways we could push it out, it's not yet been added to the package (and probably won't be yet either). So basically, these changes won't be available until the next version of Opera hits.

Yeah I thought we will need to wait for next version, just wanted to point that out 👍

There's still some time until next release of Opera, because last one was released just 4 days ago, so there's time for revision of this PR and if anything is wrong I can still change it.

I just want to add these parameters, because with auto upgrading packages I need to remove this pin every week or two and that's a little annoying.

AdmiringWorm:
The changes so far looks good, but I need to test if they are POSHv2 compatible.
I'm a little unsure if it is, with the regard of line 7,8 and 19 in the install script.

If everything works as it should, then it would be ready to be merged in.

BTW, just a tip for the future. You don't need to update the `<description>` element in the nuspec file, it's enough to update the README (the description is automatically updated from the README.md file during update)

AdmiringWorm:
@jack1142 thank you for your contribution

Issue #1065, opened 2018-07-05
(pdfcreator) Error while running 'C:\ProgramData\chocolatey\lib\PDFCreator\tools\chocolateyInstall.ps1'.


basictheprogram:
## Expected Behavior

choco install pdfcreator would install

## Current Behavior

Installation was not successful. Exit code was '1'

```powershell
PS C:\Users\Administrator> choco install pdfcreator
Chocolatey v0.10.11
Installing the following packages:
pdfcreator
By installing you accept licenses for the packages.
Progress: Downloading PDFCreator 3.2.2... 100%

PDFCreator v3.2.2 [Approved]
pdfcreator package files install completed. Performing other installation steps.
Print Spooler service state:  / Running
Print Spooler service new state: Auto / Running
Installing PDFCreator...
ERROR: Running ["C:\ProgramData\chocolatey\lib\PDFCreator\tools\PDFCreator-3_2_2-Setup.exe" /VERYSILENT /NORESTART /COMPONENTS="program,ghostscript,comsamples,languages,languages\bosnian,languages\catalan,languages\catalan_valencia,languages\chinese_simplified,languages\chinese_traditional,languages\corsican,languages\croatian,languages\czech,languages\danish,languages\dutch,languages\english,languages\estonian,languages\finnish,languages\french,languages\galician,languages\german,languages\greek,languages\hebrew,languages\hungarian,languages\italian,languages\irish,languages\ligurian,languages\latvian,languages\lithuanian,languages\norwegian_bokmal,languages\polish,languages\portuguese_br,languages\romanian,languages\russian,languages\serbian_cyrillic,languages\slovak,languages\slovenian,languages\spanish,languages\swedish,languages\turkish,languages\valencian_avl" ] was not successful. Exit code was '1'. See log for possible error messages.
The install of pdfcreator was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\PDFCreator\tools\chocolateyInstall.ps1'.
 See log for details.
This is try 1/3. Retrying after 300 milliseconds.
 Error converted to warning:
 (5) Access is denied: [\\?\C:\ProgramData\chocolatey\lib-bad\PDFCreator]
This is try 2/3. Retrying after 400 milliseconds.
 Error converted to warning:
 (5) Access is denied: [\\?\C:\ProgramData\chocolatey\lib-bad\PDFCreator]
Maximum tries of 3 reached. Throwing error.
[...]

PR #1079, opened 2018-07-18
pdfcreator - fix non-silent installer

teknowledgist:
## Description
* Added an AHK script to walk through the installer without user intervention.  
* Added necessary lines to the chocolateyinstall.ps1 to run the AHK.
* Added an autohotkey.portable dependency in the .NUSPEC (and re-arranged the elements to match the `choco new` template order which I find more logical).

## Motivation and Context
The free, PDFCreator installer is no longer silent and now requires AutoHotKey (or similar) to install without user-intervention.  

## How Has this Been Tested?
Tested install and uninstall on Win7x64 and Win10(Ent1803)x64 and didn't notice any differences in the results from previous pdfcreator installs, *except:*

* The install still opens a URL (to push PDF Architect) on completion, but until there is a Capture-URL function to block installers that do this (something I'm working on), it will have to remain.

I have *not* tested it on the Chocolatey Test Environment because I have never had any luck getting it to work (my own ignorance I guess).  However, the changes I introduced shouldn't break what was obviously working before.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/pdfcreator/tools/chocolateyinstall.ahk
+++ b/automatic/pdfcreator/tools/chocolateyinstall.ahk
@@ -1,0 +1,68 @@
+DetectHiddenWindows, on
+__Welcome:
+; Welcome window
+WinWait, Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard
+WinHide, Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard
+; Next button
+ControlClick, &Next, Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard
+
+__Components:
+; Select Components window
+WinWait, Setup - PDFCreator, Select Components, 1
+if ErrorLevel {
+  if WinExist(Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard) {
+    goto, __Welcome
+  }
+}
+; Next button
+ControlClick, &Next, Setup - PDFCreator, Select Components
+
+__Ready:
+; Ready to Install window
+WinWait, Setup - PDFCreator, Ready to Install, 1
+if ErrorLevel {
+  if WinExist(Setup - PDFCreator, Select Components) {
+    goto, __Components
+  }
+}
+; Install button
+ControlClick, &Install, Setup - PDFCreator, Ready to Install
+
+__PDFArchitect:
+; PDF Architect (Don't want) window
+WinWait, Setup - PDFCreator, PDF Architect, 1
+if ErrorLevel {
+  if WinExist(Setup - PDFCreator, Ready to Install) {
+    goto, __Ready
+  }
+}
+; Next button
+ControlClick, No thanks, Setup - PDFCreator, PDF Architect
+
+__OptionalOffer:
+; Install Complete Optional Offer window (definitely don't want)
+WinWait, Setup - PDFCreator, Installation Completed, 30
+if ErrorLevel {
+  if WinExist(Setup - PDFCreator, PDF Architect) {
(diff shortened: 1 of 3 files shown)


AdmiringWorm:
As I mentioned in the issue related to this PR, I don't believe it would be correct of us to add autohotkey to circumvent their intention of denying domain users to install the personal/free edition of this software.

I'm willing to discuss this though:
https://github.com/chocolatey/chocolatey-coreteampackages/issues/1065#issuecomment-408955490

majkinetor:
> As I mentioned in the issue related to this PR, I don't believe it would be correct of us to add autohotkey to circumvent their intention of denying domain users to install the personal/free edition of this software.

In my opinion, 

This is not the first time we do this, and this is in domain of philosophy, not programming. Automation like this is not unlawful AFAIK because you can't say something along the lines of .. _automated clicking the operating system button is considered hack_ or whatever and if you do say that, its probably unlawful (at least in EU). IANAL, but I am pretty sure in this.

There is Chocolatey opt-out policy for vendors if they have a problem with specific packages related to their software and that is official channel for stuff like this.

I vote to merge this. Almost a month past since @AdmiringWorm invited others to comment, I suggest we wait for another week and if no additional responses merge this.

automatic/pdfcreator/tools/chocolateyinstall.ahk
```diff
@@ -0,0 +1,68 @@
+DetectHiddenWindows, on
+__Welcome:
+; Welcome window
+WinWait, Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard
```
majkinetor: What if it doesn't show for some reason (crach for example). AHK script will stay in the tray. You need timeout here.

automatic/pdfcreator/tools/chocolateyinstall.ahk
```diff
@@ -0,0 +1,68 @@
...
+WinHide, Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard
+; Next button
+ControlClick, &Next, Setup - PDFCreator, Welcome to the PDFCreator Setup Wizard
```
majkinetor: What if there is another setup window ? Not very likely but you should probably get a hwnd of the single one and work with `ahk_id %hwnd%` afterwards. Even better would be for AHK to actually start the installer and get its hwnd.

Not very high probability of errors tho.

automatic/pdfcreator/tools/chocolateyinstall.ahk
```diff
@@ -0,0 +1,68 @@
...
+
+__Completing:
+; Completing the PDFCreator Setup Wizard
+WinWait, Setup - PDFCreator, Completing the PDFCreator Setup Wizard, 1
+if ErrorLevel {
+  if WinExist(Setup - PDFCreator, Installation Completed) {
+    goto, __OptionalOffer
+  }
+}
+while WinExist(Setup - PDFCreator, Completing the PDFCreator Setup Wizard) {
```
majkinetor: This could loop forever.

automatic/pdfcreator/tools/chocolateyinstall.ahk
majkinetor: Looking more into this its not possible to do it with ahk_id since windows probably change but they all probably belong to same PID which can be used to filter out unwanted windows. 

Not terribly important tho.

majkinetor:
https://github.com/chocolatey/chocolatey-coreteampackages/issues/1065

PR #1106, opened 2018-09-06
(nuget.commandline)  Update using tools.json feed

### LINKED ISSUE #1076: (nuget.commandline) Latest version not updated due to SemVer2
## Expected Behavior

nuget.commandline should be updated to latest version.

## Current Behavior

Latest version of nuget.commandline on nuget.org is [4.7.0+9245481f357ae542f92e6bc5e504fc898cfe5fc0](https://www.nuget.org/packages/NuGet.CommandLine/4.7.0) (updated with https://github.com/NuGet/Home/issues/7114). As it uses SemVer2 metadata this is not compatible with the implementation of the update script nuget.commandline on Chocolatey is still at 4.6.2.

## Possible Solution<

Make update script compatible with SemVe2 version  numbers.

pascalberger:
## Description
Update nuget.commandline package using tools.json feed instead from nuget.commandline NuGet package.

## Motivation and Context
In the past few versions several time the nuget.commandline package was not updated after the NuGet package reached recommodation/blessing state. This PR changes update to use the JSON file describing all versions. It also updates as soon as a new released version is available even if it is not the recommended/blessed version yet (like currently 4.8.1)

Fixes #1076

## How Has this Been Tested?
Tested update locally

## Types of changes

## Checklist:

### DIFF
--- a/automatic/nuget.commandline/README.md
+++ b/automatic/nuget.commandline/README.md
@@ -2,3 +2,6 @@
 
 NuGet is the package manager for the Microsoft development platforms including .NET. NuGet gives you access to thousands of packages from other developers on [nuget.org](http://nuget.org), and the NuGet tools let you create, share, and host packages of your own.
 
+## Notes
+
+This package contains **preview** and **released** and **recommended** versions.
--- a/automatic/nuget.commandline/legal/VERIFICATION.txt
+++ b/automatic/nuget.commandline/legal/VERIFICATION.txt
@@ -3,7 +3,7 @@ Verification is intended to assist the Chocolatey moderators and community
 in verifying that this package's contents are trustworthy.
 
 The embedded software have been downloaded from the software authors
-nuget package located on <https://www.nuget.org/packages/NuGet.CommandLine/>
+website located at <https://www.nuget.org/downloads>
 and can be verified by doing the following:
 
 1. Download the following <https://api.nuget.org/packages/nuget.commandline.4.7.1.nupkg>
--- a/automatic/nuget.commandline/update.ps1
+++ b/automatic/nuget.commandline/update.ps1
@@ -1,17 +1,9 @@
 import-module au
 
-$packageName = 'NuGet.CommandLine'
+$releases = "https://dist.nuget.org/tools.json"
 
 function global:au_SearchReplace {
    @{
-        # Those might be useful for some other nuget packages
-
-        # ".\tools\chocolateyInstall.ps1" = @{
-        #     "(?i)(^\s*[$]packageName\s*=\s*)('.*')"= "`$1'$packageName'"
-        # }
-        # ".\tools\chocolateyUninstall.ps1" = @{
-        #     "(?i)(^\s*[$]packageName\s*=\s*)('.*')"= "`$1'$packageName'"
-        # }
         ".\legal\VERIFICATION.txt" = @{
           "(?i)(\s*download the.*)<.*>" = "`$1<$($Latest.URL32)>"
           "(?i)(^\s*checksum\s*type\:).*" = "`${1} $($Latest.ChecksumType32)"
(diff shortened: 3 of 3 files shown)


pascalberger:
@majkinetor Or should we better implement this with streams? One package for released (including preview versions) and the other for recommendet version?

majkinetor:
I don't think streams are needed here but prereleases. So I think we should just add `-preview` to current package version if its preview and call it a day. Streams make sense if both versions are independently updated and I don't think its the case here - it seems linear: preview, preview, preview, release, preview, preview ... If that is not the case, streams are the only option and IMO better then 2 packages as its easier to maintain.

pascalberger:
@majkinetor They have three steps: Preview -> Release -> Blessed (what they mention as recommended on the page and should be reflected in the nuget.commandline package). See https://github.com/NuGet/Home/issues/4851

If we just provide one package and people update they're on the Release track. If they only want Blessed releases they need to check the nuget.org website. I'm OK with doing it like this and avoid two separate packages with all the issues involved (like dependency issues, etc).

But they release updates to the Release or Blessed version, even if newer preview versions already exists. E.g. here (you can check release dates in the https://dist.nuget.org/tools.json file):

```
4.7.0: 2018-05-07T23:00:00.0000000Z
4.8.0-preview3: 2018-07-06T23:00:00.0000000Z
4.7.1: 2018-08-10T23:00:00.0000000Z
```

So I assume three streams for preview, release and blessed makes sense?

pascalberger:
I've updated the PR to have two streams:

- One for release and preview 
- One for blessed releases

majkinetor:
> But they release updates to the Release or Blessed version, even if newer preview versions already exists

It doesn't matter since they will alway go as `-preview` and can be recognized upstream as such. The probleem is release/blessed. I propose simple solution - add 4th number in the build if blessed (for example 100, high enough to allow for some fixes to be used on it - 666 or 999 sound interesting as another example). So you will know without looking at official site what is blessed  and you can even act on it in a script via `choco info`

majkinetor:
For example this would look like this on gallery:

4.8.0-preview3
4.7.1
4.7.0.666
4.7.0

Another idea, blessed package could be virtual package pointing to blessed release, but not DRY ....

pascalberger:
I'm not really sold on the idea of implementing a custom versioning system for handling this case. As it might confuse people more than that they will get it (and still requires custom scripting).

How about going with the current approach in this PR and pushing everything to Chocolatey, with an update to the description that package also contains non-blessed versions.

If someone asks for a package containing only blessed releases there still can be a virtual package containing only blessed release.

Btw. Another side effect of this PR is that we went (intentionally) out of sync with the nuget.commandline package. I originally implemented this package like this, with checking the json, but I think @ferventcoder prefered to have it mirroring the NuGet package.

pascalberger:
> What I meant here is not fixed words such as this package contains release and blessed versions but to actually change README.md via au_SearchReplace and insert somewhere BLESSED RELEASE as you see one.

I don't think this makes sense as this is a mutable value, while the Chocolatey package description is immutable. Like 4.8.1 is now on state `Released` but probably would become `ReleasedAndBlessed` at some point.

majkinetor:
> I don't think this makes sense as this is a mutable value, while the Chocolatey package description is immutable. Like 4.8.1 is now on state Released but probably would become ReleasedAndBlessed at some point.

If that is so, the only way to keep single package and have everything without information loss is adding 4th version number. I don't believe users would be confused if appropriate note is given.

However, IMO all that is overkill for this package. Lets go with simplest thing like you proposed, and if ever needed, we can eventually add virtual package for blessings.

majkinetor:
> Like 4.8.1 is now on state Released but probably would become ReleasedAndBlessed at some point.

What is the point of streams then ? Blessed stream will always have push conflict, right?

majkinetor:
I believe adding just 1 new line in blessed section: `PackageName = 'nuget.commandline.blessed'` (or something shorter) would make it work with 2 packages being updated.

pascalberger:
@majkinetor I've now implemented it simpler, as it just checks for the newest entry in the list regardless of the release state

PR #1116, opened 2018-09-25
(poi) Added poi.

drelatgithub:
## Description
Integrated and automated [poi](https://poi.io), a scalable kancolle viewer. ([source](https://github.com/poooi/poi))

## Motivation and Context
Poi is a browser and toolset especially designed for the web browser game Kantai Collection (KanColle). It is free and open source, and gains popularity among KanColle players. This is to integrate the installation of poi, and automation is needed since poi is updated frequently.

## How Has this Been Tested?
The install/uninstall processes have been tested on Windows 10 pro locally and on the chocolatey-test-environment.

## Types of changes

## Checklist:


automatic/poi/README.md
```diff
@@ -0,0 +1,5 @@
+﻿# <img src="https://cdn.rawgit.com/poooi/poi/master/assets/icons/poi.png" width="48" height="48"/> [poi](https://chocolatey.org/packages/poi)
+
+[POI](https://poi.io) is a scalable KanColle browser and tool.
+
+More information could be obtained [here](https://github.com/poooi/poi/blob/master/README.md)
```
AdmiringWorm: This description is way too little, please add more description to the Readme.
As well as adding a `## Features` section of the different features of the software.

automatic/poi/poi.nuspec
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <!-- == PACKAGE SPECIFIC SECTION == -->
```
AdmiringWorm: You don't need these comments, however I'm not going to require you to remove them. Just an FYI

automatic/poi/update.ps1
```diff
@@ -0,0 +1,48 @@
...
+        if($stableUrls.Count -gt 0) {
+            $url = $stableUrls | select -First 1
+            $stableReleaseFound = $true
+        }
+        else {
+            $lastUrl = $urls | select -Last 1
+            $lastVersionString = "?after=$(Split-Path(Split-Path($lastUrl)) -Leaf)"
+        }
+    }
+    until($stableReleaseFound)
```
AdmiringWorm: This could potentially loop infinitely, lets say sometime in the future they change how they name the executables. In this case (eventually) this do loop would never exit.

Would be better in to change the match for the links (further up) to something along the lines of:
`$urls $download_page.links | ? href -match 'poi-setup\-d+\.\d+\.\d+\.exe$' | select -first 1 -expand href`
(you could also use the regex `poi-setup\-\d+(\.\d+){1,2}\.exe$`, or just `\d+(\.\d+){1,2}\.exe$`), you'll need to test those first of course.

automatic/poi/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,19 @@
...
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+
+$packageArgs = @{
+  packageName   = $env:ChocolateyPackageName
+  fileType      = 'exe'
+  file          = "$toolsDir\"
+
+  softwareName  = 'poi*'
+
+  silentArgs    = '/S'
```
AdmiringWorm: I see that this software with this is only installing as the current user, is it possible to pass in an argument to have it install as All Users instead.
If possible, all users installation should always be used with chocolatey, due to the nature of the software.

automatic/poi/poi.nuspec
```diff
@@ -0,0 +1,29 @@
...
+    <id>poi</id>
+    <version>9.0.0</version>
+    <owners>chocolatey, POI Contributors</owners>
```
AdmiringWorm: POI Contributors in this case should be replaced with the user you have on chocolatey, your name or your github username.
This is meant as the owner of the package, not the software.

automatic/poi/poi.nuspec
```diff
@@ -0,0 +1,29 @@
...
+    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/poi</packageSourceUrl>
+    <!-- ============================== -->
+
+    <!-- == SOFTWARE SPECIFIC SECTION == -->
+    <!-- This section is about the software itself -->
+    <title>poi</title>
+    <authors>POI Contributors</authors>
+    <projectUrl>https://poi.io</projectUrl>
+    <iconUrl>https://cdn.rawgit.com/poooi/poi/master/assets/icons/poi.png</iconUrl>
```
AdmiringWorm: You also need to add this icon the icons directory in this repository, you don't need to update this element though, just add the icon to the icons directory and a script of ours will take care of changing it when the package is updated.

automatic/poi/update.ps1
drelatgithub: The main concern here is that currently I am only searching for stable versions, and I think it is possible (though unlikely) that the first page of the github release is filled all with beta versions (`1.2.3-beta.0`, `1.2.3-beta.1`, etc), so that no stable versions could be found on the first page, so that one might need to turn to the next page. But I think you are right about the infinite loop, so I might need to set a maximum number of pages to search for (like 3), and if no stable versions can be found in the limit, the script will abort with an error. Do you think that this could solve the problem?

From here, I also have a question about the prerelease versions. Currently I'm not considering beta versions, but how should I do the versioning if I want to include those prerelease versions, when NuGet does not allow versions other than pure numbers separated by dots? Because if that is possible, it would be really easy to parse the github page, since I'll just need to select the latest release, regardless of whether it is a stable or pre release.

automatic/poi/poi.nuspec
drelatgithub: Fixed in 5be34ca

automatic/poi/poi.nuspec
drelatgithub: Fixed in 80c1b3e

automatic/poi/poi.nuspec
drelatgithub: Fixed in 80c1b3e

automatic/poi/README.md
drelatgithub: Fixed in 845b10a

automatic/poi/tools/chocolateyinstall.ps1
drelatgithub: Fixed in fbae7a1

drelatgithub:
`LICENSE.txt` will be updated in `update.ps1` (3b51ca2).
`VERIFICATION.txt` has been added (970e551)

automatic/poi/update.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+    $license = (Invoke-WebRequest -Uri "https://cdn.rawgit.com/poooi/poi/master/LICENSE" -UseBasicParsing).Content
```
AdmiringWorm: Please don't download the license like this, because of the following reasons:

1. As it is, downloading it during update provides zero benefits in its current state. If you need to download it during updating, you also need to verify that the license haven't change.
2. Having this code in the `au_GetLatest` means the license will be downloaded every time, this isn't needed. Better to have it in either `au_BeforeUpdate` or `au_AfterUpdate` which will only be called if a new version is available.
3. You're using a rawgit cdn url with a branch name. This is permanently cached the first time the url is requested, and will always provide the same file even if the file have been changed upstream.

Bottom line, unless you're willing to add verification that the license being downloaded is the same type as the one we have before (or a different license that allows distribution), please add the license itself in this PR, and remove the code to download the license.

automatic/poi/update.ps1
AdmiringWorm: There are several ways to do this, but the easiest IMO would be to.
1. Add streams to the package, so every version on the page would be checked (see [librecad](https://github.com/chocolatey/chocolatey-coreteampackages/blob/b80c6f0ae6f799818ba0a9aa436fe7dff3faaa9d/automatic/librecad/update.ps1#L34) as an example)
2. Use the `Get-Version` au helper (also used in the librecad updater)

Adding streams is optional though, but would be the best option.
At a minimum though, a change to restrict the amount of looping is needed to be added.

drelatgithub: Fixed in 155038f. However I still have a small question. (Using the conversation below)

automatic/poi/update.ps1
drelatgithub: Fixed in f7c8b6b.

drelatgithub:
Thanks! My chocolatey username is drel.

AdmiringWorm:
@drelatgithub I've added you as a maintainer for the package, you should've received an email from chocolatey were you'll need to confirm being added as a maintainer

drelatgithub:
Yes, I've received the email. Thanks!

It seems that the package is rejected by the chocolatey push server. Is there anything I should do about it?
> Push ERROR
> Attempting to push poi.9.0.3.nupkg to https://push.chocolatey.org Failed to process request. 'Gateway Time-out'.  The remote server returned an error: (504) Gateway Timeout..

AdmiringWorm:
this is a known problem, and happens from time to time.
The package usually is pushed successfully anyhow. So no action needed.

PR #1126, opened 2018-10-18
(vcredist140, vcredist2017) Changed to automatic packages

AdmiringWorm:
## Description
Added a update script to automate the process of updating the vcredist140 and vcredist2017 packages.

## Motivation and Context
Basically so it doesn't have to be manually updated every time.

## How Has this Been Tested?
Since no changes is being done to the package itself (other than fixing the dependency reference on the next update), I only have tested the updating locally by using the test script file located in the scripts directory.

## Types of changes

## Checklist:

- [X] My change requires a change to documentation (this usually means the notes in the description of a package).
- [X] I have updated the documentation accordingly (this usually means the notes in the description of a package).


AdmiringWorm:
I honestly have no idea why the vcredist2017 is failing the uninstallation (somehow choco can't find the nupkg file)

AdmiringWorm:
hmm, I believe I found the reason it fails. But not found a solution yet though.
Seems to be a bug in my unit testing script that's causing this.

Anyhow, any chance that one of you would be able to review this PR?

/CC @gep13 @jberezanski

majkinetor:
It sure looks like gallery doesn't receive required attention from prominent members, particularly for mega important packages such as this.... This doesn't inspire confidence now that MS supported the project.

/cc @ferventcoder

ferventcoder:
@majkinetor not following your train of thought here. Would you mind elaborating a bit?

gep13:
@AdmiringWorm the issue here seems to be a legitimate one, where it can't uninstall a package due to it being a dependency of another package that is currently installed.  Is this something that we can handle within AU itself, or is it something that we will need to simply ignore in this case?  Other than the failing uninstallation, this PR would seem ready to merge to me.

AdmiringWorm:
@gep13 it's a known issue with the testing script I created. Which I tried to fix in the past, but haven't really been able to just yet.
I opt for it to be ignored.

There is one thing in the script that I have second thoughts about though, see the next comment for details.

automatic/vcredist2017/update.ps1
```diff
@@ -0,0 +1,27 @@
+Import-Module AU
+Import-Module "$env:ChocolateyInstall\helpers\chocolateyInstaller.psm1"
+Import-Module "$PSScriptRoot\..\..\scripts/au_extensions.psm1"
+
+$x64Release = 'https://aka.ms/vs/15/release/VC_redist.x64.exe'
+
+function global:au_SearchReplace {
+  @{
+    ".\vcredist2017.nuspec" = @{
+      "(\<dependency .+? version=)`"([^`"]+)`"" = "`$1`"[$($Latest.Version)]`""
```
AdmiringWorm: I'm having seconds thoughts about restricting this to an exact version.
I think perhaps having a minimum version would be better in this case.

Consider this:

In the future, perhaps vcredist2019 (or later) is released, and is also backwards
compatible with vcredist2015 and vcredist2017.
If we use exact constraint, users wouldn't be able to install the vcredist2019 package
if they already have vcredist2017, unless they first uninstall that package.
Thoughts?

gep13: I think this feeds into future plans for the Chocolatey nuspec, namely:

```
    <!--<provides>NOT YET IMPLEMENTED</provides>-->
    <!--<conflicts>NOT YET IMPLEMENTED</conflicts>-->
    <!--<replaces>NOT YET IMPLEMENTED</replaces>-->
```

So I don't think this is something to worry about here, at this point in time.

AdmiringWorm: > So I don't think this is something to worry about here, at this point in time.

I'm not so sure, that would depend on what comes first.
Will vcredist2019 (or something along the lines) come before choco adds that support or not (of course, only time will tell here).

PR #1129, opened 2018-10-19
KB* packages - Work around Boxstarter issue

catsburg:
## Description
Use module-qualified cmdlet name for Install-WindowsUpdate, to prevent accidental clash with the Install-WindowsUpdate command from BoxStarter. 

## Motivation and Context
When using Boxstarter to install chocolatey packages, the Install-WindowsUpdate command with the same name from Boxstarter gets called. This installs ALL critical Windows updates by default, which is unwanted behavior. To work around this issue, please ensure the correct Install-WindowsUpdate is called. See [here the relevant boxstarter issue](https://github.com/chocolatey/boxstarter/issues/293). The same fix was implemented by jberezanski on the KB-packages he maintains, see [the discussion around this fix here](https://github.com/jberezanski/ChocolateyPackages/pull/54) and [the commit with his fix here](https://github.com/jberezanski/ChocolateyPackages/commit/8472735fc039836e6299157f700f13a6572a3acb).

## How Has this Been Tested?
This fix has been tested using Boxstarter on Windows Server 2016 Standard. 

Before fix: Install-WindowsUpdate from BoxStarter.WinConfig is called, resulting in all critical Windows updates being installed.

After fix: Install-WindowsUpdate from chocolateyInstaller is called as expected.

See the relevant logging [here](https://github.com/jberezanski/ChocolateyPackages/pull/54).

## Types of changes

## Checklist:
- [x] My change requires a change to documentation (this usually means the notes in the description of a package). **Does not require an update to the descriptions. I have however updated the release notes section for KB2999226 accordingly.**
- [x] I have updated the documentation accordingly (this usually means the notes in the description of a package). **Does not require an update to the descriptions. I have however updated the release notes section for KB2999226 accordingly.**
- [ ] The added/modified package passed install/uninstall in the chocolatey test environment. **Can't get this working on my machine, maybe someone else can test this if it is deemed necessary for such a minor change?**
- [ ] The changes only affect a single package (not including meta package). **As per section 3.4 in the contribution guidelines, this falls under the exception 'specific special cases that fix common problem'.**

### DIFF
--- a/manual/kb2670838/Tools/ChocolateyInstall.ps1
+++ b/manual/kb2670838/Tools/ChocolateyInstall.ps1
@@ -16,4 +16,4 @@ $servicePackRequirements = @{
   '6.1' = @{ ServicePackNumber = 1; ChocolateyPackage = 'KB976932' }
 }
 
-Install-WindowsUpdate -Id 'KB2670838' -MsuData $msuData -ChecksumType 'sha256' -ServicePackRequirements $servicePackRequirements
+chocolateyInstaller\Install-WindowsUpdate -Id 'KB2670838' -MsuData $msuData -ChecksumType 'sha256' -ServicePackRequirements $servicePackRequirements
--- a/manual/kb2670838/kb2670838.nuspec
+++ b/manual/kb2670838/kb2670838.nuspec
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
     <metadata>
         <id>KB2670838</id>
-        <version>1.0.20170428</version>
+        <version>1.0.20181019</version>
         <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/manual/kb2670838</packageSourceUrl>
         <owners>chocolatey</owners>
         <title>Platform update for Windows 7 SP1 and Windows Server 2008 R2 SP1</title>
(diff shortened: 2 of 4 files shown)


AdmiringWorm:
I'll be honest, I'm a little hesitant about this change. I see the benefit, but I also see possible problems in the future. Let me explain.

We have no guarantee that the module name will be the same in the future, and if memory serves me right it never was the intention that all extension were to be under the `chocolateyInstaller` module name (I may be wrong though).

You may say that if the module name changes in the future, we could just change the module name being called but it isn't that simple.

Packages in this repository is expected to be compatible with the chocolatey versions that have been available between 1 year ago and today (unless otherwise specified, with for instance a dependency on chocolatey itself).

As such IMO it would be more beneficial to ask @jberezanski to perhaps update the extension package to use a different function name, and make an alias for `Install-WindowsUpdate` if no such function already exist.

That's just my two cents though.

AdmiringWorm:
hmm, I'll say i'm still a bit hesitant.
But it's probably the best option we currently have, and as such I'll merge and get an update out for these packages.

@catsburg thanks for adding these changes.

catsburg:
@AdmiringWorm Thanks for accepting the PR even though I agree this isn't ideal. We should now be able to install VS2017 and other software dependant upon KB-packages through Boxstarter again without triggering unwanted Windows Updates.

PR #1162, opened 2018-12-10
Force packageSourceUrl to be package folder in git

mkevenaar:
## Description
Added script to force packageSourceUrl to be set to the package folder in the repo.
* script based on the `Update-IconUrl.ps1` script already in the repo

## Motivation and Context
To assist someone to find the correct package source, the packageSourceUrl element has been added to the nuspec file. Sometimes only the (github) repo is added. This script forces the packageSourceUrl to be set to the `/automatic/{{ packagename }}` directory

## How Has this Been Tested?
Already running for a while in my own chocolatey-packages repo

## Types of changes

## Checklist:


AdmiringWorm:
While I see the usefulness of your script, I'm not completely sure about how useful it would be to have it run for every build.
Maybe a manual triggering of script would be enough (although the same could be said about my `Update-IconUrl` script though).

Still, I believe it warrants some discussion before we accept this PR.

/CC @majkinetor @gep13

majkinetor:
IMO, This should be AU plugin, available for all AU users, not only here. If implemented as AU plugin here we should just add it among update_all options. However, for this to work we must have plugins that run after each package is updated but before it is pushed: https://github.com/majkinetor/au/issues/116

`GithubRepository `should default to `$Env:github_user_repo` because that is used by all scripts. `$Env:au_root` should be used as defualt for Package folder.

gep13:
@AdmiringWorm this seems like a good suggestion.  While waiting on the feature on AU, I don't see any problems with adding this here, but perhaps create another issue to extract this, and perhaps the Icon update, into a Plugin once it is ready?

majkinetor:
On 2nd thought, this can actually be made as AU plugin right now with minor problem that it will not work the first time since push to gallery happens before push to git repo. But next update, it will be correctly set.

This solution looks more complicated than needed and it reinvents some stuff AU already handles for all like loading of nuspec. 

I guess there is no harm to merge it now here until I digest some stuff.

PR #1191, opened 2019-01-13
(mpc-hc) New maintainer, old team abandoned the app

jack1142:
## Description
Last release of MPC-HC maintained by MPC-HC Team is from July 2017 - they no longer maintain it. However, there is a fork, which is maintained by clsid2 (one of old MPC-HC Team contributors) and it's actively updated. It has the same license, so the package can still be embedded.
Also, this fixes not-silent uninstallation.

## Motivation and Context
App is still maintained, just by someone else now, so I think it should still be updated here.

## How Has this Been Tested?
I checked, if updated update.ps1 script updates the package properly and it does. Also checked if package is properly installed in chocolatey-test-environment and because of that I also found out that package doesn't have silent uninstall, so I fixed it as well.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/mpc-hc/tools/chocolateyUninstall.ps1
+++ b/automatic/mpc-hc/tools/chocolateyUninstall.ps1
@@ -1,0 +1,27 @@
+$ErrorActionPreference = 'Stop'
+
+$packageName         = 'mpc-hc'
+$softwareNamePattern = 'MPC-HC*'
+
+[array] $key = Get-UninstallRegistryKey $softwareNamePattern
+if ($key.Count -eq 1) {
+    $key | ForEach-Object {
+        $packageArgs = @{
+            packageName            = $packageName
+            silentArgs             = "/VERYSILENT /SUPPRESSMSGBOXES"
+            fileType               = 'EXE'
+            validExitCodes         = @(0)
+            file                   = $_.UninstallString.replace(' /SILENT', '').trim('"')
+        }
+        Uninstall-ChocolateyPackage @packageArgs
+    }
+}
+elseif ($key.Count -eq 0) {
+    Write-Warning "$packageName has already been uninstalled by other means."
+}
+elseif ($key.Count -gt 1) {
+    Write-Warning "$($key.Count) matches found!"
+    Write-Warning "To prevent accidental data loss, no programs will be uninstalled."
+    Write-Warning "Please alert package maintainer the following keys were matched:"
+    $key | ForEach-Object {Write-Warning "- $($_.DisplayName)"}
+}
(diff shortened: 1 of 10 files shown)


automatic/mpc-hc/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,27 @@
+$ErrorActionPreference = 'Stop'
```
AdmiringWorm: Why adding an uninstall script?
The uninstaller is created by inno setup which should be detected automatically by chocolateys Auto Uninstaller

majkinetor:
I wouldn't complicate life for dubious semantics.

jack1142:
I don't see that question anymore, but someone asked why I added uninstall script, so in case somebody wonders - uninstaller wasn't silent, because there was msg box asking if user wants to keep app's data. I had to suppress it with `/SUPPRESSMSGBOXES`. Also, uninstaller was executed with `/silent ` instead of `/verysilent`, so the uninstaller window was showed.

gep13:
>@majkinetor said...
I wouldn't complicate life for dubious semantics.

Not sure that I can agree here.  You can think of it as semantics if you want, but there is no getting around the fact that this fork, while it may share the same heritage of code base, is a different application, maintained by different people.  

I would suggest that a new package name be used (which is a trivial change to the existing package source files), which would mean that should (as it is mentioned that in the readme that this is a possibility)

>Development is limited to small bug fixes.

in the future the maintainers of the original application continue development, that new versions of that application can still be pushed to the existing package id.

jack1142:
Okay, assuming you wouldn't want to keep it as the same package, can I send it here with different package id or should I rather place it in my own repo and maintain it myself?
Also, could you help me with the name I should use - just `mpc-hc-clsid2` or something different?

AdmiringWorm:
@jack1142 feel free to send it here under a different name (you can actually just use the same PR, easy enough to just change the folder name, and then the title. Everything else should be changed automatically when it gets updated. If I remember correctly)

> Also, could you help me with the name I should use - just mpc-hc-clsid2 or something different?

Honestly, I have no idea of what to call the new package. But I guess that id is as good as any other.

jack1142:
Okay, when you'll come to the decision, whether the package should have the same or different name, please let me know :)

AdmiringWorm:
@jack1142 I would like to see the package id being changed.

jack1142:
Okay, it's 2 votes for changing and 1 for not changing, so I'll change this pull.

jack1142:
Okay, I changed it.
Also, I left chocolateyUninstall.ps1 for mpc-hc, because it didn't have silent uninstall before anyway, so it will fix it.

AdmiringWorm:
@jack1142 since this is now a new package, if you let me know your chocolatey username, I'll make sure you get added as a maintainer for this package.

PR #1205, opened 2019-01-18
(mp3tag) Support for beta versions

jack1142:
## Description
This adds support for beta versions of mp3tag to update script. As a releases url it uses forum post by developer, which is edited on every update with either stable or beta version (indicated by "Status" in post).
Also, starting with version 2.91h (beta), uninstall can be silent, so I added chocolateyUninstall.ps1 for that.

## Motivation and Context
I'm sure some people use beta versions of this app to have most recent functionality.

## How Has this Been Tested?
I tested if update script works (I also used one of older edits of the post, with stable status and it works as well. After that I tested the package in chocolatey test environment to see if package install and uninstall properly and silently.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/mp3tag/update.ps1
+++ b/automatic/mp3tag/update.ps1
@@ -34,7 +34,8 @@ function global:au_GetLatest {
     $result = $content -match "<strong>\s*Status:\s*</strong>\s*<strong>\s*([a-z]+)\s*</strong>"
 
     if (!$result) {
-      throw "mp3tag status not found on $releases"
+      Write-Host "mp3tag status not found on $releases"
+      return 'ignore'
     }
 
     $status = $Matches[1]
@@ -50,7 +51,8 @@ function global:au_GetLatest {
     if ($status -eq 'Beta') {
       $version += "-beta"
     } elseif ($status -ne 'Stable') {
-      throw "mp3tag status is not recognizable"
+      Write-Host "mp3tag status is not recognizable"
+      return 'ignore'
     }
 
     return @{ URL32 = $url32; Version = $version }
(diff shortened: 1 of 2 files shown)


automatic/mp3tag/update.ps1
```diff
@@ -19,16 +19,25 @@ function global:au_SearchReplace {
...
       throw "mp3tag version not found on $releases"
     }
 
     $version = $Matches[1]
     $versionSuffix = $Matches[2]
+
+    $result = $content -match "<strong>\s*Status:\s*</strong>\s*<strong>\s*([a-z]+)\s*</strong>"
+
+    if (!$result) {
+      throw "mp3tag status not found on $releases"
```
majkinetor: Maybe better to 'ignore' here instead err ? Would you really wont to get error record for this ?

jack1142: If the script won't know if status is Stable or Beta it may push the beta package without -beta suffix (so it would't  count as prerelease).
Possibly we could add -beta suffix for every status that isn't "Stable", but it still wouldn't be perfect. I can change it, I just want you to know that it's a possibility that update script will treat the version as not prerelease when it is or the other way (depending on what we choose to do).
From what I've seen, this format is kept for quite a while, so chances that it will not match in the first place are really small and if it don't match at some point, we would be able to fix it before pushing it as prerelease, if it's stable.

majkinetor: It looks like you misunderstood. Returning 'ignore' will not update the package nor it will make it to throw error, it will show up in ignored section of the report along with your custom Write-Host message.

https://github.com/majkinetor/au#handling-update-errors

jack1142: Oh, sorry, I didn't know 'ignore' is a thing in au.

PR #1209, opened 2019-01-21
(chocolatey-core.extension) Changed key to be single value

AdmiringWorm:
This is a compatibility fix to make the function
compatible with Powershell V2. Without this fix
the 'UninstallString'/'InstallLocation' value
will always be null.

### DIFF
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 1.3.4
+- Bugfix `Get-AppInstallLocation`: Changed key to be forced as a single value (instead of array)
+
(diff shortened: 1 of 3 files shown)


majkinetor:
This is not correct due to the fact that current code has a branch that will never get executed now.

AdmiringWorm:
I'm not sure I follow, how is this not correct?

Basically the only difference is that there won't be an array available, if no key is found the result would be null so it continues the execution further down the list.

I did see I missed something regarding the verbose output of the # of keys found, but other than that...

Of course, I could change it to iterate through all matches, which is something I considered first.

AdmiringWorm:
@majkinetor I've been thinking since my last comment.

Perhaps iterating through the entire list would be more appropriate?
And exit when the first valid path is found?
Thoughts?

PR #1210, opened 2019-01-21
(chocolatey-core.extension) Unescape app name pattern when calling 'Get-UninstallRegistryKey'

### LINKED ISSUE #784: (chocolatey-core.extension) regex pattern vs wildcard pattern usage in Get-AppInstallLocation
(chocolatey-core.extension)  The first parameter for the Get-AppInstallLocation function, $AppNamePattern, is documented/commented to be a "Regular expression pattern" and the function itself uses it as a regex pattern when using the -match operator in its body, BUT Get-AppInstallLocation also calls Get-UninstallRegistryKey, which does NOT use a regex pattern but instead a wildcard pattern.  Regex patterns and wildcard patterns aren't interchangeable like that, however (ex. a regex pattern "Notepad\\+\\+.*" versus wildcard pattern "Notepad++\*").

## Expected Behavior
Since Get-AppInstallLocation takes a regular expression pattern (used with -match operator) as its parameter, it shouldn't use other functions or operators that treat that same parameter value as a wildcard pattern (used with -like operator).  Get-UninstallRegistryKey (called by Get-AppInstallLocation) takes a wildcard pattern.  Because of this interchangeable use, unexpected matches or lack of matches may occur when looking for the path.  Errors occur ("ERROR: Bad argument to operator '-match': parsing "Notepad++\*" - Nested quantifier +..") when passing in a wildcard pattern into Get-AppInstallLocation because of the use of -match (ex. "Notepad++*").

When passing a regex pattern into the Get-AppInstallLocation method, if a regex is pattern is provided, the calls -like operation used in Get-UninstallRegistryKey will be unexpectedly unmatched.

## Possible Solution

Change Get-AppInstallLocation to not call any functions or operators that use wildcard pattern matching, since the parameter in this function is a regex pattern.

I would imagine this would mean maybe adjusting Get-UninstallRegistryKey to have an alternative parameter to pass in a regex pattern instead of a wildcard pattern, then have Get-AppInstallLocation call Get-UninstallRegistryKey with the new regex pattern parameter instead of the default (current) wildcard pattern.  This would allow Get-UninstallRegistryKey to continue to function with existing calls while also supporting an alternative of using regex pattern matching instead of wildcard patterns so that other functions using wildcards could still use the same functions/logic.

1. choco install notepadplusplus -y

I'm not sure what exactly is triggering the error I receive (see gist below), but when I attempt to install "notepadplusplus" (which installs "notepadplusplus.install"), I get a bad argument error due to a wildcard pattern being used.  I discovered this issue in Get-AppInstallLocation in investigating that (https://github.com/chocolatey/chocolatey-coreteampackages/issues/782 was submitted in regards to adjusting the notepadplusplus.install package as a seperate fix & PR https://github.com/chocolatey/chocolatey-coreteampackages/pull/783 has also been submitted as a fix localized to the "notepadplusplus.install" package).

## Context

This occurred when attempting to install the "notepadplusplus.install" package on a Windows 7 Pro 32 bit machine.

## Your Environment
* Package Version used: 1.3.1
* Operating System and version: Windows 7 Pro (Windows v6.1.7601.65536) (32 bit)
* Chocolatey version: v0.10.7
* Install/uninstall gist: https://gist.github.com/michaelray/fa3ad48d81026988810923226d86e7bf

Related issue:
https://github.com/chocolatey/chocolatey-coreteampackages/issues/782

AdmiringWorm:
This is needed in some rare cases when there is an actual
need to use a regex pattern when calling 'Get-AppInstallLocation'.
fixes #784

This PR is dependent on the previous PR I opened: #1209 

Below is a list of which packages have been tested, both before the change (1.3.3) and after the change (1.3.4). Every package have been tested by installing to a custom directory (when possible), and directory not named the same as what checked for when calling `Get-AppInstallLocation`

|Package Name| Version 1.3.3 | Version 1.3.4 | Notes
|-----------:|:-------------:|:-------------:|:------|
|7zip.install|     ✔️       |      ✔️       |       |
|activepresenter|  ✔️       |      ✔️       |       |
|autohotkey.install| :x:     |      ❔       | Wrong Wildcard/Regex passed to function, probably won't match anything at all |
|avidemux    |     ❌       |       ❔       | Wrong Wildcard/Regex passed to function, only default location will match |
|cdburnerxp  |     ✔️       |       ❔       | Not picked up by parsing registry (expected), and directory can't be changed |
|clementine  |     ✔️       |       ❔       |       |
|codeblocks  |     ❔        |       ❔       |       |

More to come

### DIFF
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.3.4
 - Bugfix `Get-AppInstallLocation`: Changed key to be forced as a single value (instead of array)
+- Bugfix `Get-AppInstallLocation`: Added unescaping of app name pattern when calling 'Get-UninstallRegistryKey' ([#784](https://github.com/chocolatey/chocolatey-coreteampackages/issues/784))
 
 ## 1.3.3
 
--- a/extensions/chocolatey-core.extension/CHANGELOG.md
+++ b/extensions/chocolatey-core.extension/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 1.3.4
+- Bugfix `Get-AppInstallLocation`: Changed key to be forced as a single value (instead of array)
+
 ## 1.3.3
 
 - Bugfix `Get-AppInstallLocation`: fix path is directory
--- a/extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
+++ b/extensions/chocolatey-core.extension/chocolatey-core.extension.nuspec
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>chocolatey-core.extension</id>
-    <version>1.3.3</version>
+    <version>1.3.4</version>
     <title>Chocolatey Core Extensions</title>
     <summary>Helper functions extending core choco functionality</summary>
     <authors>chocolatey</authors>
--- a/extensions/chocolatey-core.extension/extensions/Get-AppInstallLocation.ps1
+++ b/extensions/chocolatey-core.extension/extensions/Get-AppInstallLocation.ps1
@@ -40,7 +40,9 @@ function Get-AppInstallLocation {
     $ErrorActionPreference = "SilentlyContinue"
 
     Write-Verbose "Trying local and machine (x32 & x64) Uninstall keys"
-    $key = Get-UninstallRegistryKey $AppNamePattern | select -First 1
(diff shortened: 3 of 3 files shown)


majkinetor:
I am not sure anymore that is the right path to take here. Its also not certain that current packages work with this change, or I would add this one line change back when it was reported. 

Maybe better idea is to execute registry call 2 times, one with re and then with glob if re finds nothing.

AdmiringWorm:
@majkinetor I did actually test a few packages after applying this change, which all seemed to work as expected.

Having a call to execute the registry several times doesn't really seem right either.

Pretty much the only thing that `[registry]::Unescape` does is to remove the backslashes `\` if they are used (which, the only package I've seen that uses them is Notepad++), so it shouldn't make much of a different elsewhere.

But I can do some more testing to make sure.

Issue #500, opened 2016-12-25
Create extension for additional installers


gep13:
To cover the following:

- VSCode Extensions
- Google Chrome extensions
- Powershell Gallery

gep13:
This would cover:

- [ ] googlechrome-editthiscookie
- [ ] googlechrome-emberinspector
- [ ] googlechrome-github-expandinizr
- [ ] googlechrome-zenhub
- [ ] vscode-csharpextensions
- [ ] vscode-ember-cli
- [ ] vscode-ember-frost
- [ ] vscode-icons

gep13:
@pascalberger you might be interested in this for some of your VSCode extensions as well.

@bcurran3 you might be interested in this for some of your googlechrome extensions.

pascalberger:
@gep13 How would the extension look like for VSCode? And what is its advantage over directly installing the extension?

gep13:
@pascalberger the extension would likely contain the two lines that you currently include in the chocolateyInstall.ps1 file, and the function would likely take a single parameter which is the name of the VSCode extension that is being installed.  The benefit being that "if" the installation mechanism for VSCode extensions ever changed, then you change it in one place, the chocolatey extension, instead of in all the packages that are currently doing it themselves in the chocolateyInstall.ps1 file.

pascalberger:
@gep13 OK, got it :)

AdmiringWorm:
@gep13 any update on this one?

gep13:
@AdmiringWorm I haven't forgotten about this one, just somewhat down the list of things to do just now 😢

majkinetor:
Closing due to no activity

bcurran3:
If I can figure out how to make a helper (which is on my todo list for at least one other idea), I'll do the Chrome extension helper. 

New baby due in 4 days, behind in some work, and busy being lazy right now!

majkinetor:
I now think that this is probably not worth it. Majority of packages should be embedded. As such, perhaps extensions for AU would be more welcome to have.

PR #1228, opened 2019-02-18
Add chocolatey-vscode.extension

pascalberger:
## Description
Create a new extension for managing Visual Studio Code extensions.

## Motivation and Context
Provides robust routines for installing and uninstalling Visual Studio Code extensions which can be used in specific packages.

Requires also Core extension to be added as a dependency.

Fixes part of #500 

## How Has this Been Tested?
Tested locally

## Types of changes

## Checklist:

### DIFF
--- a/extensions/chocolatey-vscode.extension/README.md
+++ b/extensions/chocolatey-vscode.extension/README.md
@@ -1,0 +1,52 @@
+# chocolatey-vscode.extension
+
+This is a Chocolatey extension that simplifies building Chocolatey packages which install Visual Studio Code extensions.
+
+## Installation
+
+Install via chocolatey: `choco install chocolatey-vscode.extension`.
+
+The module is usually automatically installed as a dependency.
+
+## Usage
+
+To create a package that uses a function from this extension add the following to the `nuspec` specification:
+
+```xml
+<dependencies>
+    <dependency id="chocolatey-vscode.extension" version="SPECIFY_LATEST_VERSION" />
+</dependencies>
+```
+
+**NOTE**: Make sure you use adequate _minimum_ version.
+
+To test the functions you can import the module directly or via the `chocolateyInstaller.psm1` module:
+
+```powershell
+PS> import-module $Env:ChocolateyInstall\helpers\chocolateyInstaller.psm1
+PS> import-module $Env:ChocolateyInstall\extensions\chocolatey-vscode\*.psm1
+```
+
+You can now test any of the functions:
+
+```powershell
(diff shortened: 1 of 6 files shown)


gep13:
@pascalberger seems like a good suggestion to me!

Does this also work if you provide the path to the vsix, as is done in the `chocolatey-vscode` package?

extensions/chocolatey-vscode.extension/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+This extension requires the `chocolatey-core.extension`.
+
+## Usage
+
+To create a package that uses a function from this extension add the following to the `nuspec` specification:
+
+```xml
+<dependencies>
+    <dependency id="chocolatey-core.extension" version="SPECIFY_LATEST_VERSION" />
```
AdmiringWorm: since this is a required dependency, would probably be better to have this package depend on it.
Instead of relying on the maintainer to specify the dependency.

AdmiringWorm:
@gep13 it uses exactly the same logic, so yes it should work as well

extensions/chocolatey-vscode.extension/README.md
pascalberger: done

pascalberger:
@gep13 It's the same code, so should work. I added also an example to the function to document this posibility.

gep13:
>@pascalberger said...
I added also an example to the function to document this posibility.

So, my only concern then would be that the variable name for the the function doesn't match what is possible.  Should be change the variable name?

pascalberger:
@gep13 Any suggestion for a better name?

`ExtensionNameAndOptionalVersionOrFullPathToVsix` would be accurate but a little bit verbose 😉

AdmiringWorm:
@pascalberger I agree, that would definitely be too verbose.

Maybe having an alias for the parameter could be enough, like aliasing it to `extensionPath` or something similiar.

Just a thought

pascalberger:
I changed parameter name to `extensionId`, which is what VsCode uses for the CLI parameter, and explained in the comment the different options. 

@gep13 @AdmiringWorm Would this work for you?

AdmiringWorm:
It works for me.
@gep13 what about you?

gep13:
>@AdmiringWorm said...
what about you?

>@pascalberger said...
Would this work for you?

Perfect! 👍

majkinetor:
It makes sense to embed the extension for several reasons, so it makes sense to provide quick install option from the tools directory.

PR #1248, opened 2019-03-17
Brave fix icon urls

JipSterk:
## Description
points jsdelivr to the right location for icon hosting 

## Types of changes


AdmiringWorm:
I'm afraid that this PR is incorrect.
The changes you have done, will not allow the icon to be changed in the future.
This is why the url is currently as it is (it uses the current hash of the icon file).

PR #1268, opened 2019-04-24
Improve julia installer by making it silent

musm:
Make the julia installer silent for a cleaner user experience.

## Description
Silent installation of the julia package

## Motivation and Context
Although this change is not required, it makes the user experience cleaner.
In particular, the julia installer provided by julialang is actually an archive. Therefore it makes sense to first extract the exe archive and then run the msi installer.

Notes: both the 32 and 64 bit downloads extract to the same installer name.
 
Addresses: https://github.com/chocolatey-community/chocolatey-coreteampackages/issues/1266

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,7 +15,7 @@ branches:
 
 environment:
   # Set au version to use or omit to use the latest. Specify branch name to use development version from Github
-  au_version: master
+  au_version:
   au_push: true
   # Use 1 to test all, or N to split testing into N groups
   au_test_groups: 4
--- a/automatic/julia/tools/chocolateyinstall.ps1
+++ b/automatic/julia/tools/chocolateyinstall.ps1
@@ -2,11 +2,12 @@
 
 $toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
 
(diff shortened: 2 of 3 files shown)


musm:
I would appreciate it if others could test these changes and report back.

musm:
I'll need some help decoding what is wrong with appveyor

majkinetor:
FYI, the error is not visible due to the regression bug in unreleased version of AU this repository uses: https://github.com/majkinetor/au/issues/181.

musm:
can we expect a new release of `au`  soon?

majkinetor:
That is irrelevant for this problem as it can use latest public version (which can't ignore push errors people currently have). I wanted to release already, but waited to see if drastic change I did works as expected, but it turns out it made this problem. I will make this repository use latest official release so you will see the error.

Also, anybody can actually help resolving the error which is better :)

musm:
Great thanks. Indeed:

> Also, anybody can actually help resolving the error which is better :)


👍

majkinetor:
You can try again with some dummy commit here to trigger the build.

df7acecaadd41ea2a57b22e33c819ead086778f4

AdmiringWorm:
Most likely, the cause of the failure is related to the updater can no longer replace with the necessary file paths, due to the change in the install script.

If you look in the update script, I bet it tries to replace the values for `File` and `File64`

musm:
I will fix the issue through julia

Issue #1270, opened 2019-04-26
(dropbox) Latest version Fails to Install (helper)


iconism:
Failures
- dropbox (exited -1) - Error while running 'C:\ProgramData\chocolatey\lib\drop
box\tools\chocolateyinstall.ps1'.

iconism:
ERROR: The term '\helper.ps1' is not recognized as the name of a cmdlet, functio
n, script file, or operable program. Check the spelling of the name, or if a pat
h was included, verify that the path is correct and try again.
The upgrade of dropbox was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\dropbox\tools\chocolateyinsta
ll.ps1'.

iconism:
2019-04-30 11:28:31,861 9036 [INFO ] - VERBOSE: Importing alias 'Start-ChocolateyProcess'.
2019-04-30 11:28:31,877 9036 [DEBUG] - Running 'ChocolateyScriptRunner' for dropbox v71.4.108 with packageScript 'C:\ProgramData\chocolatey\lib\dropbox\tools\chocolateyinstall.ps1', packageFolder:'C:\ProgramData\chocolatey\lib\dropbox', installArguments: '', packageParameters: '',
2019-04-30 11:28:31,877 9036 [DEBUG] - Running 'C:\ProgramData\chocolatey\lib\dropbox\tools\chocolateyinstall.ps1'
2019-04-30 11:28:31,908 9036 [ERROR] - ERROR: The term '\helper.ps1' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
2019-04-30 11:28:31,939 9036 [DEBUG] - Built-in PowerShell host called with ['[System.Threading.Thread]::CurrentThread.CurrentCulture = '';[System.Threading.Thread]::CurrentThread.CurrentUICulture = ''; & import-module -name 'C:\ProgramData\chocolatey\helpers\chocolateyInstaller.psm1'; & 'C:\ProgramData\chocolatey\helpers\chocolateyScriptRunner.ps1' -packageScript 'C:\ProgramData\chocolatey\lib\dropbox\tools\chocolateyinstall.ps1' -installArguments '' -packageParameters '''] exited with '-1'.
2019-04-30 11:28:31,939 9036 [DEBUG] - Calling command ['"C:\Windows\System32\shutdown.exe" /a']
[...]

PR #1272, opened 2019-05-02
(dropbox) issue with helper.ps1 (#1272) [AU dropbox]

RedBaron2:
@AdmiringWorm @majkinetor @gep13 
This will fix Issue #1270 
Due to a Posh v3 variable used that defaults to null in Posh v2 systems. Changed to use same variable if not defined. Fixed the version check to check for pre version suffix and remove it.

## Types of changes

## Checklist:

## Original Location
  - [x] Issue #1270 
  - [ ] Issue 2 Link

### DIFF
--- a/automatic/dropbox/tools/chocolateyinstall.ps1
+++ b/automatic/dropbox/tools/chocolateyinstall.ps1
@@ -1,9 +1,11 @@
 ﻿$ErrorActionPreference  = 'Stop'
+ if(!$PSScriptRoot){ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent }
 . "$PSScriptRoot\helper.ps1"
 
 $version = '72.3.127-beta'
 
 if (!(IsVersionAlreadyInstalled $version)) {
+    if ( $version -match "beta" ) { $version = $version -replace("-beta","") }
     $stop_dropbox = if (Get-Process -Name Dropbox -ErrorAction SilentlyContinue) {$false} else {$true}
 
 
--- a/automatic/dropbox/tools/chocolateyinstall.ps1
+++ b/automatic/dropbox/tools/chocolateyinstall.ps1
@@ -5,7 +5,6 @@
 $version = '72.3.127-beta'
 
 if (!(IsVersionAlreadyInstalled $version)) {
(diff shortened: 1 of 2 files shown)


automatic/dropbox/tools/chocolateyinstall.ps1
```diff
@@ -1,9 +1,11 @@
 ﻿$ErrorActionPreference  = 'Stop'
+ if(!$PSScriptRoot){ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent }
 . "$PSScriptRoot\helper.ps1"
 
 $version = '72.3.127-beta'
 
 if (!(IsVersionAlreadyInstalled $version)) {
+    if ( $version -match "beta" ) { $version = $version -replace("-beta","") }
```
AdmiringWorm: Why did you add this line?
It isn't used anywhere else, other than in the version installed check.

automatic/dropbox/tools/chocolateyinstall.ps1
```diff
@@ -1,9 +1,11 @@
 ﻿$ErrorActionPreference  = 'Stop'
+ if(!$PSScriptRoot){ $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent }
```
AdmiringWorm: :+1:

automatic/dropbox/tools/chocolateyinstall.ps1
RedBaron2: The version check is trying to check the version installed which dropbox doesn't install with a `-beta` tag. I can make a gist of what I'm saying because I have the `--pre` version installed. The current check is a `False` and will install the dropbox version again.
Mainly it is a fix that should go in the `helper.ps1` file.

RedBaron2: I'll make the change in the update file so a valid version is used for the check.

AdmiringWorm: If that was the intention, then you added the replacement check in the wrong place, it would have needed to be added before the version check takes place, not after.

Anyways, it is better to just pass the actual version during update than doing such replacements during install.

AdmiringWorm:
I'll force out a new stable release of dropbox with this fix.

PR #1295, opened 2019-06-19
Terminate Skype Process after Installation

### LINKED ISSUE #1009: (skype) Fix for "skype automatically starts after installation (no known way to disable it yet)"
## Expected Behavior

For Skype to NOT run after install.

## Current Behavior

Skype automatically runs after install

## Possible Solution

Add the following post install and pre-uninstall:

```
if ((get-process "skype" -ea SilentlyContinue) -eq $Null){ 
    Write-Host "Skype currently NOT running."
  }else{ 
    Write-Host "Stopping Skype process..."
    Stop-Process -processname "skype"
  }
```
Better yet would be checking if Skype is already running before running Install-ChocolateyPackage and then letting it continue to run or stop running based on previous state.

```
if ((get-process "skype" -ea SilentlyContinue) -eq $Null){ 
    $SkypeRunning="False" # Always false on new install.
   } else { 
    $SkypeRunning="True"
    Write-Host "Skype found running." -foreground magenta 
}
```
  Install-ChocolateyPackage

```
if ($SkypeRunning -eq "False"){
if ((get-process "skype" -ea SilentlyContinue) -eq $Null){ 
    Write-Host "Skype currently NOT running."
  }else{ 
    Write-Host "Stopping Skype process..."
    Stop-Process -processname "skype"
  }
}
```

Install Skype package

## Context

If updated in background (scheduled task) in non interactive mode, Skype can unknowingly be stuck running in the background until next reboot.

## Your Environment

* Package Version used: 7.41.0.10101
* Operating System and version: Microsoft Windows [Version 10.0.16299.371]
* Chocolatey version: Chocolatey v0.10.10 Professional
* Install/uninstall gist: n/a

psdavid-muc:
Skype: After updating Skype, the software starts automatically. This is an unwanted feature/bug/behaviour.

## Description
Before executing the installation process, i've defined a termination variable that gets the process of skype. After installation, i added a "Start-ChocolateyProcessAsAdmin" that executes the contents of the termination variable. 
This way, Skype does not start after installation

## Motivation and Context
- Why is this change required? What problem does it solve?
Currently if you're updating skype, right after the update Skype is started automatically and running as admin (if i remember correctly). That is a unwanted behaviour.

The fix proposed in #1009 imho is not a good solution, because the skype process would be started as SYSTEM since chocolatey runs with SYSTEM rights. 
Unless there is a process spawned in userland that stops and starts processes with current userland rights, this should not be implemented.

- If it fixes an open issue, please link to the issue here.
This fixes #1009 

## How Has this Been Tested?
- Please describe in detail how you tested your changes.
I've created a custom choco package for our organization that works just fine.

## Types of changes

## Checklist:

(I'm sorry, i haven't done this before..)

### DIFF
--- a/automatic/skype/tools/chocolateyInstall.ps1
+++ b/automatic/skype/tools/chocolateyInstall.ps1
@@ -1,5 +1,7 @@
 ﻿$ErrorActionPreference = 'Stop'
 
+$terminateSkypeProcess = 'Get-Process -Name "Skype" | Stop-Process -Force'
+
 $packageArgs = @{
   packageName    = $env:ChocolateyPackageName
   softwareName   = 'Skype*'
@@ -11,3 +13,5 @@ $packageArgs = @{
 }
 
 Install-ChocolateyPackage @packageArgs
+
+Start-ChocolateyProcessAsAdmin -Statements $terminateSkypeProcess


bcurran3:
The problem I see with this is that it will ALWAYS quite Skype after an install/upgrade. It doesn't take into account users who may WANT it continuously running but OK with it being down for a few minutes while it's upgraded.

> Before executing the installation process, i've defined a termination variable that gets the process of skype. 

I'm not trying to be overly critical here, but there's no need for the variable. It never changes and doesn't ever have the possibility to change. It's just a static string that is defined and then used once. Defining it before the install process or after the install process doesn't matter as it's only used one as a parameter to the Start-ChocolateyProcessAsAdmin command.  You can just as easily put it inline with the Start-ChocolateyProcessAsAdmin command. 

After installation, i added a "Start-ChocolateyProcessAsAdmin" that executes the contents of the termination variable.
This way, Skype does not start after installation

Actually Skype DOES start after installation, it just get's halted. :)

majkinetor:
There is simply no sure way to start it AFAIK once it is stopped as there might be two users involved (admin might be separate user). 

Somebody correct me if I am wrong, but If I run skype and then update is requested which must involve separate admin user in which context chocolatey is run and installed, then that user can't start skype on behalf of another non-admin user (this can be done via scheduled task but password of non-admin user will be required). 

If admin and non-elevated user are one and the same, then this can be done via scheduled task.

psdavid-muc:
> I'm not trying to be overly critical here, but there's no need for the variable. It never changes and doesn't ever have the possibility to change. It's just a static string that is defined and then used once. Defining it before the install process or after the install process doesn't matter as it's only used once as a parameter to the Start-ChocolateyProcessAsAdmin command. You can just as easily put it inline with the Start-ChocolateyProcessAsAdmin command.
> 
> > After installation, i added a "Start-ChocolateyProcessAsAdmin" that executes the contents of the termination variable.
> > This way, Skype does not start after installation
> 
> Actually Skype DOES start after installation, it just get's halted. :)

Yes, you're right, i should have found better words for that. And using a variable once instead of pasting the path is just laziness.

> Somebody correct me if I am wrong, but If I run skype and then update is requested which must involve separate admin user in which context chocolatey is run and installed, then that user can't start skype on behalf of another non-admin user (this can be done via scheduled task but password of non-admin user will be required).
> 
> If admin and non-elevated user are one and the same, then this can be done via scheduled task.

Definitely. If you think about this in corporate context, you don't want users to be able to do stuff with admin-rights, so we don't allow users to be able to update Skype on their own.

With that, *restarting skype* is very complex. I imagine you would need a seperate process or service (that belongs to chocolate) that runs in user space and waits for stuff like "hey, i've just killed the skype process, after i'm done updating, be a dear and start that process in your userspace again." But i imagine that would require a rework of the chocolatey architecture.

So, yes, killing the Skype process after updating isn't very attractive, but a stupid/lazy and simple way to get it done. I know it might be impractical for the user but that is something i can neglect, since they just start skype again.

psdavid-muc:
@majkinetor do you have any other ideas on how to restart the skype process?
If not, would it be possible to merge this pull request?

majkinetor:
The right code must:

- Check if Skype is already running prior to installation and remember that
- If it was running before, do not kill it.
- If it should kill it, it needs to handle if it is not running (for example slow computer) and wait for it to start for some time.

psdavid-muc:
I see.
But there is still the problem of elevated and non-elevated context: For a script to restart the process in the user context you have to provide user credentials. AFAIK there is no automatic resolution to this, since user accounts and passwords are not unique.

Determining what user is currently logged on might be possible, but switching into that without asking the user for a password is impractical.

Writing a script that - through a parameter - restarts skype for the current user is fine if the current user is a admin user, but won't work for non-admin users.

majkinetor:
I think we should ignore that as there is no solution to that problem. I believe that majority of choco users run as admins and do not have this problem so its unreasonable to make package worse because of few users. Also, we already used the exact same thing on couple of other packages before we realized that this might happen.

Also, I am judging from the fact that nobody complained, at least here.

AdmiringWorm:
> Also, I am judging from the fact that nobody complained, at least here.

And none on disqus, through google groups or through email either. At least not that I can remember.

psdavid-muc:
So i'll just close the PR then?

majkinetor:
You can provide said changes if you want. I said nobody complained about elevation issues so it isn't as important.

PR #1311, opened 2019-08-06
libreoffice-still fix 404 errors

psdavid-muc:
## Description
I've created two powershell functions to show a available download link to the 32bit and 64bit installer for the LibreOffice verstion based on a given hash value.

## Motivation and Context
If a newer version of - for example - LibreOffice Still - is published, the previous version (say 6.0.7) is moved to the archive. But not as a simple downloadable link, rather then there are some more versions generated (for example 6.0.7.1, 6.0.7.2, 6.0.7.3). 

On the initial release of the choco package, the given SHA256 value is the same as the archived 6.0.7.3 version. To determine which archived version is a SHA256-matching one i've created the powershell function "Get-ArchiveURL". It needs a Version string and two Hash-Strings. Based on the hash string and the version the matching archive URL is being provided.

I've modified the chocolateyInstall-script in a way that, if the current version has not been archived yet, the packageArgs will be used unaltered.
But if the version has been archived, only the URL parameters of the packageArgs are modified, since based on the SHA256 value the software is the same.

This not only works with libreoffice-still but also with libreoffice-fresh. This package basically provides always either the official current release download link or the archived one.

## How Has this Been Tested?
I've tested this with a legacy version (6.0.7) and a current release (6.2.5).

## Types of changes

## Checklist:

## Original Location
- Original Repository: https://github.com/chocolatey-community/chocolatey-coreteampackages/tree/master/automatic/libreoffice-streams
- Open Issues:
 - i'm surprised that no issue is opened for this yet.

### DIFF
--- a/automatic/libreoffice-streams/tools/chocolateyInstall.ps1
+++ b/automatic/libreoffice-streams/tools/chocolateyInstall.ps1
@@ -1,16 +1,32 @@
 ﻿$ErrorActionPreference = 'Stop'
 
+Import-Module $PSScriptRoot\utils\archive-helper.psm1
+
 $packageArgs = @{
   packageName            = 'libreoffice'
   fileType               = 'msi'
-  url                    = 'https://download.documentfoundation.org/libreoffice/stable/6.2.5/win/x86/LibreOffice_6.2.5_Win_x86.msi'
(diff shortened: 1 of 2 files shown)


psdavid-muc:
AFAIK @wget is already working on a proper solution.

PR #1340, opened 2019-09-24
[gom-player] Bump to version 2.3.45.5307

usta:
Without this emergency PR choco couldn't install gom player because of gomlab doesn't keep its versions in the archive and new versions always have with the same name with the older name. Thus gom-player's old version's ( without this PR ) sha256 hash never ever matches with the current one and choco users couldn't able to install gom-player ( without --ignorechecksum flag)


AdmiringWorm:
Thank you for the pull request, but unfortunately this is not the correct way to update a AU chocolatey package (which is almost every package in this repository), setting the latest version would actually prevent au updater to update the package at all. It is possible that a new updater.ps1 script is needed to get these changes across (unfortunately, I do not got the possibility to look into that personally until next month).

Bottom line:
I appreciate the pull request, but unfortunately the changes needs to be made in the `update.ps1` script, and potentially used a custom push command (we take care of that command) before this could eventually be merged.

majkinetor:
This looks like an upstream problem. AU updater looks here:

* https://www.gomlab.com/ajax/update.gom?page=1&lang=eng&product=GOMPLAYER

which is official update info page and it doesn't contain latest version. Once it does, AU will update it.  

So I would report it to Gom Player devs.

Alternative is to ask the same crew for the redistribution rights, then we can embedd player in the package and stuff like this one would never happen.

PR #1350, opened 2019-09-30
(tvbrowser) update++

### LINKED ISSUE #1349: (tv-browser) Package is outdated
TV browser complains in AU report.

majkinetor:
I tried to fix #1349 ~but couldn't finish with testing~.

I used https://www.tvbrowser.org/index.php?id=windows as download location which provides _full_ version that doesn't require JRE but it actually pops up in the middle of installation this:

image

This installer ignores `/S` parameter so it must be handled via AHK.
So I abandoned it for the _lite_ versions which comes without JRE.

Update works fine and install too, but if app doesn't find java it will open browser only for u to download their jre.

It also mentions that adoptopenjdk is supported so it should work with this package:

* AdoptOpenJDK11JRE or
* AdoptOpenJDK

However, those do not set registry stuff by default which is a problem since AFAIK you can't pass parameters to dependencies in nuspec file (while there is some `globalparams` on cli available it isn't of much use here).

This seemed to work on local test and I could run the app without any problems after it:
```
cinst adoptopenjdk -ia ADDLOCAL=FeatureMain,FeatureJavaHome,FeatureEnvironment,FeatureOracleJavaSoft

# The above command doesn't work correctly with adoptopnejdk11jre
```

but I didn't do testing within VM so I can't be sure. Note that JRE is better option to have here.

## EDIT

Copying/Symlinking entire adoptjre install into tv browser install dir under subfolder `java` works.

```
New-Item -Type SymbolicLink -Path 'C:\Program Files\TV-Browser\java' -Value 'C:\Program Files\AdoptOpenJDK\jdk-11.0.4+11-jre'
```

Alternativelly, I found that adding following registry key after `cinst adoptopenjdk11jre` works:
```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SOFTWARE\JavaSoft\JDK\11]
"JavaHome"="C:\\Program Files\\AdoptOpenJDK\\jdk-11.0.4+11-jre\\"
````

So, depending on `adoptopenjdk11jre` with above registry snippet added after installation should work. Not sure if it will colide with something tho... 

So IMO symlink is better idea, if done in posh2 compatibile manner (via cmd.exe mklink).

/cc @AdmiringWorm

### CODE REVIEW

**Comment by majkinetor** [comment]
I opted for symlink and this looks like working fine.

File: automatic/tv-browser/tv-browser.nuspec [hunk 1]
```diff
@@ -12,22 +12,42 @@
...
+![screenshot](https://cdn.rawgit.com/chocolatey/chocolatey-coreteampackages/automatic/tv-browser/screenshot.png)
 ]]></description>
     <summary>TV-Browser is a digital TV guide.</summary>
     <releaseNotes />
     <copyright />
     <tags>foss cross-platform tv-browser tv television guide admin</tags>
     <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/automatic/tv-browser</packageSourceUrl>
     <dependencies>
-      <dependency id="javaruntime" version="6.0.30" />
+      <dependency id="adoptopenjdk11jre" />
```
AdmiringWorm: Should specify a minimum version as per [1.1.10 Dependency versions](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md#1110-dependency-versions)

automatic/tv-browser/update.ps1
```diff
@@ -12,26 +12,19 @@ function global:au_SearchReplace {
       "(?i)(^\s*checksum\s*type\:).*"     = "`${1} $($Latest.ChecksumType32)"
       "(?i)(^\s*checksum(32)?\:).*"       = "`${1} $($Latest.Checksum32)"
     }
-    ".\tools\chocolateyInstall.ps1" = @{
-      "(?i)(^\s*file\s*=\s*`"[$]toolsPath\\).*"   = "`${1}$($Latest.FileName32)`""
```
AdmiringWorm: Why do you keep removing this every single time you touch a package?

Do you enjoy creating problems further down the line? Or unneeded questions by users of the package?

automatic/tv-browser/tools/chocolateyInstall.ps1
```diff
@@ -5,12 +5,20 @@ $toolsPath = Split-Path -parent $MyInvocation.MyCommand.Definition
...
   validExitCodes = @(0)
+  softwareName   = 'TV-Browser*'  
 }
-
 Install-ChocolateyInstallPackage @packageArgs
-
 Get-ChildItem $toolsPath\*.exe | ForEach-Object { Remove-Item $_ -ea 0; if (Test-Path $_) { Set-Content "$_.ignore" } }
+
+$jreDir = gi $Env:ProgramFiles\AdoptOpenJDK\jdk-11*-jre | select -First 1
+if (!$jreDir) { Write-Warning "Couldn't find AdoptOpenJDK JRE folder"; return }
```
AdmiringWorm: Shouldn't we test for the adoptopenjdk directory before installing the application? And then throw an exception if it doesn't exist?

automatic/tv-browser/update.ps1
majkinetor: Lets agree to disagree on that one, then.

automatic/tv-browser/tools/chocolateyInstall.ps1
majkinetor: I am testing (`!$jreDIr`). 
I opted to let the app bee even without this as a) it still might work depending on JRE you already have b) its relativelly easy to fix by hand if needed. I could throw there too..

automatic/tv-browser/tv-browser.nuspec
majkinetor: Added.

automatic/tv-browser/tools/chocolateyInstall.ps1
majkinetor: All of this is removed in latest commit due to newer JRE dependency setting it correctly.

PR #1370, opened 2019-10-24
(python3) add test for installdir

### LINKED ISSUE #1303: (python) Wrongly reports install success and location when previous install exists
I tried to install python through choco this morning on a new machine.

While choco reported successful installation on my machine at C:\Python37.

ChocoPython3Issue

To my surprise, 'python' was reported as an unknown command. There was no C:\Python37 either.

I proceeded to edit the chocolatey install script 'C:\ProgramData\chocolatey\lib\python3\tools\chocolateyInstall.ps1' to remove the /quiet switch and only then it became clear what the issue was:

2019-07-04 09_53_52-tools

The machine I received from my employer had VS2019 deployed and this held an install of Python in a non-standard location with no entry on PATH.

While I am aware this is an issue specific to my machine, I think the package should have some kind of logical check to ensure the binaries are actually installed: the absence of error codes does not necessarily indicate success...

Issues like this can easily confuse users (I also spent 30 mins figuring it out :) )

majkinetor:
closes #1303

### DIFF
--- a/automatic/python3/README.md
+++ b/automatic/python3/README.md
@@ -4,7 +4,7 @@ Python 3.x is a programming language that lets you work more quickly and integra
 
 ## Package Parameters
 
-- `/InstallDir` - Installation directory
+- `/InstallDir` - Installation directory. **NOTE**: If you have pre-existing python3 installation, this parameter is ignored and existing python install location will be used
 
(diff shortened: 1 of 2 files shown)


AdmiringWorm:
@majkinetor thank you for taking care of the check 😃

PR #1374, opened 2019-11-01
vcredist140 14.24.28127.4

ImportTaste:
Microsoft has had a bad habit as of late when it comes to their vcredist download locations, it was a pain to find this outside of automatic updates through Windows Update.

### DIFF
--- a/automatic/vcredist140/tools/data.ps1
+++ b/automatic/vcredist140/tools/data.ps1
@@ -1,12 +1,12 @@
 ﻿$installData32 = @{
-  Url = 'https://download.visualstudio.microsoft.com/download/pr/9565895b-35a6-434b-a881-11a6f4beec76/4A8157B2FF422C259DDAA2D0E568C0C0AFAB940E1F6E0E482EF83E90DDBAD2D6/VC_redist.x86.exe'
-  Checksum = '4a8157b2ff422c259ddaa2d0e568c0c0afab940e1f6e0e482ef83e90ddbad2d6'
+  Url = 'https://download.visualstudio.microsoft.com/download/pr/348da5f2-c5d4-4fbf-8360-d1b907780672/E59AE3E886BD4571A811FE31A47959AE5C40D87C583F786816C60440252CD7EC/VC_redist.x86.exe'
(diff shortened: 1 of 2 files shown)


AdmiringWorm:
We can not update the package to a version that isn't listed anywhere.

There is a requirement for chocolatey packages for being able to verify the binaries being installed is the same as the ones from an official download location (ie, a website).

As this version is only available from windows update (or so it seems anyway, as I have been unable to find it mentioned anywhere else), there is no way to verify that except for doing it through windows update. As such it can not be made available on chocolatey.org yet.

AdmiringWorm:
I will leave this PR open however, until there is a public download location for the version.

ImportTaste:
@AdmiringWorm Did you look at the commit? I put the URLs and checksums for the official binaries in already. You can download them and see for yourself.

AdmiringWorm:
@ImportTaste yes, I did look at the commit.

I also said `download location`, not `download url`.
These two are not the same, the one you are talking about is the `download url`, while the `download location` would be the page a user needs to navigate to to find that download url.

ImportTaste:
I see. Well, I think it would be a good idea to at least make it a prerelease package. Microsoft has been known to not be the most organized in regards to the vcredist binaries. They leave so many outdated versions up that are top results on Google, and sometimes the most up-to-date versions are quite hard to find, like vcredist2013's being hidden away on a [support page](https://support.microsoft.com/en-us/help/3138367/update-for-visual-c-2013-and-visual-c-redistributable-package) that isn't even linked to anywhere else on their website.

But if you don't want to do that, no big deal, I won't press further.

AdmiringWorm:
> I see. Well, I think it would be a good idea to at least make it a prerelease package.

Even if publishing it as a pre-release, it would still be the same problem when it comes to moderation (sure, it skips human moderation, but so do every package in this repo. Which is why we need to follow the guidelines before we merge anything in).

> But if you don't want to do that, no big deal, I won't press further.

While I would love to, it can't (as I said earlier) happen until there is a page we can browse to grab the link.

AdmiringWorm:
I noticed today that the package have been updated to the version you wanted to add in this PR.

So apparently MS have figured out their mistake, and finally provided the latest version on the page we parse.

As such this PR is no longer needed, I do appreciate you trying to help out though. 😃

PR #1386, opened 2019-11-18
(chocolatey-core.extension) Adds support for common powershell parameters

AdmiringWorm:
This PR adds support for the most common powershell
arguments and switches to the function that did not implement
them, or did not fully support them.


majkinetor:
Those commands can be used outside of Chocolatey as well.

AdmiringWorm:
@mwallner there is no difference when running any of the functions through chocolatey, these common parameters are only useful if/when debugging outside of chocolatey.

> is there a scenario where this could get back to us in terms of "freezing" a choco install?

I have not found such as scenario during my testing.

PR #1401, opened 2019-12-29
(vim)Symlink to new version

matsuhav:
Delete old version and turn them into symlinks to new version.

## Description

Overwrite old version's directory with symlink to new version's directory.
When there are already symlinks, relink it.
This example illustrates my aim:
```
**First install
C:\tools\vim\vim80

**Updated to vim81
C:\tools\vim\vim80 -> vim81
C:\tools\vim\vim81

**Updated to vim82
C:\tools\vim\vim80 -> vim82
C:\tools\vim\vim81 -> vim82
C:\tools\vim\vim82
```
## Motivation and Context

Current vim package doesn't delete old version when minor version upgraded.
Vim's minor version has updated recently, and now most users have two versions on their machine.
```
**Current situation
C:\tools\vim\vim81
C:\tools\vim\vim82
```
This is actually my mistake. I overlooked this upgrade scenario when I fixed this package.

This change turns `vim81` into symlink to `vim82` so that we can keep using the old path.
This is very convenient because we don't have to change taskber icon, tile on Start and all other settings which uses vim's path.

In this change, I had to use `mklink` and `rmdir` in `cmd.exe` because `New-Item -ItemType SymbolicLink` fails on test-env and Powershell can't unlink directory symlinks. It will delete link-target directory's
contents too.

## How Has this Been Tested?

I tested in the chocolatey test environment.

Install, upgrade scenario. 

## Types of changes

## Checklist:

### DIFF
--- a/automatic/vim/tools/chocolateybeforemodify.ps1
+++ b/automatic/vim/tools/chocolateybeforemodify.ps1
@@ -1,5 +1,5 @@
 ﻿$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
-$installDir = Get-Content -Path "$toolsDir\installDir"
+$installDir = Get-Content "$toolsDir\installDir"
 $shortversion = '82'
 try {
   # Is dlls locked?
--- a/automatic/vim/tools/chocolateyinstall.ps1
+++ b/automatic/vim/tools/chocolateyinstall.ps1
@@ -22,6 +22,6 @@ $installArgs = @{
 
 Install-ChocolateyZipPackage @packageArgs
 Start-ChocolateyProcessAsAdmin @installArgs
-Copy-Item "$installDir\vim\vim$shortversion\vimtutor.bat" $env:windir
+Copy-Item -Path "$installDir\vim\vim$shortversion\vimtutor.bat" -Destination $env:windir
 Set-Content -Path "$toolsDir\installDir" -Value $installDir
 Create-SymbolicLink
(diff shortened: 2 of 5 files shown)


majkinetor:
Maybe better solution would be to always update the same folder (and change vim path to just 'vim' without version or 'vim8') ?

matsuhav:
After I saw your comment, I tried the renaming strategy.
And I found that Vim's installer program which is called in `chocolateyinstall.ps1` fails when directory is renamed. The installer requires it is `vim82` and writes it in the registry. We can't rename it.

majkinetor:
> We can't rename it.

Yes we can. There is even a package parameter `InstallDir`.

matsuhav:
`C:\tools` is `$InstallDir` in the first example. Version prefixed directory such as `vim\vim82` is created automatically by unzipping. We can't change this `vim82` name.
```
** After unzip
$InstallDir\vim\_vimrc
$InstallDir\vim\vim82\install.exe
$InstallDir\vim\vim82\vim.exe
$InstallDir\vim\vim82\gvim.exe
...
```
Vim's installer is `$InstallDir\vim\vim82\install.exe`. It creates shortcuts and writes to registry.
It requires it is in `vim82` directory. [This](https://github.com/vim/vim/blob/master/src/dosinst.c#L203) is the error printing line in the source code.
We can't rename `vim82` to `vim` before running `install.exe`, and we can't do it after since all shortcuts already point to `vim82`.

The folder structure I tried was like this.
```
** After unzip and rename
$InstallDir\vim\_vimrc
$InstallDir\vim\vim\install.exe <- Not run!
$InstallDir\vim\vim\vim.exe
$InstallDir\vim\vim\gvim.exe
...
```

majkinetor:
OK, it looks like you have done your homework. Thanks.

AdmiringWorm:
Not really a fan on making symlinks in cases like this, would have preferred that the old directory was moved to the new location instead, and update any shortcuts that had been created in previous versions.

But nonetheless, it is better than leaving shortcuts pointing to the wrong versions.

AdmiringWorm:
@matsuhav your changes have been merged, thanks for your contribution 👍

THese changes will be available when the next version of VIM is released.

matsuhav:
Thank you very much!

Issue #1409, opened 2020-01-09
(firefox) extend package to integrate the installer command-line options


FreeMinded:
The Firefox Installer support a range of command-line options [1]. This package currently has no support for them. The command-line options are useful i.e for defining if a desktop shortcut should be created, if the configurations set through distribution/policies.json should be kept and a few more.

## Expected Behavior
Command-line Options can be passed to the installer through the --params mechanism.

## Current Behavior
As a workaround command line options can be passed using --install-arguments or -ia. But this does not allow to check for correct input or to set defaults in the package which differ from the installer defaults.

## Possible Solution
Implement package parameters in the package. An example implementation could be seen in the Wazuh agent package [2]. 

1: https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/FullConfig.html
2: https://github.com/itigoag/chocolatey.wazuh-agent/blob/master/package/tools/chocolateyinstall.ps1

FreeMinded:
We actually offer to create a pull request for this if there is a reasonable change to get it approved.

AdmiringWorm:
Feel free to open a PR to add this functionality.

As long as all the changes follow our [Contrubution Guidelienes](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md) there are a high likelihood that it will be accepted.

Be aware that it can take time sometimes before a PR is pulled in, mainly because the main maintainers here is doing all this on their spare time, and may have other priorities than reviewing pull requests.

FreeMinded:
I just updated the PR to reflect the current FF version. Any chance we can get this reviewed before the next release?

PR #1422, opened 2020-02-05
(firefox) extend package to integrate the installer command-line options

FreeMinded:
## Description
integrated command-line installer options to be used as package parameters. See  https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/FullConfig.html#command-line-options

Related Issue: #1409

## Motivation and Context
This change allows to use package parameter for the installer options, not just the locale.

The use of `=`  instead of `:` in the package parameters is due to the locale using it this way already. This could be changed but would probably consist a breaking change in some use cases.

This is my first PR. Happy to learn what could be inproved.

## How Has this Been Tested?
Test by locally building and installing the Firefox package.
I don't know how to test the automated package build process.

## Types of changes

## Checklist:


automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
 
 ## Package Parameters
 
-- `l=<locale>` - Install given Firefox locale. For example `choco install Firefox --params "l=en-GB"`. See the [official page](https://releases.mozilla.org/pub/firefox/releases/latest/README.txt) for a complete list of available locales.
+- `l=<locale>` - Install given Firefox locale. See the [official page](https://releases.mozilla.org/pub/firefox/releases/latest/README.txt) for a complete list of available locales.
+
+Command-line options for installer configuration. See the [official page](https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/FullConfig.html) for details and defaults.
+- `/InstallDirectoryPath=<path>`
```
majkinetor: All other packages use `:`.

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
+- `/InstallDirectoryName=<path>`
```
majkinetor: `/InstallDir` is used by  other packages and choco enterprise.

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
+- `/TaskbarShortcut=<true/false>`
```
majkinetor: We don't use true/false like this. This one is always true if exists.

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
+- `/DesktopShortcut=<true/false>`
```
majkinetor: `/NoDesktopShortcut`, by deafult it is installed.

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
+- `/StartMenuShortcut=<true/false>`
```
majkinetor: `/NoStartMenuShortcut`

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
+- `/MaintenanceService=<true/false>`
+- `/RemoveDistributionDir=<true/false>`
+- `/PreventRebootRequired=<true/false>`
```
majkinetor: This is something done by choco, I would remove this.

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
-- `l=<locale>` - Install given Firefox locale. For example `choco install Firefox --params "l=en-GB"`. See the [official page](https://releases.mozilla.org/pub/firefox/releases/latest/README.txt) for a complete list of available locales.
+- `l=<locale>` - Install given Firefox locale. See the [official page](https://releases.mozilla.org/pub/firefox/releases/latest/README.txt) for a complete list of available locales.
+
+Command-line options for installer configuration. See the [official page](https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/FullConfig.html) for details and defaults.
+- `/InstallDirectoryPath=<path>`
+- `/InstallDirectoryName=<path>`
+- `/TaskbarShortcut=<true/false>`
+- `/DesktopShortcut=<true/false>`
+- `/StartMenuShortcut=<true/false>`
+- `/MaintenanceService=<true/false>`
```
majkinetor: I would remove this param and always specify it, it goes along with unatended spirit of choco.

automatic/firefox/README.md
```diff
@@ -17,7 +17,22 @@ Bringing together all kinds of awesomeness to make browsing better for you.
...
+- `/RemoveDistributionDir=<true/false>`
```
majkinetor: I would remove this one as it is only for uninstall and doesn't seem important in choco context.

automatic/firefox/README.md
FreeMinded: That's correct and I agree. I used `=` because the package is already using it for the language settings `l=<locale>` and I didn't what to change that. If it's required it can be changed of course.

automatic/firefox/README.md
FreeMinded: OK, I used the options as provided by Firefox. I can change it to `/InstallDir` and map it to `/InstallDirectoryPath`.

automatic/firefox/README.md
FreeMinded: You're right. I will change that.

automatic/firefox/README.md
FreeMinded: You're right. I will change that.

automatic/firefox/README.md
FreeMinded: You're right. I will change that.

automatic/firefox/README.md
FreeMinded: OK, didn't know that.

automatic/firefox/README.md
FreeMinded: I read your comment as to not install the Maintenance Service by default. Is that correct? Currently Firefox updates itself when installed through chocolatey. I wouldn't want to change that because I assume many people use chocolatey as "one-time installer" and the rely on the update mechanism off FF.

automatic/firefox/README.md
FreeMinded: This is actually very important and the main reason for this PR. The Distribution directory contains the policies.json file which allows you to configure FF (see https://support.mozilla.org/en-US/kb/customizing-firefox-using-policiesjson). On updating FF this directory get's removed by the installer. I will change it to reflect the always true if exists paradigm.

automatic/firefox/README.md
FreeMinded: I just looked at this. It seems that the FF installer can request a reboot by default. So in my opinion this should be set to true always. Any thoughts about this?

automatic/firefox/README.md
majkinetor: You can also support both if you like.

majkinetor: Maybe, better:

InstallDir = Path\Name

So we have 3 options here. One combinational and standard one, other two are regular FF options. Everybody happy.

automatic/firefox/README.md
majkinetor: OK, why not.

automatic/firefox/README.md
majkinetor: I didn't understand it that way. If it is related to autoupdate, then lots of people would actually like to disable it and use choco instead. So `/NoAutoUpdate` would be nice option for those people. 

I understood that such FF option does that actually. Take a look:

> This service allows Firefox to install updates without requiring you to click yes in the Windows User Account Control (UAC) dialog - Read more at https://www.shouldiremoveit.com/Mozilla-Maintenance-Service-5231-program.aspx

So it looks like it prevents UAC prompt only by runing maintenance service as Windows service. Without it, you still have maintenance service but it is not Windows service, and you have to click. If that is so, that FF option should be alwasy specified (why would I click in choco land?).

Now, if you find a way to implement `/NoAutoUpdate` even better. Such option is HIGHLY needed for developers using Selenium for example as they always need to prevent automatic updates.

automatic/firefox/README.md
majkinetor: Then it should be always set to false  and eventually add `/RemoveDistributionDir` to enable it ?

automatic/firefox/README.md
majkinetor: If you do it that way, you can't use all options together.

InstallDir - ignore Path\Name because it contains it.
No installDir, provide Path and/or Name.

automatic/firefox/README.md
FreeMinded: I just found that we have to change it to use `:` anyway. The way it's currently implemented, the package parameters have to be followed by `=`. I.e. `NoDesktopShortcut` fails.

automatic/firefox/README.md
FreeMinded: Not sure I got you right. By using `/InstallDir:path\name` mapped to `/InstallDirectoryPath=` it should be possible to install anywhere you want.
The installer option `/InstallDirectoryName=` is simply setting the name within the ProgramFiles folder. So this can be ignored as it's possible to accomplish the same with `/InstallDir:path\name`

automatic/firefox/README.md
FreeMinded: You're right, i mixed that up. I like the idea with `/NoAutoUpdate`. If set the MaintenanceService should not be installed. The auto update of FF can be disabled by setting `policies.json`
```
{
  "policies": {
    "DisableAppUpdate": true
  }
}
```

automatic/firefox/README.md
FreeMinded: I totally agree but hesitated to change the defaults. It makes more sense to keep the directory by default, especially in a managed software context.

FreeMinded:
I'm still working on the changes...

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -48,3 +89,24 @@ if ($alreadyInstalled -and ($env:ChocolateyForce -ne $true)) {
 
   Install-ChocolateyPackage @packageArgs
 }
+
+if ($PackageParameters['InstallDir']) {
+  $installPath = $PackageParameters['InstallDir']
+}
+else {
+  $installPath = "C:\Program Files\Mozilla Firefox"
```
FreeMinded: @majkinetor do you know of a way to get the default installation path? This will probably break for 32 bit installs.

FreeMinded: anyone?

majkinetor: Why do you need it ? Can you leave it empty and use `Get-AppInstallLocation` from core extension afterwards to get it when user didn't provide it (if you need it later for something)

FreeMinded: I need to know where to put the `policies.json` files which is in a subdirectory `distribution` in the location of the exe. That's exactly the command I was looking for. Thanks, will use that.

majkinetor: There is also more legit option, to use Registry Uninstaller stuff. From my  machine:

```

Registry key: HKLM\Software\Microsoft\Windows\CurrentVersion\Uninstall\Mozilla Firefox 72.0.1 (x64 en-US)
DisplayName   Mozilla Firefox 72.0.1 (x64 en-US)  
DisplayVersion   72.0.1  
Publisher   Mozilla  
InstallLocation   C:\Program Files\Mozilla Firefox  
UninstallString   "C:\Program Files\Mozilla Firefox\uninstall\helper.exe"  
DisplayIcon   C:\Program Files\Mozilla Firefox\firefox.exe,0  
HelpLink   {https://support.mozilla.org}  
URLInfoAbout   {https://www.mozilla.org}  
URLUpdateInfo   {https://www.mozilla.org/firefox/72.0.1/releasenotes}  
```

FreeMinded:
@majkinetor we just pushed the finalized package. Could you please review and test again. Thanks!

FreeMinded:
We started using this package in production. Everything is working fine. Would be nice to have it merged soon. Thanks for all your help @majkinetor and @sbaerlocher

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -17,26 +19,65 @@ if (Get-32bitOnlyInstalled -product $softwareName) {
   )
 }
 
+$args = ""
+
+# Command Line Options from the Firefox installer
+# https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/FullConfig.html
+
+# Always prevent Firefox installer to require a reboot
+$args = $args + " /PreventRebootRequired=true"
```
majkinetor: Use `+=` please

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -7,6 +7,8 @@ $toolsPath = Split-Path $MyInvocation.MyCommand.Definition
 $packageName = 'Firefox'
 $softwareName = 'Mozilla Firefox'
 
+$PackageParameters = Get-PackageParameters
```
majkinetor: `$pp` please, its repeated a lot and other packages use it.

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -17,26 +19,65 @@ if (Get-32bitOnlyInstalled -product $softwareName) {
...
+# https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/FullConfig.html
+
+# Always prevent Firefox installer to require a reboot
+$args = $args + " /PreventRebootRequired=true"
+
+# Prevent RemoveDistributionDir by default
+$args = $args + " /RemoveDistributionDir=false"
+
+
+if ($PackageParameters['InstallDir']) {
```
majkinetor: There is a 1 line form: 

- `$sa += if ($pp.NoTaskbar) { " /RemoveDistributionDir=false" }`


Also, change `$args`, to some other var name, its used by powershell for function args.

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -17,26 +19,65 @@ if (Get-32bitOnlyInstalled -product $softwareName) {
...
+
+if ($PackageParameters.RemoveDistributionDir) {
+  $args = $args + " /RemoveDistributionDir=true"
+}
+
+if ($PackageParameters.NoAutoUpdate) {
+  $args = $args + " /MaintenanceService=false"
+}
+
 if ($alreadyInstalled -and ($env:ChocolateyForce -ne $true)) {
```
majkinetor: Not correct  because env vars do not support powershell boolean vars.  Just use if `(... -and $Env:ChocolateyForce)`

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -17,26 +19,65 @@ if (Get-32bitOnlyInstalled -product $softwareName) {
...
-    Url = "https://download.mozilla.org/?product=firefox-73.0.1-ssl&os=win&lang=${locale}"
-
-    silentArgs = '-ms'
+    packageName    = $packageName
+    fileType       = 'exe'
+    softwareName   = "$softwareName*"
+    Checksum       = $checksums.Win32
+    ChecksumType   = 'sha512'
+    Url            = "https://download.mozilla.org/?product=firefox-73.0.1-ssl&os=win&lang=${locale}"
+    silentArgs     = "$($args) /S"
```
majkinetor: equiv to `"$args /S"`

automatic/firefox/tools/chocolateyInstall.ps1
```diff
@@ -48,3 +89,24 @@ if ($alreadyInstalled -and ($env:ChocolateyForce -ne $true)) {
 
   Install-ChocolateyPackage @packageArgs
 }
+
+if ($PackageParameters['InstallDir']) {
+  $installPath = $PackageParameters['InstallDir']
```
majkinetor: do not use angles but dots: `$pp.InstallDir` vs `$pp['InstallDir']`

automatic/firefox/tools/chocolateyInstall.ps1
FreeMinded: thanks, this is fixed now

automatic/firefox/tools/chocolateyInstall.ps1
FreeMinded: Great, thanks for this hint. It's implemented now.

FreeMinded:
@majkinetor thanks for the constructive feedback. It's implemented now and upgraded to FF 74.0. Feel free to merge it. :)

PR #1445, opened 2020-03-25
(julia) fix silent install argument

drelatgithub:
## Description

The Julia installer has changed on 1.4.0. `/S` no longer silences the installer, causing the installation to [hang and eventually fail](https://gist.github.com/choco-bot/6b7df821f5bf3c3a990d67c4ddec0a48). This fix changes the silent installation argument from `/S` to `/VERYSILENT`.

## How Has this Been Tested?

Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/julia/tools/chocolateyinstall.ps1
+++ b/automatic/julia/tools/chocolateyinstall.ps1
@@ -10,7 +10,7 @@ $packageArgs = @{
 
   softwareName  = 'Julia*'
 
-  silentArgs    = '/S'
+  silentArgs    = '/VERYSILENT'
   validExitCodes= @(0)
 }
 $packageVersion = "1.4.0"
--- a/automatic/julia/tools/chocolateyinstall.ps1
+++ b/automatic/julia/tools/chocolateyinstall.ps1
@@ -10,7 +10,7 @@ $packageArgs = @{
 
   softwareName  = 'Julia*'
 
-  silentArgs    = '/VERYSILENT'
+  silentArgs    = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
   validExitCodes= @(0)
 }
 $packageVersion = "1.4.0"
--- a/automatic/julia/tools/chocolateyuninstall.ps1
+++ b/automatic/julia/tools/chocolateyuninstall.ps1
@@ -3,7 +3,7 @@ $packageArgs = @{
   packageName   = $env:ChocolateyPackageName
   softwareName  = 'Julia*'
   fileType      = 'exe'
-  silentArgs    = '/VERYSILENT'
+  silentArgs    = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
   validExitCodes= @(0)
 }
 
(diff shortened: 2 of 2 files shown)


drelatgithub:
BTW I always wanted to ask this good old question: is there any "Chocolatey" way of uninstalling specific version of a software when multiple versions are installed? In other words, in the ideal Chocolatey logic, is it even allowed to have multiple versions of the same software installed on one machine?

automatic/julia/tools/chocolateyinstall.ps1
```diff
@@ -10,7 +10,7 @@ $packageArgs = @{
 
   softwareName  = 'Julia*'
 
-  silentArgs    = '/S'
+  silentArgs    = '/VERYSILENT'
```
AdmiringWorm, line 13: I assume their installer is now an `InnSetup` installer, if so can you please make the following change:
```suggestion
  silentArgs    = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
```

automatic/julia/tools/chocolateyuninstall.ps1
```diff
@@ -3,7 +3,7 @@ $packageArgs = @{
   packageName   = $env:ChocolateyPackageName
   softwareName  = 'Julia*'
   fileType      = 'exe'
-  silentArgs    = '/S'
+  silentArgs    = '/VERYSILENT'
```
AdmiringWorm, line 6: Same as the install script, please make the following change:
```suggestion
  silentArgs    = '/VERYSILENT /SUPPRESSMSGBOXES /NORESTART /SP-'
```

AdmiringWorm:
> BTW I always wanted to ask this good old question: is there any "Chocolatey" way of uninstalling specific version of a software when multiple versions are installed?

Other than relying on the automatic uninstaller, then no there isn't a "Chocolatey" way of uninstalling specific versions of a software.
Other than specifying the version in the pattern used to detect the uninstallation registry key (which is highly discouraged to do).

> In other words, in the ideal Chocolatey logic, is it even allowed to have multiple versions of the same software installed on one machine?

It is allowed. Ideally, only one version would be installed at the time though, but since not all software is the same and some can be used with different versions side-by-side there isn't anything to disallow this.

drelatgithub:
@AdmiringWorm Thanks for the clarification! That makes a lot of sense. I asked because I could not find an ideal way of installing this package, because the installer of this software has a side-by-side installing behavior. Now that's the case, do you think it is appropriate to uninstall *all* versions of this software in the uninstall script?

AdmiringWorm:
> Now that's the case, do you think it is appropriate to uninstall all versions of this software in the uninstall script?

I do not believe that would be appropriate TBH. We wouldn't know if the other versions have been installed on purpose by the user or not, so removing all versions of the software could in a few small cases be undesirable for the user.

I do plan to experiment a little with upgrading to tell our install script to remove the old version that was upgraded from, but have not yet started working on something like this yet.

AdmiringWorm:
@drelatgithub thank you for going through the problem of fixing the install and uninstall arguments for this package. I appreciate it.

PR #1450, opened 2020-03-30
(brave) fix the '-2147219198' error

TakataSanshiro:
# Description
Checking the already installed version before downloading the nuspec version.
If the already installed version is greater equal than the nuspec version, then skip installing the outdated version.

## Motivation and Context
Brave is updating automatically faster than the Chocolatey repository.
If upgrading to the outdated version, then an error occurred. (Exit code was '-2147219198')
Some reports about it are available on the package page.

## How Has this Been Tested?
Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/brave/tools/helpers.ps1
+++ b/automatic/brave/tools/helpers.ps1
@@ -1,15 +1,17 @@
-if (-not $env:ChocolateyForce) {
-  try {
-    [array]$key = Get-UninstallRegistryKey -SoftwareName 'Brave*'
-    $installedVersion = $key.Version[3..($key.Version.length - 1)]
-    $installedVersion = "$installedVersion" -replace '\s',''
+function CheckInstalledVersion() {
+  [array]$key = Get-UninstallRegistryKey -SoftwareName 'Brave*'
+  $installedVersion = $key.Version[3..($key.Version.length - 1)]
+  $installedVersion = "$installedVersion" -replace '\s',''
 
-    if ($installedVersion -ge [Version]::Parse($env:ChocolateyPackageVersion))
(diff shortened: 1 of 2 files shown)


automatic/brave/tools/helpers.ps1
```diff
@@ -0,0 +1,15 @@
+[array]$key = Get-UninstallRegistryKey -SoftwareName 'Brave*'
```
AdmiringWorm, line 1: Can you please change the code in this file to be inside a function instead of running immediately when the scripts run?

It would make it easier in the future if there is a need to extend this file with more helper functions.

automatic/brave/tools/helpers.ps1
```diff
@@ -0,0 +1,15 @@
...
+
+    if ($installedVersion -ge [Version]::Parse($env:ChocolateyPackageVersion))
+    {
+      $alreadyInstalled = $installedVersion
+      return
+    }
+  } catch {
+    # Installed version couldn't be checked, attempt installation
+  }
+}
```
AdmiringWorm, line 15: Sorry, I should have been clearer in what I meant.
I did not mean to move the code regarding acquiring the installed version inside the `if` statement, but rather to move all of the code inside this file inside a `function Your-FunctionName` instead, and then call that function from the install script.

TakataSanshiro: Is it ok?

AdmiringWorm:
@TakataSanshiro thank you for your contribution.

I made a few changes to the code, and have queued a new version of the application to be pushed to the community repository

PR #1457, opened 2020-04-16
Update chocolateyInstall.ps1

sffred:
The package sources has a new update

### DIFF
--- a/automatic/gimp/tools/chocolateyInstall.ps1
+++ b/automatic/gimp/tools/chocolateyInstall.ps1
@@ -3,9 +3,9 @@
 $packageArgs = @{
   packageName    = $env:ChocolateyPackageName
   fileType       = 'exe'
-  url            = 'https://download.gimp.org/mirror/pub/gimp/v2.10/windows/gimp-2.10.14-setup.exe'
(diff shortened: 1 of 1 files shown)


RedBaron2:
@sffred 
This is an AU automatic updating package. Please look into checking to find out why the Checksum is wrong? This could possibly be fixed by using something similar to the cutepdf package. 

FYI: The title for this PR is not correct with the contribution guides

sffred:
@RedBaron2 Sorry for my ignorant about the automatic package system and contribution guidelines. However I have to say that the checksum I provided is probably right. 
It is a revision of gimp 2.10.18, and it differs to gimp-2.10.18 you are now providing. Maybe you can help to make it right- I mean the outdated package, not the checksum.

AdmiringWorm:
To make this happen, we, unfortunately, need to rewrite the package update checker to take into account only bumping the revisions when a bug fix is released.

Please subscribe to #1464 where this work will be tracked.

As @RedBaron2 already mentioned, this is an automatic package and as such we do not accept changes to the URL and checksum in the install script. Thus I will be closing this issue in the meantime.

sffred:
@AdmiringWorm Thanks for your reply. I know the situation now, and I'm not able to rewrite the automatic script. Thank the community for your efforts.

Issue #1458, opened 2020-04-16
(firefox) description displayed improperly on chocolatey.org


FreeMinded:
The description is incorrectly displayed on https://chocolatey.org/packages/Firefox

Package Parameters
`/l:&lt;locale&gt;`

![Firefox Readme](https://user-images.githubusercontent.com/1763127/79471101-746fdf00-8002-11ea-995d-682fa1312ea1.png)

Correct would be 
`/l:<locale>`

It's correct in the readme file on githib (I did that myself)

Does anyone know how it could be fixed?

RedBaron2:
@FreeMinded 
The solution is right in front of you.
Replace the entries with the HTML code of `&lt;locale&gt;` for the readme file. The code will show how you want of `<locale>`. The reason is because the nuspec file is an XML file in utf8 encoding. The readme file is translated as the HTML code and then placed in the nuspec file. The translation is automatic due to the encoding.
Reference the package [`wincompose`](https://chocolatey.org/packages/wincompose)

AdmiringWorm:
@RedBaron2 unfortunately that won't help at all.
Because au wraps the actual description inside a `[[CDATA` section, it is the same as the description for the `wincompose` package.

The `[[CDATA` section is used, so you don't have to escape the HTML/XML inside the XML file.

Bottom line, updating the readme to escape the angle brackets won't have anything to say, it will need to be changed to something other than angle brackets

FreeMinded:
Thanks for the Feedback! Looks like it was "wrong" from the beginning. See [here](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/5b41c7da5750bdcd54efcfd0ed462cb3cec9616d/automatic/firefox/Readme.firefox.md)

I just opened a PR.

RedBaron2:
@AdmiringWorm 
The markdown specs seem to be the problem.
> The contents of a code block are literal text, and do not get parsed as [Markdown](https://github.github.com/gfm/#example-80):

@FreeMinded 
I will state that the PR #1460  use of Capital Letters instead of the <> seems like a very good fix

PR #1460, opened 2020-04-17
(Firefox) replace angle brackets with capital letters

FreeMinded:
## Description
Replace angle brackets in Readme with capitals because angle brackets are not displayed correctly on chocolatey.org/packages
 
## Motivation and Context
See #1458

## How Has this Been Tested?
This is purely cosmetically, no functionality change

Any better suggestion is welcome!


AdmiringWorm:
@FreeMinded thank you for making these changes. They will be part of the description for the next version of firefox and firefoxesr

Issue #1480, opened 2020-05-20
(avidemux) Outdated


aminya:
The latest version is 2.7.5
https://github.com/chocolatey-community/chocolatey-coreteampackages/tree/master/automatic/avidemux

gep13:
To add to this, I have just been back to test the previously reject package version:

https://chocolatey.org/packages/avidemux/2.7.4

Looks like there is a problem with the silent installation arguments.  When testing on the Chocolatey Test Environment, it gets to this point:

![image](https://user-images.githubusercontent.com/1271146/82835653-93dd1000-9ebc-11ea-8c5c-c493225d35f8.png)

And then doesn't progress any further.

majkinetor:
This requires some non-trivial update.

Like I said [here](https://chocolatey.org/packages/avidemux#comment-4930608308), we have a working package forever so if there is no personal use in coreteam, there is no incetive to fix random software (we got from 2 active maintainers to 0.1 active maintainers in last year).

aminya:
There is always a possibility to email the original developer: http://avidemux.sourceforge.net/

majkinetor:
That rarelly works, but u cant blame one for trying.

majkinetor:
https://chocolatey.org/packages/avidemux/2.7.5

PR #1488, opened 2020-05-30
(avidemux) Update update.ps1

Hattshire:
to support unmatched releases

## Motivation and Context
#1480 

## How Has this Been Tested?

Raw PS-tested/Not tested

## Types of changes

## Checklist:

### DIFF
--- a/automatic/avidemux/legal/VERIFICATION.txt
+++ b/automatic/avidemux/legal/VERIFICATION.txt
@@ -7,7 +7,6 @@ Package can be verified like this:
 
 1. Go to
 
-   x32: https://sourceforge.net/projects/avidemux/files/avidemux/2.7.4/avidemux_2.7.4_win32.exe/download
    x64: https://sourceforge.net/projects/avidemux/files/avidemux/2.7.4/Avidemux_2.7.4%20VC%2B%2B%2064bits.exe/download
 
    to download the installer.
@@ -16,7 +15,6 @@ Package can be verified like this:
    - Use powershell function 'Get-FileHash'
    - Use Chocolatey utility 'checksum.exe'
 
-   checksum32: 5C4C73841E88FF098216FD325D0A0CED2DD3D395CA39CE9160D1AE1B70C841F1
    checksum64: 21E9BCA46E9E22E078435F2C463B51A639FC7DC0A8A7A0D1003FD70DE3C25C2E
 
 Using AU:
--- a/automatic/avidemux/tools/chocolateyInstall.ps1
+++ b/automatic/avidemux/tools/chocolateyInstall.ps1
@@ -9,7 +9,6 @@ $silentArgs = @('/S')
 $packageArgs = @{
   packageName    = 'avidemux'
   fileType       = 'exe'
-  file           = "$toolsDir\avidemux_2.7.4_win32.exe"
   file64         = "$toolsDir\Avidemux_2.7.4%20VC%2B%2B%2064bits.exe"
   silentArgs     = $silentArgs
   validExitCodes = @(0, 1223)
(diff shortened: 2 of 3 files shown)


gep13:
@Hattshire thanks for taking the time to do this.  Were you also able to look at the additional requirement regarding the silent arguments required to make the installation work unattended?

Hattshire:
> 
> 
> @Hattshire thanks for taking the time to do this. Were you also able to look at the additional requirement regarding the silent arguments required to make the installation work unattended?

The installer is based on QT installer framework, which only have silent updates, not installations. An automation script can be given apparently.

Hattshire:
nvm, chocolateyInstall.ps1:l16, @AdmiringWorm wrote the solution 9 months ago but used `Get-OSArchitectureWidth`, which is not supported on some PS versions, at least on mine it's not.

Hattshire:
Also note that the PS AppVeyor uses apparently has `Get-OSArchitectureWidth` implemented, see https://ci.appveyor.com/project/chocolateycommunity/chocolatey-coreteampackages/builds/33211071#L257

AdmiringWorm:
@Hattshire `Get-OSArchitectureWidth` is a chocolatey helper, not a built-in PowerShell function.

So please revert back to using that again.

https://chocolatey.org/docs/helpers-get-osarchitecture-width

Hattshire:
Upgrading from 2.7.1 needs prior uninstall, should be called `chocolateyUninstall.ps1` from `chocolateyInstall.ps1`?

majkinetor:
Yes you can do that.

I think your current updater will mix x32 and x64 bit versions. Maybe its better to just use x64 from now on ?

Hattshire:
> I think your current updater will mix x32 and x64 bit versions. 

The latest of the two, as the developer is releasing per-architecture fixes, I think there's a probability that it could happen again. However I believe that the dev will publish both newer versions as `.6` instead of x32 `.5` and x64 `.6`

> Maybe its better to just use x64 from now on ?

I think so, but there's x32 support upstream. Someone using choco will use it? idk

majkinetor:
> The latest of the two, as the developer is releasing per-architecture fixes, I think there's a probability that it could happen again. However I believe that the dev will publish both newer versions as .6 instead of x32 .5 and x64 .6

So what is going to happen when 1 ver has x32 and other x64 ?

Hattshire:
> 
> 
> > The latest of the two, as the developer is releasing per-architecture fixes, I think there's a probability that it could happen again. However I believe that the dev will publish both newer versions as .6 instead of x32 .5 and x64 .6
> 
> So what is going to happen when 1 ver has x32 and other x64 ?

Lets say the latest x32 version is 1.2.3 and the latest x64 is 1.2.4:
 - Package version: 1.2.4 -> CHOCOPKG-v1.2.4.nupkg
 - x32 binary version: 1.2.3 -> Install-x32-v1.2.3.exe
 - x64 binary version: 1.2.4 -> Install-x64-v1.2.4.exe

In the other way around if the latest versions are 5.6.7 for x32 and 5.6.6 for x64:
 - Package version: 5.6.7 -> CHOCOPKG-v5.6.7.nupkg
 - x32 binary version: 5.6.7 -> Setup32-v567.exe
 - x64 binary version: 5.6.6 -> Setup64-v566.exe

That's how I see it, however, I have no problem whatsoever changing it to x64-only releases, it would be easier and cleaner than this sketchy way of providing symmetrical-to-upstream releases.

majkinetor:
Exactly. Its not correct by any definition and road we never took so far (although this is not the first case). I suggest we should go with x64 only version. Its video tool too, it should be used with x64 arch.

Perhaps this deserves an extra note, that since versino x.y.z. package is only x64 bit due to the upstream "problems".

majkinetor:
Thx @Hattshire for making this work again.

PR #1491, opened 2020-06-04
(nomacs)/.portable - use github release page instead of outdated nomacs website

mwallner:
nomacs/nomacs.portable: the previously used page to check for new releases doesn't seem to be updated by the software maintainers any more, directly use the GitHub release page instead.

verified it actually get's the correct version by manually running the iwr & regex - link match in powershell.

### DIFF
--- a/automatic/nomacs/update.ps1
+++ b/automatic/nomacs/update.ps1
@@ -5,9 +5,7 @@ $releases = 'https://github.com/nomacs/nomacs/releases'
 function global:au_SearchReplace {
    @{
         ".\legal\VERIFICATION.txt" = @{
-            "(?i)(\s+x32:).*"            = "`${1} $($Latest.URL32)"
             "(?i)(\s+x64:).*"            = "`${1} $($Latest.URL64)"
-            "(?i)(checksum32:).*"        = "`${1} $($Latest.Checksum32)"
             "(?i)(checksum64:).*"        = "`${1} $($Latest.Checksum64)"
         }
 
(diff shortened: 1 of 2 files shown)


automatic/nomacs/update.ps1
```diff
@@ -21,8 +21,8 @@ function global:au_BeforeUpdate  { Get-RemoteFiles -NoSuffix -Purge }
 
 function global:au_GetLatest {
     $download_page = Invoke-WebRequest -Uri $releases -UseBasicParsing
-    $url     = $download_page.links | ? href -like '*/nomacs-*' | % href | select -First 1
-    $version = $url -split '-|\.zip' | select -Last 1 -Skip 1
+    $url     = $download_page.links | ? href -like '*/nomacs-*x64.msi' | % href | select -First 1
+    $version = $url -split '\/' | select -Last 1 -Skip 1
     @{
         Version      = $version
         URL64        = 'http://download.nomacs.org/nomacs-setup-x64.msi'
```
majkinetor, line 28: Please change URL for install version to be taken from the github releases.

mwallner: fixed, but had to rip out the 32bit url, as it's not present on github releases

AdmiringWorm: That is fine, as I don't think official 32bit editions are even provided by the developers anymore.

However, removing the 32bit URL from `update.ps1` that also means it will need to be removed from the install script (the file argument), same with references to url32 in `VERIFICATION.txt`, could you please do that?
Would be great if you could also add the same 32bit check in the `nomacs` package that is used in the `.portable` package.

mwallner: DOH! thank's for catching this! - see GH-1494

PR #1492, opened 2020-06-08
Fix keeweb

wget:
## Description
Keeweb added an arm 64 version of the exe which was breaking the x64 detection. Fix that.

## Motivation and Context
Take the opportunity to use the GitHub API rather than the HTML parsing.

## How Has this Been Tested?
On local Windows 7 and Windows 10 dev based environment + the test instance of the verifier.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/keeweb/update.ps1
+++ b/automatic/keeweb/update.ps1
@@ -1,80 +1,69 @@
 import-module au
 
-$releases = 'https://github.com/keeweb/keeweb/releases'
-
 function global:au_BeforeUpdate { Get-RemoteFiles -Purge -NoSuffix }
 
 function global:au_SearchReplace {
-    @{
-        ".\legal\verification.txt" = @{
-            "(?i)(32-Bit.+)\<.*\>" = "`${1}<$($Latest.URL32)>"
-            "(?i)(64-Bit.+)\<.*\>" = "`${1}<$($Latest.URL64)>"
-            "(?i)(checksum type:\s+).*" = "`${1}$($Latest.ChecksumType64)"
-            "(?i)(checksum32:\s+).*" = "`${1}$($Latest.Checksum32)"
-            "(?i)(checksum64:\s+).*" = "`${1}$($Latest.Checksum64)"
-        }
-        ".\tools\chocolateyInstall.ps1" = @{
-            "(?i)(^\s*file\s*=\s*`"[$]toolsPath\\).*" = "`${1}$($Latest.FileName32)`""
-            "(?i)(^\s*file64\s*=\s*`"[$]toolsPath\\).*" = "`${1}$($Latest.FileName64)`""
-        }
+  @{
+    ".\legal\verification.txt" = @{
+      "(?i)(32-Bit.+)\<.*\>" = "`${1}<$($Latest.URL32)>"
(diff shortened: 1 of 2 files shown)


automatic/keeweb/update.ps1
```diff
@@ -1,7 +1,5 @@
 import-module au
 
-$releases = 'https://github.com/keeweb/keeweb/releases'
```
AdmiringWorm, line 3: This URL should have be used to replace the URL specified in the `VERIFICATION.txt` file instead of removing it alltogether.
Not sure why that hasn't been done, though.

automatic/keeweb/update.ps1
```diff
@@ -22,48 +20,36 @@ function global:au_SearchReplace {
...
-    if ($url32.Authority -cnotmatch $baseUrl) {
-        $url32 = $scheme + '://' + $baseUrl + $url32
-    }
-    $url32SegmentSize = $([System.Uri]$url32).Segments.Length
-    $filename32 = $([System.Uri]$url32).Segments[$url32SegmentSize - 1]
-
-    $url64 = $downloadedPage.links | ? href -match '64.exe$' | select -First 1 -expand href
-    if ($url64.Authority -cnotmatch $baseUrl) {
-        $url64 = $scheme + '://' + $baseUrl + $url64
+    $jsonAnswer = (Invoke-WebRequest -Uri "https://api.github.com/repos/keeweb/keeweb/releases/latest" -UseBasicParsing).Content | ConvertFrom-Json
```
AdmiringWorm, line 23: As mentioned in the other PR you opened, this will cause issues eventually as it is used without an API token.
Rate limiting will be hit quite often.
Please change to use the web URL, or use it with an API token.

automatic/keeweb/update.ps1
```diff
@@ -22,48 +20,36 @@ function global:au_SearchReplace {
...
-
-        $parsed = $line -split ' |\n' -replace '\*',''
-
-        if ($parsed[1] -cmatch $filename32) {
+        $parsed = $line -Split ' |\n' -Replace '\*',''
+        if ($parsed[1] -Match $filename32) {
             $hash32 = $parsed[0]
         }
-        if ($parsed[1] -cmatch $filename64) {
+        if ($parsed[1] -Match $filename64) {
```
AdmiringWorm, line 48: `cmatch` is used on purpose here, as that is case sensitive match. By default, PowerShell uses case insensitive matching.

wget: Taken care of.

automatic/keeweb/update.ps1
wget: Taken care of.

automatic/keeweb/update.ps1
wget: Since that URL is not used at all, there is no point in replacing the content of VERIFICATION.txt ([e.g. like in Clementine](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/e1ba7e225d190aed9bf0bce035c0b7ab64bba22c/automatic/clementine/update.ps1#L17)) each time. Any way, when that link will change we will have to change it in the nuspec file as well any way, so...

wget:
Same issue as #1493 since we are using the API key.

EDIT: Sorry for having applied the 4 spaces to 2 spaces file change combined in a single commit with the other modification :/

PR #1493, opened 2020-06-08
(keepassxc) Ensure beta versions are not detected by using GitHub API

wget:
## Description
Keepassxc 2.6 beta shouldn't have been pushed. Prevent that and use the GitHub API for proper detection.

## Motivation and Context
See above.

## How Has this Been Tested?
On local Windows 7 and Windows 10 dev based environment.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/keepassxc/keepassxc.nuspec
+++ b/automatic/keepassxc/keepassxc.nuspec
@@ -6,7 +6,7 @@
   <metadata>
     <id>keepassxc</id>
     <title>KeePassXC</title>
-    <version>2.6.0-beta1</version>
+    <version>2.5.4</version>
     <authors>https://github.com/keepassxreboot/keepassxc/graphs/contributors</authors>
     <owners>wget, hans0</owners>
     <summary>KeePassXC is a community fork of KeePassX which aims to incorporate stalled pull requests, features, and bug fixes that have never made it into the main KeePassX repository.</summary>
--- a/automatic/keepassxc/legal/VERIFICATION.txt
+++ b/automatic/keepassxc/legal/VERIFICATION.txt
@@ -6,14 +6,14 @@ The installer have been downloaded from the Github release page <https://github.
 and can be verified like this:
 
 1.  Download the following installers:
-    32-Bit: <https://github.com/keepassxreboot/keepassxc/releases/download/2.6.0-beta1/KeePassXC-2.6.0-beta1-Win32.msi>
-    64-Bit: <https://github.com/keepassxreboot/keepassxc/releases/download/2.6.0-beta1/KeePassXC-2.6.0-beta1-Win64.msi>
+    32-Bit: <https://github.com/keepassxreboot/keepassxc/releases/download/2.5.4/KeePassXC-2.5.4-Win64.msi>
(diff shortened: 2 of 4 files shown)


AdmiringWorm:
Why remove the support for beta versions?
There is no good reason IMO for doing that when it is already being picked up.

Also, since you are using the rest API without an API key, most of the time that this package will be checked for updates, we will hit a rate limit, which is quite common on appveyor.

wget:
> Why remove the support for beta versions?
There is no good reason IMO for doing that when it is already being picked up.

@AdmiringWorm Beta versions are not intended to be distributed. In the past, upstream contacted us to remove a previously published beta version. The latter was even involving data loss and, IIRC, kdbx databases were even becoming corrupted.

So beta versions shouldn't be published. We corrected this issue in the past but we are facing it again due to the way the were parsing it.

> Also, since you are using the rest API without an API key, most of the time that this package will be checked for updates, we will hit a rate limit, which is quite common on appveyor.

Parsing the GitHub html is fragile as detecting the latest release has never been done properly and never will be. The only robust and correct solution is to use the GitHub API.

As for the rate limit, since this repo is a core team packages repo, just add an API key on our AppVeyor instance for us to export in our scripts, done 🤗✅

AdmiringWorm:
> @AdmiringWorm Beta versions are not intended to be distributed. In the past, upstream contacted us to remove a previously published beta version. The latter was even involving data loss and, IIRC, kdbx databases were even becoming corrupted.

Fair enough, although I still don't see a reason to deny them. Especially considering that a pre-release package won't be installed unless the user explicitly decides to do so (and that is always something that should be considered doing on their own risk), no matter though.

> Parsing the GitHub html is fragile as detecting the latest release has never been done properly and never will be. The only robust and correct solution is to use the GitHub API.

So you are saying that the other packages in this repo that uses GitHub as the source of truth is fragile?
It isn't really that much different than parsing from the API, with the exception that things may break if GitHub decides to change the layout, but still, No matter as long as an API key is used.

> As for the rate limit, since this repo is a core team packages repo, just add an API key on our AppVeyor instance for us to export in our scripts, done 🤗✅

There already is an API key on the AppVeyor instance (environment variable: `github_api_key`), this is what is used to update the gist and create releases on the repository.

BTW: You are also getting the following error on the integration tests for this package.
```
Invalid URI: The hostname could not be parsed.
```
https://ci.appveyor.com/project/chocolateycommunity/chocolatey-coreteampackages/builds/33385984#L225

wget:
@AdmiringWorm When you wrote `github_api_key`, are you speaking of an OAuth2 token (sent in a header) or HTTP Basic authentication? (for the latter I'll need the username as well).

AdmiringWorm:
> @AdmiringWorm When you wrote github_api_key, are you speaking of an OAuth2 token (sent in a header) or HTTP Basic authentication? (for the latter I'll need the username as well).

It is a OAuth2 token, or more commonly known in the GitHub settings a `Personal Access Token`.

wget:
@AdmiringWorm Ok. Then we can merge? :D

PR #1472, opened 2020-05-09
(vscode-insiders*) Add vscode-insiders package

felipecassiors:
## Description

This adds a new package to Chocolatey registry, `vscode-insiders`, and was heavily inspired by the `vscode` itself.

I just did all the changes that were required, such as adjusting description, and gathering all the right variables. I also adjusted the AU updater.

## Motivation and Context
The package `vscode-insiders` in Chocolatey is currently very outdated. And since the Insiders version has so many things in common with the plain `vscode`, I decided to bring it here.

## How Has this Been Tested?
I used https://github.com/majkinetor/chocolatey-test-environment and `Test-Package -Vagrant $au_Vagrant` to test many times. Everything worked as expected in all the situations, installing both the meta and Install variant, and also uninstalling.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Migrated package (a package has been migrated from another repository)
- [X] New package?

## Checklist:

- [X] My code follows the code style of this repository.
- [ ] My change requires a change to documentation (this usually means the notes in the description of a package).
- [ ] I have updated the documentation accordingly (this usually means the notes in the description of a package).
- [X] All files are up to date with the latest [Contributing Guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)
- [X] The added/modified package passed install/uninstall in the chocolatey test environment.
- [ ] The changes only affect a single package (not including meta package).

PR #1512, opened 2020-07-28
Add script for testing in Windows Sandbox

felipecassiors:
## Description

This adds a script for testing in Windows Sandbox. Much easier than spinning up a Vagrant machine, this way you can just:

```powershell
.\Scripts\Test-Sandbox.ps1 <nuspec/nupkg>
```

And it will open a Windows Sandbox, install Chocolatey and try to
install the specified nuspec/nupkg with `choco install`.

## Motivation and Context

This is much easier to use and faster than spinning up a virtual machine for simple testing purposes.

Most of the code was taken from
https://github.com/microsoft/winget-pkgs/pull/827.

## How Has this Been Tested?

I'm unable to test it now because of a bug in Windows Sandbox: https://aka.ms/AA928xa.

If someone else not affected by this bug could try, I would appreciate.

## Screenshot

I'll surely post one when I'm able to run it.

## Types of changes

## Checklist:

- [X] My change requires a change to documentation (this usually means the notes in the description of a package).


felipecassiors:
I think I must copy the entire folder to Sandbox when the file is a nuspec.

scripts/Test-Sandbox.ps1
```diff
@@ -0,0 +1,73 @@
...
+
+# Initialize Temp Folder
+
+$tempFolder = Join-Path -Path $PSScriptRoot -ChildPath 'Test-Sandbox_Temp'
+
+New-Item $tempFolder -ItemType Directory -ea 0 | Out-Null
+
+Get-ChildItem $tempFolder -Recurse | Remove-Item -Force
+
+Copy-Item -Path $Nuspec -Destination $tempFolder
```
felipecassiors, line 32: I think I must copy the entire folder if the parameter is a nuspec file. Not needed when it's nupkg.

majkinetor:
I like it.

Couple of comments.

- You should invoke `choco install -y`, because currently you have to interact with prompts.
- It would be nice to have option for pre/post PowerShell scripts, because you might need try automate other aspects of test within VM

majkinetor:
OK, here is my version: 
* https://github.com/majkinetor/postgrest-test/blob/master/Test-Sandbox.ps1

Differences:
1. No longer can you specify nuspec file, just run it from package folder without any params
2. You can use it for other stuff rather then chocolatey, provide `-Script` option
3. You can map array of folders

felipecassiors:
> I like it.
> 
> Couple of comments.
> 
> * You should invoke `choco install -y`, because currently you have to interact with prompts.
> * It would be nice to have option for pre/post PowerShell scripts, because you might need to automate other aspects of test within VM

Great, I will address them. About the second bullet, how about an option to execute nothing, just start the Sandbox, mount the files and install Chocolatey?

Off-topic: can you please reopen https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1472?

majkinetor:
> Great, I will address them. About the second bullet, how about an option to execute nothing, just start the Sandbox, mount the files and install Chocolatey?

I already did all that, so I will add my version into it. 

> Off-topic: can you please reopen #1472?

I can't the open button is grayed due to _"master branch was force pushed or recreated"_. Please make a fresh one using the same branch.

felipecassiors:
> OK, here is my version which is more general
> 
> * https://github.com/majkinetor/postgrest-test/blob/master/Test-Sandbox.ps1
> 
> Differences:
> 
> 1. No longer can you specify nuspec file, just run it from package folder without any params
> 2. You can use it for other stuff rather then chocolatey, provide `-Script` option
> 3. You can map array of folders
> 
> [Sample](https://github.com/majkinetor/postgrest-test/blob/master/.build.ps1#L33-L37)

I completely missed this comment, and I rewrote my script from scratch, basically.

felipecassiors:
Ok, I implemented most of the things you did in yours, except that I didn't use a string of map folders. I think it's a bit overkill for the purpose, and with a single entry, we can `cd` directly to it in the terminal.

felipecassiors:
But so far I was not able to perform a successful local installation of `vscode.install` for instance. It works in my local machine by specifying the `nuspec`, but just because the dependencies are already present.

I already tried in many ways. when not using the `nuspec`, Chocolatey says that no package was found in the specified source. I don't think this is a problem with this script and Sandbox itself, but it would be good to know how to properly use it for testing packages.

tunisiano187:
Really good script,
just FYI, i'm using it that way.

```
.\Test-Sandbox.ps1 { $files = Get-ChildItem ./ToTest/*.nupkg
 foreach ($file in $files) {
  write-host "Testing $(($file.Name).replace('.nupkg',''))"
  if($(choco install -dy $(($file.Name).replace('.nupkg','')) -s "./ToTest/;https://chocolatey.org")) {
  mv $file ./Working/
  } else {
  Write-Debug "$(($file.Name).replace('.nupkg','')) Not working"
  mv $file ./Error/
  }
 } 
}
```

majkinetor:
> But so far I was not able to perform a successful local installation of vscode.install for instance. It works in my local machine by specifying the nuspec, but just because the dependencies are already present.

My version works with this scenario since you can change invocation script and add missing dependencies.

I am considering adding this script in the AU itself in the `Test-Package` function. Then every AU repository there would have it and you could override default script inside `update.ps1` script via something like `au_TestPackage`. 

So don't be surprised if this PR gets stale here , but feel free to offer improvements until I figure it out what I want :).

felipecrs:
> My version works with this scenario since you can change invocation script and add missing dependencies.

Did you check that I ported back your improvements?

tunisiano187:
> However, is there a reason for not merging this PR the way it is currently so people could use it already, and then delete the script when you add it to AU?

I think it's nteresting to keep it separated too.
I'm using it from another perspective and it's interesting.
Thanks

PR #1517, opened 2020-08-04
(vscode-insiders*) Add vscode-insiders* package

felipecassiors:
This is a follow-up of https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1472, because of https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1512#issuecomment-668495125.

## Description

This adds a new package to Chocolatey registry, `vscode-insiders`, and was heavily inspired by the current package `vscode` itself.

I just did all the changes that were required, such as adjusting description, icon, screenshot, and gathering all the right variables. I also adjusted the AU updater.

VSCode Insiders gets new builds every day, so the version which goes to Chocolatey has a forth field with the date of the build, extracted from the timestamp provided by VSCode update's API.

## Motivation and Context
The package `vscode-insiders` in Chocolatey is currently very outdated. And since the Insiders version has so many things in common with the plain `vscode`, I decided to bring it here.

## How Has this Been Tested?
I used https://github.com/majkinetor/chocolatey-test-environment and `Test-Package -Vagrant $au_Vagrant` to test many times. Everything worked as expected in all the situations, installing both the meta and Install variant, and also uninstalling.

## Types of changes

## Checklist:


felipecassiors:
Open questions from https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1472:

## https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1472#issuecomment-664734373:

1. How about we just have one package: `vscode-insiders` and not `vscode-insiders.install`?
   1. If yes, should we deprecate `vscode.install` as well?

1. How about we rename `vscode` to just `code`, and then `vscode-insiders` to `code-insiders` only? This is how it's on Snap Store, and this is the preference of Visual Studio Code team itself. The good is that it matches the CLI executable: `code` and `code-insiders`. See https://github.com/microsoft/winget-pkgs/issues/277.

1. Should I refactor the `vscode` and `vscode.install` in this PR as well?


## https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1472#issuecomment-664767129:

Another issue that I just noticed: builds of VS Code Insiders are published daily, differentiated with the Git SHA. Example:

```bash
{
  "url": "https://az764295.vo.msecnd.net/insider/88e641626546dceb53285aa0ddb44786324cb1bf/VSCodeSetup-ia32-1.48.0- 
  insider.exe",
  "name": "1.48.0-insider",
  "version": "88e641626546dceb53285aa0ddb44786324cb1bf",
  "productVersion": "1.48.0-insider",
  "hash": "671f6ce24cc06fffaafae5934ddf3afa545af21f",
  "timestamp": 1595866312178,
  "sha256hash": "61be694086d2534299450afb3187794d688a510bce9110ac3c310e494c2829a2",
  "supportsFastUpdate": true
}
```

Where `version` is the Git SHA. The question now is, how do we map this in the Chocolatey version? 

The best thing would be if Chocolatey had the support for [NuGet's SemVer 2.0](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#semantic-versioning-200), then we could write something like:

```xml
<version>1.48.0+88e641626546dceb53285aa0ddb44786324cb1bf</version>
```

By the way, are there any plans to support that?

For now, we could do something like:
- Use the `timestamp` provided: `1.48.0.1595866312178`
- Use the current Chocolatey build date: `1.48.0.20200728`
- Use the `timestamp` converted to a date: `1.48.0.20200727`
- Use sequential build numbers generated by Chocolatey: `1.48.0.0`, `1.48.0.1`

Any advice?

felipecassiors:
The CI is fine now, however I see a [weird line](https://ci.appveyor.com/project/chocolateycommunity/chocolatey-coreteampackages/builds/34472311#L227):

```log
[3/3] vscode.install is updated to 1.47.3.20200804 (21.13s)
```

I did not touch the `vscode.install`. Is it normal?

felipecassiors:
Ok, I took the approach to convert the given timestamp to date. It's working and it will reflect the real date of the package.

automatic/vscode-insiders/README.md
```diff
@@ -0,0 +1,30 @@
...
+* `/DontAddToPath` - Don't add Visual Studio Code Insiders to the system PATH.
+
+Example: `choco install vscode-insiders --params "/NoDesktopIcon /DontAddToPath"`
+
+## Notes
+
+* The package uses default install options except that it adds context menu entries and Visual Studio Code Insiders isn't started after installation.
+* For disabling the auto-update functionality see the [Visual Studio Code Auto Update Deactivation package](https://chocolatey.org/packages/visualstudiocode-disableautoupdate).
+
+![screenshot](https://cdn.jsdelivr.net/gh/chocolatey-community/chocolatey-coreteampackages@2467e9b00946347a4b9c2a311bec52540fbb9f99/automatic/vscode-insiders/screenshot.png)
```
AdmiringWorm, line 30: Instead of duplicating the screenshot, you can just use the same URL here as you do in `vscode-insiders.install` instead.

felipecassiors:
> Just one minor thing, before I will approve this.
> If you could also rebase your PR against the latest changes in the repo and fix the commits (adding the package name they are for, ie `(vscode-insiders)` or `(vscode-insiders.install)`), I would appreciate it (otherwise, I can do that for you instead).

One question: how about having just a single version of the package, `vscode-insiders` instead of `vscode-insiders.install`?

felipecassiors:
@AdmiringWorm I squashed all commits. By the way, you could use the merge by squash method by default on PR's that contain commits not compliant with the standard. After all, GitHub does not encourage us to rewrite the Git history.

AdmiringWorm:
@felipecassiors I am aware, however, squashing through GitHub also replaces the actual author of the commit (although reported correctly on the website), and it also rewrites the git history as well so there isn't anything different there.

> One question: how about having just a single version of the package, vscode-insiders instead of vscode-insiders.install?

Nah, I think we should keep it similar to the original vscode packages. This also allows us to eventually provide a .portable version of vscode-insiders if there is a need in the future (which is the reason why there is a `vscode` and `vscode.install` package, since there is also a `vscode.portable` package).


One last thing, @steviecoaster since you are the original maintainer of the `vscode-insiders` package, would you mind taking a look over this PR before we merge it in?

felipecassiors:
Awesome!

Btw, in case you missed, [@steviecoaster commented in the previous PR](https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1472#issuecomment-626213030).

felipecassiors:
I just noticed one problem:

1. Install this package
2. Open Code Insiders
3. Force reinstall

It will fail because `chocolateyBeforeModify.ps1` does not run before a reinstallation.

I'll make it run in `chocolateyInstall.ps1` so. By the way, I suggest considering run `chocolateyBeforeModify.ps1` by default on reinstallations.

AdmiringWorm:
> It will fail because chocolateyBeforeModify.ps1 does not run before a reinstallation.

I would say that is expected, chocolateyBeforeModify only runs when upgrading and uninstalling a package.
Forcing a new installation of a package is generally considered bad practice, so that isn't something I would encourage and someone doing that is doing it on their own risk.
As such, just keep the stopping of the service/process in the beforemodify without moving them.

> I'll make it run in chocolateyInstall.ps1 so. By the way, I suggest considering run chocolateyBeforeModify.ps1 by default on reinstallations.

That would be a feature request for chocolatey itself, and nothing we can do in the package.
So if you really want that, you need to open up an issue over at https://github.com/chocolatey/choco (if one don't already exist).

felipecassiors:
> Forcing a new installation of a package is generally considered bad practice, so that isn't something I would encourage and someone doing that is doing it on their own risk.
> As such, just keep the stopping of the service/process in the beforemodify without moving them.

Too late, sorry. But it's good, makes the package more error-prone. Also helps when the user had the package installed by other means and wanted to reinstall it by using Chocolatey. It won't hurt.

automatic/vscode-insiders.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,38 @@
+$ErrorActionPreference = 'Stop'
+
+# Close Code Insiders before installing
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+. $toolsPath\chocolateyBeforeModify.ps1
```
AdmiringWorm, line 5: Do not source other chocolatey related scripts.
`chocolateyBeforeModify.ps1`, `chocolateyUninstall.ps1` and `chocolateyInstall.ps1` should never be sourced from other scripts, and will be requested to be changed by a moderator.

felipecassiors: Makes sense.

AdmiringWorm: You can rather move the code to a function inside a `helpers.ps1` file (convention in this repo), have that file sourced and called in the uninstall and install script (and remove the before modify script).

Remember additional error handling may be necessary in that case, to ensure the install/uninstall won't fail if a process can't be found.

felipecassiors: It won't be necessary to have additional error handling because the two `Get-Process` calls already have `-ea 0`.

felipecassiors: Ok, got it, you're saying that the code will run twice in the `update`?

felipecassiors: How should I create the `chocolateyUninstall.ps1` just for this call? I'm not sure if creating it will prevent the auto uninstall to run.

It's not exactly necessary, since double checking, if the program is open, won't hurt.

AdmiringWorm: Yup, indeed.

felipecassiors:
Ok, now it should be done for real.

AdmiringWorm:
@felipecassiors your changes have been merged, thanks for your contribution 👍

felipecassiors:
@AdmiringWorm Do you have a clue why the tests are failing for today's version?

https://chocolatey.org/packages/vscode-insiders/1.49.0.20200818

AdmiringWorm:
@felipecrs yes, because the `vscode-insiders.install` package haven't yet been approved (first version always requires a human moderator), thus a new version of that package can't be pushed but one was pushed for the `vscode-insiders` package.

I have temporarily disabled pushing of the `vscode-insiders` package until the `vscode-insiders.install` package is approved.

felipecrs:
I see, thank you so much!

automatic/vscode-insiders.install/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,39 @@
+$ErrorActionPreference = 'Stop'
+
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+. $toolsPath\helpers.ps1
+
+Close-VSCodeInsiders
```
ferventcoder, line 6: 👍

PR #1527, opened 2020-08-23
(Sysinternals) Add start menu entries for GUI apps

cmorty:
## Description
Add start menu entries for the Gui-App of Sysinternals. They are added to a `Sysinternals Suite` folder.
By passing `/NoStartMenu` no entries are created.
They folder is removed upon uninstalling.

## Motivation and Context
This will allow starting the apps from the start menu, as well as ease starting them using Windows-Key+Typing

## How Has this Been Tested?
Installed and uninstalled.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/sysinternals/Readme.md
+++ b/automatic/sysinternals/Readme.md
@@ -9,8 +9,9 @@ It does not contain non-troubleshooting tools like the BSOD Screen Saver or NotM
 
 - `/InstallDir` - Installation directory, by default Chocolatey tools directory.
 - `/InstallationPath` - the same as `InstallDir`
+- `/NoStartMenu` - Set to true to prevent installation of a menu item.
 
-Example: `choco install sysinternals --params "/InstallDir:C:\your\install\path"`
+Example: `choco install sysinternals --params "/InstallDir:C:\your\install\path /NoStartMenu"`
 
 ## Notes
 
@@ -18,4 +19,4 @@ Example: `choco install sysinternals --params "/InstallDir:C:\your\install\path"
 - This package by default installs to tools directory which will create shims for all applications. When you install to different directory, shims are not created but directory is added to the PATH.
 - This package downloads the nano edition of sysinternals suite when installing it on a nano server.
 - To have GUI for the tools, install [nirlauncher](https://chocolatey.org/packages/nirlauncher) package and use `/Sysinternals` package parameter.
-
+- Only tools that have a GUI are added to the start menu.
--- a/automatic/sysinternals/tools/chocolateyInstall.ps1
+++ b/automatic/sysinternals/tools/chocolateyInstall.ps1
@@ -21,6 +21,23 @@ if (Is-NanoServer) {
   $packageArgs.checksum = '7b6fab7f0ada0bc2b0748f3bc29e93846869f9dd175a0ca070bbebc07ee5de09'
  }
 
+if (-Not $pp.NoStartMenu) {
+  Write-Verbose "Creating start menu entry"
+  set-alias 7z $Env:chocolateyInstall\tools\7z.exe
+  $linkfiles = get-childitem "$installDir" -include "*.exe" -recurse
+  $startmenufolder = "$($env:ProgramData)\Microsoft\Windows\Start Menu\Programs\Sysinternals Suite"
+  Add-Content "$toolsPath\chocoUninstall.ps1" -Value "Remove-Item `"$($startmenufolder)`" -Recurse -Force -ea ignore"
+  # This will create a Link for all gui application and add a line chocoUninstall.ps1 to remove it on uninstall
+  foreach ($file in $linkfiles) {
+    $guiinf = 7z l $file
+    # Only link gui applications
+    if( -Not ( $guiinf -match "^Subsystem = Windows GUI" ) ) { continue }
+    Write-Verbose "Creating Start Menu enty for $($file.BaseName)"
+    $newlink ="$($startmenufolder)\$($file.BaseName).lnk"
+    Install-ChocolateyShortcut -shortcutFilePath $newlink -targetPath "$file"
+  }
+}
+
 $old_path = 'c:\sysinternals'
 if ((Test-Path $old_path) -and ($installDir -ne $old_path)) {
     Write-Warning "Clean up older versions of this install at c:\sysinternals"
--- a/automatic/sysinternals/tools/chocolateyUninstall.ps1
+++ b/automatic/sysinternals/tools/chocolateyUninstall.ps1
@@ -1,0 +1,6 @@
+
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+
+if (Test-Path $toolsPath\chocoUninstall.ps1) {
+    . $toolsPath\chocoUninstall.ps1
+}


RedBaron2:
@cmorty 
FYI: Spelling of `creating` is wrong on line 35 of `chocolateyinstall.ps1`

cmorty:
@RedBaron2 Thanks -> Fixed

automatic/sysinternals/tools/chocolateyInstall.ps1
```diff
@@ -21,6 +21,23 @@ if (Is-NanoServer) {
...
 
+if (-Not $pp.NoStartMenu) {
+  Write-Verbose "Creating start menu entry"
+  set-alias 7z $Env:chocolateyInstall\tools\7z.exe
+  $linkfiles = get-childitem "$installDir" -include "*.exe" -recurse
+  $startmenufolder = "$($env:ProgramData)\Microsoft\Windows\Start Menu\Programs\Sysinternals Suite"
+  Add-Content "$toolsPath\chocoUninstall.ps1" -Value "Remove-Item `"$($startmenufolder)`" -Recurse -Force -ea ignore"
+  # This will create a Link for all gui application and add a line chocoUninstall.ps1 to remove it on uninstall
+  foreach ($file in $linkfiles) {
+    $guiinf = 7z l $file
```
AdmiringWorm, line 32: Using `7z` directly is not acceptable.
If needed, you should rather iterate over the already extracted files.

automatic/sysinternals/tools/chocolateyInstall.ps1
```diff
@@ -21,6 +21,23 @@ if (Is-NanoServer) {
   $packageArgs.checksum = '7b6fab7f0ada0bc2b0748f3bc29e93846869f9dd175a0ca070bbebc07ee5de09'
  }
 
+if (-Not $pp.NoStartMenu) {
+  Write-Verbose "Creating start menu entry"
+  set-alias 7z $Env:chocolateyInstall\tools\7z.exe
+  $linkfiles = get-childitem "$installDir" -include "*.exe" -recurse
+  $startmenufolder = "$($env:ProgramData)\Microsoft\Windows\Start Menu\Programs\Sysinternals Suite"
+  Add-Content "$toolsPath\chocoUninstall.ps1" -Value "Remove-Item `"$($startmenufolder)`" -Recurse -Force -ea ignore"
```
AdmiringWorm, line 29: Do not add text directly to the uninstall script, rather add the paths to a separate text file that the uninstall script can read.

automatic/sysinternals/tools/chocolateyInstall.ps1
```diff
@@ -21,6 +21,23 @@ if (Is-NanoServer) {
   $packageArgs.checksum = '7b6fab7f0ada0bc2b0748f3bc29e93846869f9dd175a0ca070bbebc07ee5de09'
  }
 
+if (-Not $pp.NoStartMenu) {
+  Write-Verbose "Creating start menu entry"
+  set-alias 7z $Env:chocolateyInstall\tools\7z.exe
+  $linkfiles = get-childitem "$installDir" -include "*.exe" -recurse
+  $startmenufolder = "$($env:ProgramData)\Microsoft\Windows\Start Menu\Programs\Sysinternals Suite"
```
AdmiringWorm, line 28: Don't resolve the path this way, rather use
```suggestion
  $startmenufolder = Join-Path ([Environment]::GetFolderPath('CommonPrograms')) 'SysInternals Suite'
```

Valid values for `GetFolderPath` can be seen here: https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=netframework-4.0 (only those available in .NET 4.0 can be used)

automatic/sysinternals/tools/chocolateyInstall.ps1
```diff
@@ -21,6 +21,23 @@ if (Is-NanoServer) {
   $packageArgs.checksum = '7b6fab7f0ada0bc2b0748f3bc29e93846869f9dd175a0ca070bbebc07ee5de09'
  }
 
+if (-Not $pp.NoStartMenu) {
```
AdmiringWorm, line 24: Since the package do not already have the ability to create start menu icons, this parameter should be inverted.
So if a user wants the start menu icons, they should opt-in for them.

automatic/sysinternals/tools/chocolateyInstall.ps1
```diff
@@ -21,6 +21,23 @@ if (Is-NanoServer) {
...
+  Write-Verbose "Creating start menu entry"
+  set-alias 7z $Env:chocolateyInstall\tools\7z.exe
+  $linkfiles = get-childitem "$installDir" -include "*.exe" -recurse
+  $startmenufolder = "$($env:ProgramData)\Microsoft\Windows\Start Menu\Programs\Sysinternals Suite"
+  Add-Content "$toolsPath\chocoUninstall.ps1" -Value "Remove-Item `"$($startmenufolder)`" -Recurse -Force -ea ignore"
+  # This will create a Link for all gui application and add a line chocoUninstall.ps1 to remove it on uninstall
+  foreach ($file in $linkfiles) {
+    $guiinf = 7z l $file
+    # Only link gui applications
+    if( -Not ( $guiinf -match "^Subsystem = Windows GUI" ) ) { continue }
```
AdmiringWorm, line 34: You can use
```suggestion
    if( $guiinf -notmatch "^Subsystem = Windows GUI" ) { continue }
```

instead here

automatic/sysinternals/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,6 @@
+
+$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
+
+if (Test-Path $toolsPath\chocoUninstall.ps1) {
+    . $toolsPath\chocoUninstall.ps1
```
AdmiringWorm, line 5: Instead of sourcing a powershell script here, you should rather iterate over a list of paths stored in a text file (as mentioned above).

automatic/sysinternals/Readme.md
```diff
@@ -9,13 +9,14 @@ It does not contain non-troubleshooting tools like the BSOD Screen Saver or NotM
 
 - `/InstallDir` - Installation directory, by default Chocolatey tools directory.
 - `/InstallationPath` - the same as `InstallDir`
+- `/NoStartMenu` - Set to true to prevent installation of a menu item.
```
AdmiringWorm, line 12: This should be inverted to `/StartMenu` (and you don't need to set it to true, just passing that in is enough).

AdmiringWorm (changes requested):

automatic/sysinternals/tools/chocolateyInstall.ps1
cmorty: This detects where it is a GUI or a console application. Adding a start menu entry fro a console app seems unreasonable. Besides hacking directly in the PE, using 7z seems to be the cleanest solution. Parsing the output of shimgen seems worse.....

automatic/sysinternals/tools/chocolateyInstall.ps1
cmorty: I've seen all over the place, but you are right of course.  
Adding an example to https://chocolatey.org/docs/helpers-install-chocolatey-shortcut  and adding a check to the validation could improve the situation. I'd still prefer https://github.com/chocolatey/choco/issues/395 to be fixed, though. :)

automatic/sysinternals/tools/chocolateyInstall.ps1
cmorty: I pondered this a bit. I finally decided to put it this way, because it is a pretty common parameter: https://www.google.com/search?q=site%3Achocolatey.org%2Fpackages%2F+%2FNoStartMenu
I assume somehow using `chocolateyBeforeModify.ps1` to avoid the entries sudden popping up is not an option, right?

automatic/sysinternals/tools/chocolateyInstall.ps1
majkinetor: I don't like this - Calling external app in iteration (7z) for bunch of other apps will slow down installation a lot. 

IMO you should use hardcoded list of GUI apps. The small inconvenience is that new additions would not be visible unless list is changed, but that is really tiny problem given the frequency new GUI tools get published.

majkinetor:
> as well as ease starting them using Windows-Key+Typing

FYI, this already works if you specify some other directory then default, as executables are added on path (that also makes installation finish in few seconds rather then in few minutes that is required to create all shims).

automatic/sysinternals/tools/chocolateyInstall.ps1
RedBaron2: Just my two cents but this method of using 7zip creates dependency on 7zip

FYI:
The use of 7zip in the chocolatey tools is currently temporary and should not be used in any package. This usage by chocolatey could change in the future with our without notice.

AdmiringWorm: > Adding a start menu entry fro a console app seems unreasonable.

I agree with you here, there shouldn't be a need to do this for CLI apps.

> Besides hacking directly in the PE, using 7z seems to be the cleanest solution. Parsing the output of shimgen seems worse.....

I never said you needed to hack directly into the PE, or parsing the output of shimgen.
The files have already been extracted to disk at this point, so iterating over the already extracted files would be better (using `Get-ChilItem` for instance).
You are already just making a decision based on the name of the file, so to the end, it wouldn't be any different than what you already are doing (from a user perspective), other than you don't call 7z directly and don't open/close/reopen the archive several times.

automatic/sysinternals/tools/chocolateyInstall.ps1
AdmiringWorm: True, it is quite common for packages to have the `NoStartMenu` parameter, however, most (if not all) of these packages already added start menu icons (from the installer), or are completely new packages.
This is not a new package, and a start menu icon has not been created before.
So, as to not suddenly surprise users already used to install the package and maybe do not want the start menu icons, or have already their own implementation of creating the icons this is something that will need to be an opt-in, instead of an opt-out.

majkinetor:
Should we also break compatibility here and finally move to ToolsLocation as default ? Creating 200 or so shims is really not nice, for such a frequently updated package.

@AdmiringWorm

AdmiringWorm:
@majkinetor not sure why I completely missed your comment, anyhow.

Generally I would say yes on moving it to the `Get-ToolsLocation`, unfortunately I know making that change will most likely end up with many users (relatively) sending emails complaining the executable no longer working after installation (due to them not running `refreshenv`).
And I honestly do not want to deal with those emails.

pauby:
This has been open for 6 months with no updates. Any reason not to close it?

PR #1531, opened 2020-08-31
upgrade supertuxkart to version 1.2

pataluc:
package supertuxkart

## Description
upgrade to version 1.2


AdmiringWorm (changes requested):
This is an automatic package, as such we do not accept pull requests only changing the URLs and checksums in the install script.
Please make any necessary changes to the `update.ps1` file which is responsible for updating and replacing the arguments instead.

pataluc:
> This is an automatic package, as such we do not accept pull requests only changing the URLs and checksums in the install script.
> Please make any necessary changes to the `update.ps1` file which is responsible for updating and replacing the arguments instead.

thanks for the info, I admit that I did not try to learn how to update these packages. :)

AdmiringWorm:
@pataluc Can you open an issue for supertuxkart being outdated, this is so it won't be forgotten.
I hope to have the chance to look into getting it updated later this week.

Issue #1528, opened 2020-08-24
(graphviz) Outdated


majkinetor:
Graphviz iz outdated due to upstream changes. It requires as far as I can see full rewrite which at the end should invoke `dot -c` in admin console.

- Binaries: https://www2.graphviz.org/Packages/stable/windows/
- Changelog: https://gitlab.com/graphviz/graphviz/-/blob/master/CHANGELOG.md#2441-2020-06-29

It now provides both x32 and x64 cmake builds:

- https://forum.graphviz.org/t/new-simplified-installation-procedure-on-windows/224

Smattr:
FYI the Graphviz maintainers are currently trying to assess the delta in functionality between the 2.38 release and 2.44.1: https://forum.graphviz.org/t/windows-x86-x86-64-builds/305/6

RedBaron2:
@Smattr 
Are you saying for someone to go ahead and work on fixing the update script? The forum thread doesn't make it clear.

Smattr:
I think there's no reason to hold off on updating the Chocolatey script if you have time to work on this. However, be aware there has been lots of turbulence in the Graphviz world between the 2.38 release and the 2.44.1 release. There are likely binaries that ship in the 2.38 release that are missing from the 2.44.1 release. Users who are upgrading may find this surprising.

If you post a PR for this and tag me in it, I can review the changes from a Graphviz perspective.

Smattr:
One other thing, we've had users reporting the 2.44.1 installer is detected by several AV programs. The general thinking is that this is a false positive, but we have limited ability to evaluate these results or influence them: https://gitlab.com/graphviz/graphviz/-/issues/1773.

RedBaron2:
@Smattr 
I will work on updating the update script, but the install/zip files being flagged by VirusTotal is disturbing.
This might not cause issue for FOSS users of chocolatey, but paid users of chocolatey could be denied update/install of the program.
Please keep us updated on the [Issue #1773](https://gitlab.com/graphviz/graphviz/-/issues/1773)

AdmiringWorm:
@RedBaron2 I agree, the amount of anti-virus scanners getting triggered on VirusTotal is disturbing.
And for the chocolatey repository it is way to high (12/53 for the zip archive, and 22/70 for the executable) without a sufficient reason for the flagging (just saying false flags would not be a sufficient reason in this case).

[...]

PR #1535, opened 2020-09-08
(graphviz) new updater for cmake/msbuild & stable/development

RedBaron2:
@AdmiringWorm @gep13 @pascalberger @majkinetor 
/cc @smattr @magjac 
## Description

This will introduce the new package of `graphviz-cmake` along with `--pre` builds for the `msbuild` & `cmake`
This is still an embedded package.

## Motivation and Context

This will fix issue #1528 plus add new package `graphviz-cmake` 

## Types of changes

## Checklist:

## Original Location
  - [x] Issue 1 #1528 
  - [ ] Issue 2 Link


Smattr:
Thanks for updating this, @RedBaron2. Some initial high level comments:

1. It's great that the whitespace changes are gone. Thanks for addressing this. However, this PR now doesn't touch the license files at all. Is there an established Chocolatey procedure for what to do when the upstream license changes between releases?
2. The PR contains a number of .tmp files that get renamed/copied to filenames with other extensions. Is there a reason these don't have the destination extension to begin with? Maybe this is a Chocolatey convention. My level of understanding of Chocolatey is very minimal.
3. The verification files now have no files or checksums listed. Is this the norm? It seems unclear what a user would do with these files or what they would learn by reading them.

RedBaron2:
@Smattr 
I'll try to address your concerns in order
1. The license files are the same ones used currently in the repository. Since these are unchanged. I saw no need to replace/reupload them with my PR.

2. The`.tmp` files are renamed because they are template files. These are only used by AU to make the files required to make a proper chocolatey package. These are not included in the final packaged `.nupkg` file.

3. The blanked out verification files will be filled with the proper information when ran by AU update. Please, take a look at the appveyor run and download the `.nupkg` file to find out how it looks when finished. The `.nupkg` files are nothing more the a fancy named `.zip` file.

Smattr:
Thanks, @RedBaron2. I think that explains (2) and (3) to me. When you say "currently in the repository," you mean this repository? My point was that automatic/graphviz/legal/LICENSE.txt does not match [the upstream Graphviz license](https://gitlab.com/graphviz/graphviz/-/blob/master/LICENSE). Surely these are intended to agree?

majkinetor:
> Surely these are intended to agree?

Yes, if so, the latest license should be added in the package.

RedBaron2:
@Smattr @majkinetor 
The latest License file is now uploaded and part of this PR

RedBaron2:
@majkinetor @AdmiringWorm 
Just a reminder that this needs to be checked. 
Thanks for the assistance.

majkinetor:
I believe you need to call `dot -c` after installation to generate config file.

magjac:
> I believe you need to call dot -c after installation to generate config file.

Correct, although in next release we will remove this need.

majkinetor:
OK, then no change is needed.

majkinetor:
Why would users care for cmake vs msbuild ? Users want tool that works, not tool built in particular engine. I think cmake should become next version and msbuild should be abandoned. Furthermore, this separation makes package code needlessly complex.

magjac:
> Why would users care for cmake vs msbuild ?

Historically, MSBuild was used to build Graphviz for Windows, but a couple of years ago it was decided that there were too many problems with it and a build system based on CMake was developed. The CMake build system produces both 32-bit and 64-bit binaries as opposed to the MSBuild build system that only produces 32-bit binaries. OTOH, the CMake build system does not yet produce all tools, only the most essential ones. We plan to improve this in future.

If you *have* too choose I would recommend you to use only the CMake builds. For the vast majority of users, they are probably sufficient.

More info:
* https://forum.graphviz.org/t/new-simplified-installation-procedure-on-windows/224
* https://forum.graphviz.org/t/the-story-of-windows-releases-since-2-38/236
* https://gitlab.com/graphviz/graphviz/-/issues/1753 and other issues detailing differences between the builds.
* https://gitlab.com/graphviz/graphviz/-/issues/1218 for the reasoning behind developing a build system based on CMake

majkinetor:
> If you have too choose

We don't have to do anything.  We are just talking and planning for the sake of maintenance and better end result.

The graphviz 2.38 is embedded and will work forever. People can install it using `cinst graphviz --version 2.38` if they need all the tools missing in newer versions for now.

Msbuild seems to be backward compatibility moment, and it looks its only there as there is no feature parity yet, which will be resolved in the future, and thus msbuild will certainly be removed fully. Furthermore, cmake version has x64 bit build.

Thats why I think we should continue with cmake version only. This decision, along with what it means should be explained in package notes.

RedBaron2:
@majkinetor 
Going forward with on the cmake only version will make the current version depreciated.
If this is the decision of the _Chocolatey-CoreTeam_  I have now problems with it. 

> This decision, along with what it means should be explained in package notes.

I will gladly update this PR with the required text, but I don't want to put in "a wall of text" for the community to read. Please mention what changes are needed text for the readme.

**Conflicting files**
_automatic/graphviz/Graphviz.nuspec_
 IMHO This should never have been in the package from the start, and should not be a point to hold this PR.

majkinetor:
> I will gladly update this PR with the required text, but I don't want to put in "a wall of text" for the community to read. Please mention what changes are needed text for the readme.

- Starting from version X.Y.Z. this package contains cmake build of the graphwiz. If you encounter problem with missing executables for the time being continue using the last msbuild version M.N.K.

> Going forward with on the cmake only version will make the current version depreciated.

Not sure what you mean here - we don't need any package deprecation.

RedBaron2:
> * Starting from version X.Y.Z. this package contains new cmake build of the graphviz. If you encounter problem with missing executables for the time being continue using the last msbuild version M.N.K.

I presume from this Note that you want to make the newest release of the graphviz package be the cmake built version.

Might I propose the following as an Example of what will happen when this is done. 
Community user has been using the graphviz version previously. Now they use the upgrade command  for chocolatey upgrading graphviz. This will install the 64-bit version of the cmake built graphviz program. All plugins/settings are not preserved when the user runs the program again. User will post to the Disqus or open a GitHub Issue here complaining about plugin/settings not being preserved when upgrading to the latest version.

If I'm wrong in your intention @majkinetor please let me know, but a lot of confusion is going to be caused by this build change with reusing the graphviz package name.

majkinetor:
And such is the price of progress. 

Anytime people update anything, there is a risk. 

Also, we prioritize maintenance here, given that we maintain 300+ packages for free.

majkinetor:
There is also a contra argument. Majority of tools that depend on graphviz will, if we create new package, continue to use the old one, which is not maintained any more. We also create confusion with multiple package names. The current one is essentially 5 years old.

RedBaron2:
I agree this is the logical progression, but I just wanted it noted in writing that this was discussed, and this will be the out come for the discussion.

@majkinetor 
Do you want me to adjust the update to **not** do the `msbuild` updates?

Smattr:
I don't have a good solution to the above, but I just wanted to acknowledge the upstream problems that led to this. The asymmetry between the CMake and MSBuild artifacts and the fact that neither is a superset has plagued Graphviz for sometime. We're working on fixing it, but unfortunately Graphviz is an under-resourced project, with @magjac being the only brave Windows maintainer.

magneticnorth:
Does anyone know how to write post install actions for Windows?  If so, then hard linking the layout programs could solve a large fraction of the problems.  (There is just one layout program, but it checks argv[0] to select a particular layout algorithm. "For historical reasons".  This means dot, neato, sfdp, fdp, circo, twopi, osage, patchwork.) Someone would have to actually try this out, too.  I think in the past we may have brute forced this by just making N copies of the same executable.

magjac:
@magneticnorth Yes, I think I know how to do that. It's on my TODO, but so far there's been more important stuff to do since users have an simple workaround by doing e.g. `dot -Kneato ...` instead of `neato ...`.

For the record; these are the tools that are missing from the Windows CMake builds:

* circo
* cluster
* diffimg
* dot2gxl
* dot_builtins
* dotty
* edgepaint
* fdp
* gc
* gv2gml
* gv2gxl
* gvedit
* gvmap
* gvmap.sh
* gvpr
* gxl2dot
* lefty
* lneato
* mingle
* neato
* osage
* patchwork
* prune
* sfdp
* smyrna
* twopi
* vimdot

These are the tools that are missing from the Windows MSBuild builds (a strict subset of the missing from the CMake builds):

* circo
* cluster
* dot2gxl
* dot_builtins
* fdp
* gv2gxl
* gvedit
* gvmap.sh
* gxl2dot
* neato
* osage
* patchwork
* sfdp
* twopi
* vimdot

majkinetor:
So this is even more complicated :)

@magjac , does `dot` depends on any of those missing tools since it is used mostly ?

> since users have an simple workaround by doing e.g. dot -Kneato ... instead of neato ....

Wait a minute
Tools are not there but they can actually be used via `dot -K` ? 

If so, we can actually create shims during installation to produce missing exes ! Shim is tiny wrapper around executable and one can also hardcode some params into it (i.e. shim `neato.exe` would be wrapping `dot.exe -kneato`).

magjac:
> @magjac , does dot depends on any of those missing tools since it is used mostly ?

No.

> Tools are not there but they can actually be used via dot -K ?

Not in general, but some of the tools are just shortcuts for running `dot -K`. These tools are:
* circo
* fdp
* neato
* osage
* patchwork
* sfdp
* twopi

>If so, we can actually create shims during installation to produce missing exes ! Shim is tiny wrapper around executable and one can also hardcode some params into it (i.e. shim neato.exe would be wrapping dot.exe -kneato). This would ofc. be specific to choco install (although you can trivially recreate such thing).

You *could* do this for the shortcut tools mentioned above, but it wouldn't add any functionality that is not already available in `dot`.

majkinetor:
> You could do this for the shortcut tools mentioned above, but it wouldn't add any functionality that is not already available in dot

The only benefit would be if somebody relied on those tools on the PATH. Although, I honestly think people should fix their scripts then rather then us doing all kind of stuff.

magjac:
> The only benefit would be if somebody relied on those tools on the PATH. Although, I honestly think people should fix their scripts, rather then us doing all kind of weird stuff.

Agreed and since the Graphviz docs documents that these tools exist, it's *us* (the Graphviz maintainers) that should implemented it in the Graphviz releases. Which we will in the semi-near future.

majkinetor:
OK, guys, thx for discussion. I would proceed with my original intent:

1. Reuse current package name
2. Use cmake build
3. Add note about it, and even link to this discussion, which we will remove once parity with old version exists.

If anyone has any objection, voice it here.

RedBaron2:
@majkinetor 
I will work on updating the PR next week

RedBaron2:
@majkinetor 
The files are updated to just process the cmake build as per the reviews. The Uppercase conflict of the nuspec file is blocking the appveyor builds.
Enjoy the weekend 👍

majkinetor:
Looks legit. Does it add it to the PATH since shims are removed ?

RedBaron2:
> Looks legit. Does it add it to the PATH since shims are removed ?
This would be a good question for @Smattr or @magjac to possibly answer.

> Also please resolve conflicts.
@majkinetor @gep13 
It has been mentioned earlier in this WIP PR about the non Linux conforming `nuspec` file. All the other packages in this repository follow the rules of lowercase names. This shouldn't be a deal breaker for this PR.

magjac:
>    Looks legit. Does it add it to the PATH since shims are removed ?
>> This would be a good question for @Smattr or @magjac to possibly answer.

The CMake installer can add to the user's or the system's PATH, but it's not the default. The user has to explicitly select it by radio buttons. How this works when running from the CLI I don't know and I cannot tell how the installer is invoked from the chocolatey installer.

majkinetor:
@RedBaron2 dot has to be on the PATH one way or another. Please check.

RedBaron2:
> dot has to be on the PATH one way or another. Please check.

I'll check this week.

RedBaron2:
@majkinetor 
The `dot` file will now be on the path, like in the original package.

@magjac 
Just FYI:
The basics of the chocolatey call to install the software is similar to if the user typed
`C:\Users\Owner\Downloads\graphviz-install-2.44.2~dev.20201020.0210-win64.exe /S`
in a  (Administrator)  command line or powershell prompt.

majkinetor:
Please rebase and lets merge this.

RedBaron2:
@majkinetor 
I'll see what I can do about a re-base, but I will make no promises about this working out well. 
I might just close this PR and submit a re-freshed PR for merge Tuesday or  Wednesday.

RedBaron2:
Closing this PR to submit #1571

majkinetor:
FYI, I think this one behaves correctly now. I fixed problems with path and added more detailed note:

https://chocolatey.org/packages/Graphviz/2.44.1.20201124

RedBaron2:
@majkinetor 
Thanks, I appreciate your help. I will probably work on making a function helper for the un/install of the bin for the Chocolatey Core extensions on the future.

PR #1555, opened 2020-10-02
(mpc-hc-clsid2) AU Fix for Stable Releases

TreyRuffy:
## Description
I have made it so that MPC-HC-CLSID2 will let the AU get the latest stable releases without outputting an error.

## Motivation and Context
This fixes the AU giving an error every time it runs because MPC-HC-CLSID2 does not have conventional update names.

## How Has This Been Tested?
I have run it locally along with a private PR to ensure the bot will run correctly. They both were successful.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/mpc-hc-clsid2/update.ps1
+++ b/automatic/mpc-hc-clsid2/update.ps1
@@ -27,8 +27,7 @@ function global:au_GetLatest {
   $re = 'x86\.exe$'
   $url32 = $download_page.Links | ? href -match $re | select -first 1 -expand href | % { 'https://github.com' + $_ }
 
-  $re = 'x64\.exe$'
-  $url64 = $download_page.links | ? href -match $re | select -first 1 -expand href | % { 'https://github.com' + $_ }
+  $url64 = $url32 -replace "x86","x64"
 
   $verRe = 'MPC-HC\.|\.x(86|64)'
   $version32 = $url32 -split "$verRe" | select -first 1 -skip 1
--- a/automatic/mpc-hc-clsid2/update.ps1
+++ b/automatic/mpc-hc-clsid2/update.ps1
@@ -29,7 +29,7 @@ function global:au_GetLatest {
   $versionHtml = $releasesPage.Links | ? title -match $re | select -first 1
   $version = $versionHtml.title
 
-  $download_page = Invoke-WebRequest -Uri "$($releases)/tag/$($version)"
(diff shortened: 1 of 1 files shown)


automatic/mpc-hc-clsid2/update.ps1
```diff
@@ -27,8 +27,7 @@ function global:au_GetLatest {
   $re = 'x86\.exe$'
   $url32 = $download_page.Links | ? href -match $re | select -first 1 -expand href | % { 'https://github.com' + $_ }
 
-  $re = 'x64\.exe$'
-  $url64 = $download_page.links | ? href -match $re | select -first 1 -expand href | % { 'https://github.com' + $_ }
+  $url64 = $url32 -replace "x86","x64"
```
AdmiringWorm, line 30: Do not rely on blindly replacing `x86` with `x64`.
There have been times in the past where the naming haven't been exact.

TBH, this change looks entirely unnecessary.
Please explain why this change is necessary.

AdmiringWorm (changes requested):

automatic/mpc-hc-clsid2/update.ps1
AdmiringWorm: I see know why you made that change.
However, my first two lines still stands as the name isn't guaranteed to be the same.

Please only update the regex to also check the version instead.

_Remember to always update the commit message as mentioned in your other message_

automatic/mpc-hc-clsid2/update.ps1
```diff
@@ -22,24 +22,24 @@ function global:au_SearchReplace {
...
+  $releasesPage = Invoke-WebRequest -Uri $releases -UseBasicParsing
 
-  $re = 'x86\.exe$'
-  $url32 = $download_page.Links | ? href -match $re | select -first 1 -expand href | % { 'https://github.com' + $_ }
+  $re = "^((?:\d+\.){2}\d+)$"
 
+  $versionHtml = $releasesPage.Links | ? title -match $re | select -first 1
+  $version = $versionHtml.title
+
+  $download_page = Invoke-WebRequest -Uri "$($releases)/tag/$($version)"
```
RedBaron2, line 32: The time out is being caused by the lack of using the `-UseBasicParsing` switch.

TreyRuffy: Thank you. I just pushed a commit to add that.

PR #1640, opened 2021-04-12
(mkdocs) Find correct python executable

naveen521kk:
## Description
Find the correct python executable by querying the registry.
Currently, the package just chooses the first `python.exe` which it seems. The `python.exe` could be from a venv(virtual environment) and could be problematic. It should query registry as specified in https://www.python.org/dev/peps/pep-0514 and then find the `python.exe` which is originally installed from an Installer and use the `python.exe` from there.

## Motivation and Context
Fixes https://github.com/chocolatey-community/chocolatey-coreteampackages/issues/1623
Prevent installing mkdocs on a Virtual environment(venv).

## Types of changes

## Checklist:

- [ ] All files are up to date with the latest [Contributing Guidelines](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)

### DIFF
--- a/automatic/mkdocs/tools/ChocolateyInstall.ps1
+++ b/automatic/mkdocs/tools/ChocolateyInstall.ps1
@@ -13,10 +13,10 @@ function FindPython {
     $allowed_python_versions
   )
   # see https://www.python.org/dev/peps/pep-0514/#structure
-  # we are querying machine regsitry only because chocolatey
+  # we are querying machine registry only because chocolatey
   # installs python in admin mode only.
-  $avaiable_installation = Get-ChildItem -Path Registry::HKEY_LOCAL_MACHINE\Software\Python\PythonCore | Select-Object Name
-  foreach ($install in $avaiable_installation) {
+  $available_installation = Get-ChildItem -Path HKLM:\Software\Python\PythonCore | Select-Object Name
+  foreach ($install in $available_installation) {
     $name_install = $install.Name
     $install_version = ($name_install -split '\\')[-1]
     if ($allowed_python_versions.Contains($install_version)) {
--- a/automatic/mkdocs/tools/ChocolateyInstall.ps1
+++ b/automatic/mkdocs/tools/ChocolateyInstall.ps1
@@ -1,5 +1,5 @@
 ﻿Update-SessionEnvironment
-
+$allowed_python_versions = @('3.9', '3.8', '3.7', '3.6', '3.5') # sync with nuspec
 $version = '1.1.2'
 
 $proxy = Get-EffectiveProxy
@@ -7,4 +7,28 @@ if ($proxy) {
   Write-Host "Setting CLI proxy: $proxy"
   $env:http_proxy = $env:https_proxy = $proxy
 }
-python -m pip install mkdocs==$version
+
+function FindPython {
+  param(
+    $allowed_python_versions
+  )
+  # see https://www.python.org/dev/peps/pep-0514/#structure
+  # we are querying machine regsitry only because chocolatey
+  # installs python in admin mode only.
+  $avaiable_installation = Get-ChildItem -Path Registry::HKEY_LOCAL_MACHINE\Software\Python\PythonCore | Select-Object Name
+  foreach ($install in $avaiable_installation) {
+    $name_install = $install.Name
+    $install_version = ($name_install -split '\\')[-1]
+    if ($allowed_python_versions.Contains($install_version)) {
+      Write-Host "Found Python Version from Registry $install_version" -ForegroundColor Yellow
+      $python_executable = Get-ItemProperty -Path "Registry::$name_install\InstallPath" | Select-Object ExecutablePath
+      Write-Host "Found Install Path from Registry $($python_executable.ExecutablePath)" -ForegroundColor Yellow
+      return $python_executable.ExecutablePath
+    }
+  }
+}
+
+$python = FindPython $allowed_python_versions
+Write-Host "Found python at '$python'. Using it."
+& "$python" -m ensurepip # make sure pip exists
+& "$python" -m pip install mkdocs==$version
--- a/automatic/mkdocs/update.ps1
+++ b/automatic/mkdocs/update.ps1
@@ -5,9 +5,9 @@ $releases = 'https://pypi.python.org/pypi/mkdocs'
 function global:au_SearchReplace {
     @{
         'tools\ChocolateyInstall.ps1' = @{
-            "(^[$]version\s*=\s*)('.*')" = "`$1'$($Latest.Version)'"
+            "(^[$]version\s*=\s*)('.*')" = "`$1'$($Latest.PyPIVersion)'"
         }
-     }
+    }
 }
 
 function global:au_GetLatest {
@@ -17,7 +17,7 @@ function global:au_GetLatest {
     $url = $download_page.links | ? href -match $re | select -first 1 -expand href
     $version = $url -split '\/' | select -last 1 -skip 1
 
-    return @{ Version = $version }
+    return @{ Version = $version; PyPIVersion = $version }
 }
 
 update -ChecksumFor none


automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
-python -m pip install mkdocs==$version
+
+function FindPython {
+  param(
+    $allowed_python_versions
+  )
+  # see https://www.python.org/dev/peps/pep-0514/#structure
+  # we are querying machine regsitry only because chocolatey
+  # installs python in admin mode only.
+  $avaiable_installation = Get-ChildItem -Path Registry::HKEY_LOCAL_MACHINE\Software\Python\PythonCore | Select-Object Name
```
pauby, line 18: The variable `$avaiable_installation` is misspelt for the first part.

You don't need to use the `Registry::HKEY_LOCAL_MACHINE` - just use `HKLM:`

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
+  foreach ($install in $avaiable_installation) {
+    $name_install = $install.Name
+    $install_version = ($name_install -split '\\')[-1]
+    if ($allowed_python_versions.Contains($install_version)) {
+      Write-Host "Found Python Version from Registry $install_version" -ForegroundColor Yellow
+      $python_executable = Get-ItemProperty -Path "Registry::$name_install\InstallPath" | Select-Object ExecutablePath
+      Write-Host "Found Install Path from Registry $($python_executable.ExecutablePath)" -ForegroundColor Yellow
+      return $python_executable.ExecutablePath
+    }
+  }
```
pauby, line 28: This code will return the first Python executable it finds in the `allowed_python_versions`. If there are multiple versions of Python installed it may not pick the latest?

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
+
+function FindPython {
+  param(
+    $allowed_python_versions
+  )
+  # see https://www.python.org/dev/peps/pep-0514/#structure
+  # we are querying machine regsitry only because chocolatey
+  # installs python in admin mode only.
+  $avaiable_installation = Get-ChildItem -Path Registry::HKEY_LOCAL_MACHINE\Software\Python\PythonCore | Select-Object Name
+  foreach ($install in $avaiable_installation) {
```
pauby, line 19: You have the misspelt variable name here again.

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
+    $name_install = $install.Name
+    $install_version = ($name_install -split '\\')[-1]
```
pauby, line 21: When you query the registry above, the object returned includes a Version and SysVersion property that you can use rather than splitting the string.

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
+    if ($allowed_python_versions.Contains($install_version)) {
+      Write-Host "Found Python Version from Registry $install_version" -ForegroundColor Yellow
+      $python_executable = Get-ItemProperty -Path "Registry::$name_install\InstallPath" | Select-Object ExecutablePath
```
pauby, line 24: Again you don't need to use `Registry::` here.

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
+      Write-Host "Found Install Path from Registry $($python_executable.ExecutablePath)" -ForegroundColor Yellow
```
pauby, line 25: You've already set the `$python_executable` variable above to point to ExecutablePath so this code won't work.

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
 $proxy = Get-EffectiveProxy
 if ($proxy) {
   Write-Host "Setting CLI proxy: $proxy"
   $env:http_proxy = $env:https_proxy = $proxy
 }
-python -m pip install mkdocs==$version
+
+function FindPython {
+  param(
+    $allowed_python_versions
```
pauby, line 13: If you're going to use an Advanced Function, can you use `[Parameter()]`, and include the variable type?

automatic/mkdocs/tools/ChocolateyInstall.ps1
```diff
@@ -1,10 +1,34 @@
...
+      Write-Host "Found Install Path from Registry $($python_executable.ExecutablePath)" -ForegroundColor Yellow
+      return $python_executable.ExecutablePath
+    }
+  }
+}
+
+$python = FindPython $allowed_python_versions
+Write-Host "Found python at '$python'. Using it."
+& "$python" -m ensurepip # make sure pip exists
+& "$python" -m pip install mkdocs==$version
```
pauby, line 34: What happens here if `$python` is not what you expect it to be?

pauby (changes requested):
A few changes to make. Can I ask you to test the code by creating a package and installing / uninstalling it before submitting the changes and confirm you've done that?

naveen521kk:
> Can I ask you to test the code by creating a package and installing / uninstalling it before submitting the changes and confirm you've done that?

No, Sorry, I haven't.  I didn't find time for that, I fix the update script so that it install correctly on Appveyor. Should that be fine right?

Also, I have used the particular code on another package which I maintain where it installes correctly, see [manimce](https://chocolatey.org/packages/manimce).

automatic/mkdocs/tools/ChocolateyInstall.ps1
naveen521kk: Yeah, it doesn't pick the latest. I am thinking about reversing the array so that the latest version is picked.
```ps1
if ($avaiable_installation.length -gt 1){
    [array]::Reverse($avaiable_installation)
}
```
is this fine?

automatic/mkdocs/tools/ChocolateyInstall.ps1
naveen521kk: I could see it is working https://ci.appveyor.com/project/chocolateycommunity/chocolatey-coreteampackages/builds/38750213#L408? Need to check what happens.

automatic/mkdocs/tools/ChocolateyInstall.ps1
naveen521kk: `$python` is from the function above, so it should contain the full path `python.exe`. If python isn't found then this will error out.

naveen521kk: Something like https://ci.appveyor.com/project/chocolateycommunity/chocolatey-coreteampackages/builds/38750305#L413 will happen

automatic/mkdocs/tools/ChocolateyInstall.ps1
naveen521kk: Removing `Registry::` errored out.
The `$name_install` contains values like `HKEY_LOCAL_MACHINE\Software\Python\PythonCore\3.7` and using `Registry::` seems to work.

RedBaron2:
@naveen521kk @pauby 
I'm going to suggest that this not be tested on appveyor. Python, python 3.x, and python 2.x are all installed in the image beforehand. This makes testing for install detection via chocolatey more complex than is absolutely necessary.

My suggestion is to make this PR a helper that will detect the program searched in the registry. This will make it usable to other packages rather than just one package.

naveen521kk:
>  I'm going to suggest that this not be tested on appveyor. Python, python 3.x, and python 2.x are all installed in the image beforehand. This makes testing for install detection via chocolatey more complex than is absolutely necessary.

It will still work though.

> My suggestion is to make this PR a helper that will detect the program searched in the registry. This will make it usable to other packages rather than just one package.

@RedBaron2 are you suggesting something like https://github.com/chocolatey-community/chocolatey-coreteampackages/issues/1579, in that case, I am not going to do it.

RedBaron2:
> >  I'm going to suggest that this not be tested on appveyor. Python, python 3.x, and python 2.x are all installed in the image beforehand. This makes testing for install detection via chocolatey more complex than is absolutely necessary.
> 
> It will still work though.

 Glad to hear it will still work


> > My suggestion is to make this PR a helper ...
> 
> @RedBaron2 are you suggesting something like https://github.com/chocolatey-community/chocolatey-coreteampackages/issues/1579, in that case, I am not going to do it.

No, I was thinking more like a helper function. I wasn't thinking of a anything like that issue that only helps one package.

Don't mind me. You seem to have it under control.

PR #1658, opened 2021-05-15
(vcredist2008, vcredist2010) Update metadata and prevent double install of 32-bit version

jberezanski:
## Description

These are similar changes as requested for vcredist2005 in https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1657.

Additionally, vcredist2008 is held in moderation due to invalid bugTrackerUrl.

## How Has this Been Tested?
Manually, in a Windows Server 2019 VM.

## Types of changes

## Checklist:

- [ ] All files are up to date with the latest [Contributing Guidelines](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)

### DIFF
--- a/manual/vcredist2008/tools/chocolateyInstall.ps1
+++ b/manual/vcredist2008/tools/chocolateyInstall.ps1
@@ -14,7 +14,7 @@ Install-ChocolateyPackage @params
 
 # Install both 32bit and 64bit on a 64bit OS
 # If a program is compiled as x86 and the 32bit version of vcredist isn't installed, then the program would fail to start.
-if (Get-ProcessorBits 64) {
+if (Get-ProcessorBits 64 -and ($env:chocolateyForceX86 -ne $true)) {
   $originalChocolateyForceX86 = $Env:chocolateyForceX86
   $Env:chocolateyForceX86 = $true
   Install-ChocolateyPackage @params
--- a/manual/vcredist2008/vcredist2008.nuspec
+++ b/manual/vcredist2008/vcredist2008.nuspec
@@ -1,32 +1,29 @@
 <?xml version="1.0"?>
 <!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
-
 <!-- IconUrl: Skip check -->
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
-    <id>vcredist2008</id>
     <version>0.0</version>
-    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/manual/vcredist2008</packageSourceUrl>
-    <owners>chocolatey-community, dtgm</owners>
-    <title>Microsoft Visual C++ 2008 SP1 Redistributable Package</title>
+    <packageSourceUrl>https://github.com/chocolatey/chocolatey-coreteampackages/tree/master/manual/vcredist2010</packageSourceUrl>
     <authors>Microsoft</authors>
-    <projectUrl>https://www.microsoft.com/download/details.aspx?id=26368</projectUrl>
-    <copyright>https://www.microsoft.com/en-us/legal/intellectualproperty/copyright</copyright>
+    <owners>chocolatey-community, dtgm</owners>
     <licenseUrl>http://msdn.microsoft.com/en-US/cc300389.aspx</licenseUrl>
+    <projectUrl>https://www.microsoft.com/en-us/download/details.aspx?id=26368</projectUrl>
+    <id>vcredist2008</id>
+    <title>Microsoft Visual C++ 2008 SP1 Redistributable Package</title>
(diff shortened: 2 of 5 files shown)


AdmiringWorm (approved):
LGTM

AdmiringWorm:
@jberezanski your changes have been merged, thanks for your contribution 👍

jberezanski:
OK, it seems we have a corner case here.

The vcredist2008 package is built by AU. It is currently in moderation and it should be re-pushed with the same version.
Your `[PUSH ...]` command triggered this logic in appveyor.yml:

```
                      if (Test-Path update.ps1 -ea 0) { ./update.ps1 }
                      choco pack; Push-Package -All;
```

Unfortunately, AU is too smart - it sees the same version on chocolatey.org and does not update the package:
```
vcredist2008 - checking updates using AU version 2021.5.15.194501
URL check
  https://download.microsoft.com/download/5/D/8/5D8C65CB-C849-4025-8E95-C3966CAFD8AE/vcredist_x86.exe
  https://download.microsoft.com/download/5/D/8/5D8C65CB-C849-4025-8E95-C3966CAFD8AE/vcredist_x64.exe
nuspec version: 0.0
remote version: 9.0.30729.616104
New version is available but it already exists in the Chocolatey community feed (disable using $NoCheckChocoVersion):
  https://chocolatey.org/packages/vcredist2008/9.0.30729.616104
```
then the next line packages and pushes version 0.0:
```
Attempting to build package from 'vcredist2008.nuspec'.
Successfully created package 'C:\projects\chocolatey-coreteampackages\manual\vcredist2008\vcredist2008.0.0.nupkg'
Chocolatey v0.10.15
Attempting to push vcredist2008.0.0.nupkg to https://push.chocolatey.org
vcredist2008 0.0 was pushed successfully to https://push.chocolatey.org
```

I've tried to fix this using `[AU vcredist2008]`, which sets the appropriate Force flag for AU. No joy - this only looks at packages in the `automatic` directory.

So, how do we handle this (and I'd like to fix this once and for all, not just for this specific package version)?

I have these ideas:
1) setting `au_Force` in the `[PUSH ...]` branch of code - however, this will result in the package being packed and pushed twice (by AU and by choco)
2) changing `update_all.ps1` so that when a package is forced, the script will also look in the `manual` directory
3) moving vcredist2005/2008/2010 to the `automatic` directory, even though they are not really automatic

My personal favorite is 2), because it seems to be free of unwanted side effects and preserves the proper distinction between manual and automatic packages. I'll try to whip up a PR.

jberezanski:
Okay, change of plans. After a quick look I can see that option 2) would be rather impractical, because the discovery of packages is performed inside AU and only the root path ("$PSScriptRoot\automatic") is passed.

In that case I guess it would be easiest and most robust to just move the packages to automatic, interpreting "automatic" as "packages managed with AU".

To guard against such mistakes in the future, we could add a check to the `[PUSH ...]` handling code in appveyor.yml, which would detect if the `update.ps1` script uses AU and warn that the package should be put in `automatic`.

AdmiringWorm:
There is a 4th option as well, which I forgot about when doing the initial review.

We can add the argument `-NoCheckChocoVersion` in each of the `update.ps1` files to disable the chocolatey.org version check.
IMO, this is probably the best option instead of moving non-automatic packages to the `automatic` directory.
Thoughts?

jberezanski:
I see another benefit of moving - possible future updates would not require remembering to [PUSH] the package.

There is still also the matter of double push: in the [PUSH] code path, the package will be packed and pushed twice - once by AU vla update.ps1, then by explicit choco commands from appveyor.yml. I'm not sure how harmful it would actually be, but it feels... wasteful.

Anyway, NoCheckChocoVersion seems like the second best option. In my own repo I'd still opt for moving, but here I won't insist, if you are convinced that moving is not right.

AdmiringWorm:
> I see another benefit of moving - possible future updates would not require remembering to [PUSH] the package.

I can see that, but be mindful that AU ain't really meant to be used with manual packages (it just can be used).

> There is still also the matter of double push: in the [PUSH] code path, the package will be packed and pushed twice - once by AU vla update.ps1, then by explicit choco commands from appveyor.yml. I'm not sure how harmful it would actually be, but it feels... wasteful.

True, I don't believe there is anything harmful, but we should probably look into if it the push from AU can be disabled in this case (I don't remember anything on the top of my head).

> Anyway, NoCheckChocoVersion seems like the second best option. In my own repo I'd still opt for moving, but here I won't insist, if you are convinced that moving is not right.

Currently, there is a clear distinction between having packages in the automatic and the manual repository. So if a package isn't meant to be updated automatically it should still be in the manual directory.
If going down the route of mixing them though, we should probably in that case completely remove the manual/automatic directories and just keep all packages in a common folder without any distinction between them (I wouldn't mind that though, but not something I would look into doing at this time. And not without having a discussion with everyone involved with the repository).

jberezanski:
> we should probably look into if it the push from AU can be disabled in this case

I've just checked and there are over a dozen existing manual packages which use AU. So I guess the double push has not been a problem before and my worries were unfounded.

The fixed package has successfully been verified on chocolatey.org, so it looks like the end of this problem. Thank you for your help!

PR #1659, opened 2021-05-16
(vcredist2005, vcredist2008, vcredist2010) Move to 'automatic' directory

jberezanski:
These packages are managed by AU. In order to be able to force their update in all scenarios, they need to be under `automatic`, even though the software version and download id is edited manually in the update.ps1 script.

See also https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1658 for more context.

The directories under `manual` were left with a "moved" message to keep links in already published packages working.


jberezanski:
This will also make updating these packages slightly easier, because they will not need to be PUSHed after committing a version change - the next automatic AU run will pick up the changes.

Issue #1671, opened 2021-06-06
(blender) Update is broken


SlimyGooblin:
Update.ps1 is currently unable to update the package, probably because of changes to the blender website's links
## Expected Behavior

update.ps1 when run should change the required information to update the package to the latest stable blender release.
## Current Behavior

Update.ps1 fails with:

    > ./update.ps1
    blender - checking updates using AU version 2020.11.21
    Exception:..\PowerShell\Modules\AU\2020.11.21\Public\Update-Package.ps1:403
    Line |
     403 |          throw "au_GetLatest failed`n$_"
         |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         | au_GetLatest failed Exception calling "Create" with "1" argument(s): "Invalid URI: The URI is empty."

## Possible Solution

Change the commands for $re, $url64, $version64
## Steps to Reproduce (for bugs)

1. Clone chocolatey-coreteampackages/automatic/blender
2. Run update.ps1 from powershell
3.
4.

## Your Environment

* Package Version used:
* Operating System and version: Windows 10 Home 20H2
* Chocolatey version: 0.10.15
* Install/uninstall gist:

PR #1673, opened 2021-06-07
fixed url and version generation in update.ps1

SlimyGooblin:
--Resubmitted as previous pull request replaced entire file unnecessarily--
 
It looks like blender's site changed how they name their installers, breaking update.ps1, so I changed the way the url and version number are generated. Hopefully with some added future-proofing.

## Description

Changed the way the variables $re, $url64, and $version64 are created
## Motivation and Context

update.ps1 doesn't currently run, so the blender package will not automatically update

* fixes: Issue #1671

## How Has this Been Tested?

I ran the modified update.ps1 and installed the generated package locally

## Types of changes

## Checklist:

### DIFF
--- a/automatic/blender/update.ps1
+++ b/automatic/blender/update.ps1
@@ -25,18 +25,22 @@ function global:au_SearchReplace {
 function global:au_GetLatest {
   $download_page = Invoke-WebRequest -Uri $releases -UseBasicParsing
 
-  $re = 'windows64\.msi\/$'
+  #regex to find download url, it should match any url that ends with a 64 bit windows msi
+  $re = '\/[^/]*windows.*64.*\.msi\/$'
   $url64 = $download_page.links | ? href -match $re | select -first 1 -expand href
 
-  $verRe = '[-]'
-  $version64 = $url64 -split "$verRe" | select -last 1 -skip 1
+  # this should get the version number based on Blender's version numbering as of 03-Jun-2020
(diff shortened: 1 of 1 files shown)


RedBaron2:
@SlimyGooblin 
Just a question not that anything is wrong more curiosity.
You find the version with regular expression of `\d\.\d\d\.\d+` , and this is correct. The question is why not something like `\d+\.\d+\.\d+`. The version in the future could change from "0.12.345” to "012.34567.890123”
Just curious, and thanks for looking into this package and making the PR.

SlimyGooblin:
@RedBaron2 
My only reason was that looking through some of blender's past releases, they've never used more than two numbers for minor versions. So I figured this would be a tiny bit "safer" since it could avoid matching the wrong text. Since both work though, changing it shouldn't be a problem if that's preferred.

RedBaron2:
@SlimyGooblin 
Making a "safe" choice is probably the best choice. I'm unaware of any community rule that states that regular expression must be one way or another. @AdmiringWorm or @majkinetor would be the best ones to review the PR. Good work.

SlimyGooblin:
@RedBaron2 
Thanks for the feedback :)

pauby:
Is this still needed? According to the [issue](https://github.com/chocolatey-community/chocolatey-coreteampackages/issues/1671) this was fixed in this (commit](https://github.com/chocolatey-community/chocolatey-coreteampackages/issues/1671).

SlimyGooblin:
@pauby
This should no longer be need, I'll close it now.

PR #1668, opened 2021-05-28
(k9s) Updated K9s update script to reflect changes to asset rename

SayakMukhopadhyay:
## Description

From 0.24.10 onwards, k9s has changed the format of the asset files. See https://github.com/derailed/k9s/releases/tag/v0.24.10. Because of this, the automatic update has been failing. See https://gist.github.com/choco-bot/a14b1e5bfaf70839b338eb1ab7f8226f#k9s

This PR updates the update script to reflect the asset rename
## Motivation and Context

I noticed the chocolatey didn't have the latest version of k9s and on investigating I found that the error is due to the asset rename. Pre 0.24.10, the asset was called `k9s_Windows_x86_64.tar.gz`. From 0.24.10 onwards it is called `k9s_v0.24.10_Windows_x86_64.tar.gz`. Note the version name in the file. The PR solves this by generating the asset file name including the version name.

## How Has this Been Tested?

I have tested this locally by running `update.ps1` and it is able to fetch the latest file and update the relevant metadata. I am running Windows 10 Pro with PS version 5.1.
I have also tested on `chocolatey-test-environment` and `vagrant provision` is successful without errors.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Migrated package (a package has been migrated from another repository)

## Checklist:

- [x] My code follows the code style of this repository.
- [ ] My change requires a change to documentation (this usually means the notes in the description of a package).
- [ ] I have updated the documentation accordingly (this usually means the notes in the description of a package).
- [x] All files are up to date with the latest [Contributing Guidelines](https://github.com/chocolatey-community/chocolatey-coreteampackages/blob/master/CONTRIBUTING.md)
- [x] The added/modified package passed install/uninstall in the chocolatey test environment.
- [x] The changes only affect a single package (not including meta package).

PR #1685, opened 2021-07-05
(k9s) Updated K9s script to handle both types of asset names

SayakMukhopadhyay:
## Description

k9s has changed the format of the asset files in version 0.24.10 which led me to create the PR #1668 to handle the change. The maintainer then decided to revert back that change. This PR handles both types of asset names incase the maintainer decides to change the name again.
## Motivation and Context

Pre `0.24.10`, the asset was called `k9s_Windows_x86_64.tar.gz`. On `0.24.10` onwards it was called `k9s_v0.24.10_Windows_x86_64.tar.gz` and in `0.24.11` to `0.24.13`, they have reverted back to `k9s_Windows_x86_64.tar.gz`.  The PR solves this by generating the asset file name first by excluding the version name and on failure including the version name.
## How Has this Been Tested?

I have tested this locally by running `update.ps1` and it is able to fetch the latest file and update the relevant metadata. I am running Windows 10 Pro with PS version 5.1.
I have also tested on `chocolatey-test-environment` and `vagrant provision` is successful without errors.
## Types of changes

## Checklist:

### DIFF
--- a/automatic/k9s/tools/chocolateyinstall.ps1
+++ b/automatic/k9s/tools/chocolateyinstall.ps1
@@ -5,7 +5,7 @@ $toolsDir      = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
 
 $packageArgs = @{
   PackageName    = $packageName
-  FileFullPath64 = Get-Item $toolsDir\k9s_*_Windows_x86_64.tar.gz
+  FileFullPath64 = Get-Item $toolsDir\k9s_*Windows_x86_64.tar.gz
   Destination    = $toolsDir
 }
 
@@ -13,10 +13,10 @@ Get-ChocolateyUnzip @packageArgs
 
 $packageArgs2 = @{
   PackageName    = $packageName
-  FileFullPath64 = Get-Item $toolsDir\k9s_*_Windows_x86_64.tar
+  FileFullPath64 = Get-Item $toolsDir\k9s_*Windows_x86_64.tar
   Destination    = $toolsDir
 }
 
 Get-ChocolateyUnzip @packageArgs2
 
-Remove-Item "$toolsDir\k9s_*_Windows_*.tar"
+Remove-Item "$toolsDir\k9s_*Windows_*.tar"
--- a/automatic/k9s/update.ps1
+++ b/automatic/k9s/update.ps1
@@ -34,7 +34,11 @@ function global:au_GetLatest {
 
   $checksumAsset = $domain + ($download_page.Links | ? href -match "checksums\.txt$" | select -first 1 -expand href)
   $checksum_page = Invoke-WebRequest -Uri $checksumAsset -UseBasicParsing
-  $filename64 = -join('k9s_v', $version, '_Windows_x86_64.tar.gz')
+  if($VersionStyle -eq 0) {
+    $filename64 = 'k9s_Windows_x86_64.tar.gz'
+  } elseif ($VersionStyle -eq 1) {
+    $filename64 = -join('k9s_v', $version, '_Windows_x86_64.tar.gz')
+  }
   $checksum64 = [regex]::Match($checksum_page, "([a-f\d]+)\s*$([regex]::Escape($filename64))").Groups[1].Value
 
   return @{
@@ -47,4 +51,13 @@ function global:au_GetLatest {
   }
 }
 
-update -ChecksumFor none -Force:$Force
+$global:VersionStyle = $null
+
+try {
+  $global:VersionStyle = 0
+  update -ChecksumFor none -Force:$Force
+}
+catch {
+  $global:VersionStyle = 1
+  update -ChecksumFor none -Force:$Force
+}


automatic/k9s/update.ps1
```diff
@@ -34,7 +34,11 @@ function global:au_GetLatest {
 
   $checksumAsset = $domain + ($download_page.Links | ? href -match "checksums\.txt$" | select -first 1 -expand href)
   $checksum_page = Invoke-WebRequest -Uri $checksumAsset -UseBasicParsing
-  $filename64 = -join('k9s_v', $version, '_Windows_x86_64.tar.gz')
+  if($VersionStyle -eq 0) {
```
AdmiringWorm, line 37: Not sure If I like that a global `VersionStyle` variable is used for this.

Is it possible to first parsing the checksum with the first style, then if that fails/is empty then retry the same checksum regex with the different style?

automatic/k9s/tools/chocolateyinstall.ps1
```diff
@@ -5,18 +5,18 @@ $toolsDir      = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
...
   PackageName    = $packageName
-  FileFullPath64 = Get-Item $toolsDir\k9s_*_Windows_x86_64.tar
+  FileFullPath64 = Get-Item $toolsDir\k9s_*Windows_x86_64.tar
   Destination    = $toolsDir
 }
 
 Get-ChocolateyUnzip @packageArgs2
 
-Remove-Item "$toolsDir\k9s_*_Windows_*.tar"
+Remove-Item "$toolsDir\k9s_*Windows_*.tar"
```
AdmiringWorm, line 22: Can you change this line to
```suggestion
Remove-Item "$toolsDir\k9s_*Windows_*.tar","$toolsDir\k9s_*Windows_*.tar.gz"`
```

Seems like we are forgetting to remove the `.tar.gz` archive.

SayakMukhopadhyay: Yeah, thats a good catch. I believe I can change that line to `Remove-Item "$toolsDir\k9s_*Windows_*.tar*"`. That should catch both of them. What do you think?

automatic/k9s/update.ps1
SayakMukhopadhyay: I am not sure if I understand this. The reason I had to use a global variable was the fact that I couldn't pass a value or context into the `au_GetLatest` method. Also the method "return" the url string which "supposedly" is handled by AU somehow. Hence I was passing a context from the global scope, which is triggering the update call, into the method using a global variable.

automatic/k9s/tools/chocolateyinstall.ps1
RedBaron2: @SayakMukhopadhyay
Just suggesting a non-global way. You could try using `Get-Version` on the asset filename. If that returns a version then it will be the new filename, and the old filename if nothing.
Just a suggestion.

automatic/k9s/update.ps1
RedBaron2: @SayakMukhopadhyay
Yes, I meant to post it here, sorry for the confusion.
AU requires a version to be passed in order to determine if the package is updated.
You are using the VersionStyle variable to determine the filename. Using the helper function `Get-Version` to get the version from the URL would eliminate the step of using your VersionStyle variable in the global scope. This would in turn make @admiringworm happier about nixing the global variable.

automatic/k9s/tools/chocolateyinstall.ps1
RedBaron2: Please see other comment, thank you.

SayakMukhopadhyay:
Sorry for the silence. Been super busy at work. I will pick this up over the weekend if I don't have work.

automatic/k9s/tools/chocolateyinstall.ps1
AdmiringWorm: > Yeah, thats a good catch. I believe I can change that line to Remove-Item "$toolsDir\k9s_*Windows_*.tar*". That should catch both of them. What do you think?

Yeah, that should work.

automatic/k9s/update.ps1
```diff
@@ -34,7 +34,11 @@ function global:au_GetLatest {
...
+    $filename64 = 'k9s_Windows_x86_64.tar.gz'
+  } elseif ($VersionStyle -eq 1) {
+    $filename64 = -join('k9s_v', $version, '_Windows_x86_64.tar.gz')
+  }
   $checksum64 = [regex]::Match($checksum_page, "([a-f\d]+)\s*$([regex]::Escape($filename64))").Groups[1].Value
```
AdmiringWorm, line 42: To refer to your question about what I mean, I was thinking something along the lines of:

```suggestion
  $regexFormat = "([a-f\d]+)\s*{0}"
  $possibleFileNames = @(
    'k9s_Windows_x64_64.tar.gz'
    ('k9s_v{0}_Windows_x86_64.tar.gz' -f $version)
  )
  foreach ($fileName64 in $possibleFileNames) {
    $regex = $regexFormat -f ([regex]::Escape($fileName64))
    if ($checksum_page -match $regex) {
      $checksum64 = $Matches[1]
      break
    }
  }
```

**NOTE: I have not tested this snippet if it works**

pauby:
This is superceded by #1713. Closing.

PR #1695, opened 2021-08-05
smartftp - removed vcredist dependency

matbech:
## Description
The latest version of SmartFTP no longer requires vcredist:
"The Visual C++ Redistributable Package (vcredist) is no longer installed."
https://www.smartftp.com/en-us/changelog/1

## How Has this Been Tested?
Manual

## Types of changes


AdmiringWorm:
The message: `"The Visual C++ Redistributable Package (vcredist) is no longer installed."` is not the same as it no longer being needed.

This only means that they themself won't install it as part of the installation, but says nothing about wether the application perhaps needs it on some systems to be able to run.

Also, you only said you did `Manual` testing, but this doesn't tell me anything.

How did you test it, what environments were used, was it on your own computer, was it in a clean VM, etc.
What steps did you do to ensure that everything worked as expected?

matbech:
Well, I'm the one responsible for this change log entry ...

AdmiringWorm:
@matbech so just to be sure, you can confirm that SmartFTP does not need any form of vcredist to run on any platform?

It is now completely self-contained?

matbech:
Yes

AdmiringWorm (approved):
Thanks for confirming

PR #1698, opened 2021-08-10
(firefox) Improve the method of detecting system locale

msoxzw:
The current method of detecting system locale is not robust enough. Take "zh-HK" as an example, it should be fallback to "zh_TW", but this package fallback "zh-HK" to "en-US". This commit detects fallback locale by sending `Accept-Language: zh-HK` header to https://www.mozilla.org/ to get Mozilla fallback locale.

### DIFF
--- a/automatic/firefox/tools/helpers.ps1
+++ b/automatic/firefox/tools/helpers.ps1
@@ -43,7 +43,9 @@ function GetLocale {
   $systemLocalizeAndCountry = (Get-UICulture).Name
   $systemLocaleTwoLetter = (Get-UICulture).TwoLetterISOLanguageName
   Write-Verbose "System locale is: '$systemLocalizeAndCountry'..."
-  $fallbackLocale = 'en-US'
+  $Response = Invoke-WebRequest 'https://www.mozilla.org/' -UseBasicParsing -Headers @{'Accept-Language'=$systemLocalizeAndCountry} -MaximumRedirection 0 -ErrorAction Ignore
+  $fallbackLocale = $Response.Headers.Location.Trim('/')
+  Write-Verbose "Fallback locale is: '$fallbackLocale'..."
 
   $locales = $localeFromPackageParameters,$localeFromPackageParametersTwoLetter, `
     $alreadyInstalledLocale, $systemLocalizeAndCountry, $systemLocaleTwoLetter, `


majkinetor (changes requested):
image

Omitting MaximumRedirects doesn't give location either.

majkinetor:
Nvm, above is in `pwsh`, in powershell it works.

thoscut:
This change breaks internalization of the package with 
```batch
choco download firefox --internalize --internalize-all-urls
```

The installation of the package fails:
```batch
Firefox v92.0.1
firefox package files upgrade completed. Performing other installation steps.
ERROR: The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Illegal characters in path.
The upgrade of firefox was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\Firefox\tools\chocolateyInstall.ps1'.
```

TheCakeIsNaOH:
@thoscut You may want to contact support to see if they would have any workarounds for this issue. I'm not sure if it would work, but you could try internalizing without `--internalize-all-urls`

Two other options would be to internalize this package manually, or possibly use [choco-remixer](https://github.com/TheCakeIsNaOH/choco-remixer) [Note it is completely unofficial]

pauby:
@thoscut That internalization works for me:

```
> choco download firefox --internalize --internalize-all-urls
Chocolatey v0.11.1 Business
Downloading existing package(s) to C:\Users\Paul\download

chocolatey-core.extension v1.3.5.1

Recompile not necessary. There is no install script. Use the nupkg file
 as is in 'C:\Users\Paul\download\chocolatey-core.extension'

Firefox v93.0
Found internalizable Chocolatey functions. Inspecting values for remote resources.

Adding -UseOriginalLocation to Install-ChocolateyPackage (if found).
 There is a bug in Chocolatey v0.10.1 that occurs when there is only
 one url replaced (no 32 and 64 bit Windows). Be sure to remove the
 parameter and recompile if you have any 0.10.1 installations. You
 can also turn off the feature 'internalizeAppendUseOriginalLocation' to not append
Updating chocolateyInstall.ps1 with local resources.
Recompiling package.

Recompiled package files available at 'C:\Users\Paul\download\Firefox'
Recompiled nupkg available in 'C:\Users\Paul'
```

However, [these lines](https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1698/files#diff-b83df586bc956a7b171344bc3c71ac191fe3fe6893aec94a0957c3287fa4ab0bR46-R48) look to be problematic if there is no valid response back (what would `$fallBackLocale` actually be).

pauby:
I've raised a [PR](https://github.com/chocolatey-community/chocolatey-coreteampackages/pull/1724) to try and address this.

PR #1903, opened 2022-06-08
[graphviz] Update auto update installer match string

adrianinsaval:
## Description
Not tested but looking at https://www.graphviz.org/download/ the `stable_` prefix was dropped since version 2.50

## Motivation and Context
automatic updates are not working, this should fix it


TheCakeIsNaOH (approved):
Working for me locally.

PR #1952, opened 2022-07-21
(kubernetes-cli) Update to the latest supported versions

rgl:
## Description

Update `kubernetes-cli` to the latest supported versions.

## Motivation and Context

See https://github.com/chocolatey-community/chocolatey-packages/discussions/1901#discussioncomment-3082957

## How Has this Been Tested?

I've tested (un)installing in my machine.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/kubernetes-cli/update.ps1
+++ b/automatic/kubernetes-cli/update.ps1
@@ -25,25 +25,46 @@ function global:au_SearchReplace {
 }
 
 function global:au_GetLatest {
+    # Only report the supported Kubernetes streams.
+    # NB Upstream Kubernetes only supports the last three minor versions.
+    #    However, they can also do pre-releases before a final version is
+    #    released, so we have to check for an extra changelogs/minor
+    #    version to be able to find all the supported streams.
+    # See https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md
+    $minor_versions_to_keep = 3
+
     $changelog_page = Invoke-WebRequest -Uri $changelogs -UseBasicParsing
-    #Only the latest three minor versions are supported: https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md
-    #However, there are prereleases before the version is released, so the latest four changelogs/minor versions need to be checked.
-    $minor_version_changelogs = $changelog_page.links | ? href -match "CHANGELOG-[\d\.]+\.md" | Select-Object -Expand href -First 4
-    
+    $minor_version_changelogs = $changelog_page.links `
+      | Where-Object href -match "CHANGELOG-(?<version>\d+\.\d+)\.md`$" `
+      | Select-Object -Expand href -First ($minor_versions_to_keep+1)
+
     $streams = @{}
-    
+
     $minor_version_changelogs | ForEach-Object {
-        $minor_version = ($_ -split "CHANGELOG-" | Select-Object -Last 1).trim(".md")
-        $minor_changelog_page = Invoke-WebRequest -UseBasicParsing -Uri ("https://github.com" + $_)
-        $url64 = $minor_changelog_page.links | ? href -match "/v(?<version>[\d\.]+)/kubernetes-client-windows-amd64.tar.gz" | Select-Object -First 1 -Expand href
+        if ($_ -notmatch "CHANGELOG-(?<version>\d+\.\d+)\.md`$") {
+          return
+        }
+        $minor_version = $matches.version
+        if ($streams.ContainsKey($minor_version)) {
+          return
+        }
+        $minor_changelog_page = Invoke-WebRequest -UseBasicParsing -Uri "https://github.com$_"
+        $url64 = $minor_changelog_page.links `
+          | Where-Object href -match "/v(?<version>\d+(\.\d+)+)/kubernetes-client-windows-amd64.tar.gz" `
+          | Select-Object -First 1 -Expand href
         $patch_version = $matches.version
-        $url32 = $minor_changelog_page.links | ? href -match "/v[\d\.]+/kubernetes-client-windows-386.tar.gz" | Select-Object -First 1 -Expand href
-        
+        if (!$url64) {
+          return
+        }
+        $url32 = $minor_changelog_page.links `
+          | Where-Object href -match "/v(?<version>\d+(\.\d+)+)/kubernetes-client-windows-386.tar.gz" `
+          | Select-Object -First 1 -Expand href
+
         $streams.Add($minor_version, @{
             Version     = $patch_version
             URL32       = $url32
             URL64       = $url64
-            ReleaseNotes= ("https://github.com" + $_)
+            ReleaseNotes= "https://github.com$_"
         })
     }
 
--- a/automatic/kubernetes-cli/update.ps1
+++ b/automatic/kubernetes-cli/update.ps1
@@ -26,17 +26,16 @@ function global:au_SearchReplace {
 
 function global:au_GetLatest {
     # Only report the supported Kubernetes streams.
-    # NB Upstream Kubernetes only supports the last three minor versions.
-    #    However, they can also do pre-releases before a final version is
-    #    released, so we have to check for an extra changelogs/minor
-    #    version to be able to find all the supported streams.
-    # See https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md
-    $minor_versions_to_keep = 3
 
     $changelog_page = Invoke-WebRequest -Uri $changelogs -UseBasicParsing
+
+    # There is quite a few versions that do not exist on chocolatey.org
+    # and since the limit of pushed packages is 10, we need to limit the amount
+    # of streams that we parse. Once packages are approved we can increase/remove
+    # the limit.
     $minor_version_changelogs = $changelog_page.links `
       | Where-Object href -match "CHANGELOG-(?<version>\d+\.\d+)\.md`$" `
-      | Select-Object -Expand href -First ($minor_versions_to_keep+1)
+      | Select-Object -Expand href -First 10
 
     $streams = @{}
 


automatic/kubernetes-cli/update.ps1
```diff
@@ -28,17 +28,22 @@ function global:au_GetLatest {
     $changelog_page = Invoke-WebRequest -Uri $changelogs -UseBasicParsing
     #Only the latest three minor versions are supported: https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md
     #However, there are prereleases before the version is released, so the latest four changelogs/minor versions need to be checked.
-    $minor_version_changelogs = $changelog_page.links | ? href -match "CHANGELOG-[\d\.]+\.md" | Select-Object -Expand href -First 4
-    
+    $minor_version_changelogs = $changelog_page.links `
+      | Where-Object href -match "CHANGELOG-(?<version>\d+\.\d+)\.md`$" `
+      | Select-Object -Expand href -Skip 1 -First 3
```
TheCakeIsNaOH, line 33: This should work when there are three supported release versions, and one prerelease version, like if `1.22`, `1.23`, and `1.24` are released, and `1.25 has prerelease versions available.

But it would break once `1.25` is released if they do not have a `1.26` prerelease available.

Feel free to correct me if I am misunderstanding this.

automatic/kubernetes-cli/kubernetes-cli.json
```diff
@@ -1,10 +1,5 @@
 ﻿{
-    "1.4":  "1.4.12",
-    "1.5":  "1.5.8",
-    "1.6":  "1.6.13",
-    "1.7":  "1.7.16",
-    "1.8":  "1.8.15",
-    "1.9":  "1.9.11",
-    "1.1":  "1.19.5",
-    "1.2":  "1.24.1"
+    "1.22":  "1.22.12",
```
TheCakeIsNaOH, line 2: No need to include the changes in this file, as AU will update it when it updates the package.

automatic/kubernetes-cli/kubernetes-cli.nuspec
```diff
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>kubernetes-cli</id>
-    <version>1.24.1</version>
```
TheCakeIsNaOH, line 6: No need to include the changes in this file, as AU will update it when it updates the package.

automatic/kubernetes-cli/legal/VERIFICATION.txt
```diff
@@ -4,12 +4,12 @@ Verification is intended to assist the Chocolatey moderators and community
 in verifying that this package's contents are trustworthy.
 
 The embedded software have been downloaded from the listed download
-location on <https://github.com/kubernetes/kubernetes/releases/latest>
+location on <https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md>
 and can be verified by doing the following:
 
 1. Download the following:
-  32-bit software: <https://dl.k8s.io/v1.24.1/kubernetes-client-windows-386.tar.gz>
```
TheCakeIsNaOH, line 11: No need to include the changes in this file, as AU will update it when it updates the package.

automatic/kubernetes-cli/update.ps1
rgl: I'm not sure, but we will notice if/when this becomes a problem once 1.25/1.26 stable is (not) released.

BTW, there might be a better way for looking for the k8s versions, but I did not try to find it. I just made that part of the existing code more clear.

automatic/kubernetes-cli/kubernetes-cli.json
rgl: I've dropped this files changes. Pleas verify.

automatic/kubernetes-cli/kubernetes-cli.nuspec
rgl: I've dropped this files changes. Pleas verify.

automatic/kubernetes-cli/legal/VERIFICATION.txt
rgl: I've dropped this files changes. Pleas verify.

TheCakeIsNaOH (commented):
Thanks for updating this, there is still the one open question, but maybe someonelse has thoughts?

automatic/kubernetes-cli/update.ps1
rgl: This is now fixed. It will download the last three released version streams.

rgl:
> Thanks for updating this, there is still the one open question, but maybe someon else has thoughts?

It should be fixed now. Please verify.

automatic/kubernetes-cli/update.ps1
```diff
@@ -25,25 +25,46 @@ function global:au_SearchReplace {
...
+    $minor_versions_to_keep = 3
+
     $changelog_page = Invoke-WebRequest -Uri $changelogs -UseBasicParsing
-    #Only the latest three minor versions are supported: https://github.com/kubernetes/sig-release/blob/master/release-engineering/versioning.md
-    #However, there are prereleases before the version is released, so the latest four changelogs/minor versions need to be checked.
-    $minor_version_changelogs = $changelog_page.links | ? href -match "CHANGELOG-[\d\.]+\.md" | Select-Object -Expand href -First 4
-    
+    $minor_version_changelogs = $changelog_page.links `
+      | Where-Object href -match "CHANGELOG-(?<version>\d+\.\d+)\.md`$" `
+      | Select-Object -Expand href -First ($minor_versions_to_keep+1)
```
AdmiringWorm, line 39: Any reason why you restrict the amount of links?

If there are no reason, this can be changed to
```suggestion
      | Select-Object -Expand href
```

So we can monitor all links (In case there are a patch being created for an older release at some point).

automatic/kubernetes-cli/update.ps1
```diff
@@ -25,25 +25,46 @@ function global:au_SearchReplace {
...
     $streams = @{}
-    
+
     $minor_version_changelogs | ForEach-Object {
-        $minor_version = ($_ -split "CHANGELOG-" | Select-Object -Last 1).trim(".md")
-        $minor_changelog_page = Invoke-WebRequest -UseBasicParsing -Uri ("https://github.com" + $_)
-        $url64 = $minor_changelog_page.links | ? href -match "/v(?<version>[\d\.]+)/kubernetes-client-windows-amd64.tar.gz" | Select-Object -First 1 -Expand href
+        if ($streams.Count -ge $minor_versions_to_keep) {
+          return
+        }
```
AdmiringWorm, line 46: We shouldn't restrict the amount of streams being added.

Please remove this if statement.

automatic/kubernetes-cli/update.ps1
```diff
@@ -25,25 +25,46 @@ function global:au_SearchReplace {
...
+        if ($_ -notmatch "CHANGELOG-(?<version>\d+\.\d+)\.md`$") {
+          return
+        }
+        $minor_version = $matches.version
+        $minor_changelog_page = Invoke-WebRequest -UseBasicParsing -Uri "https://github.com$_"
```
AdmiringWorm, line 51: Can we add a check before this line to see if the stream already contains the `$minor_version` so we don't need to do a web request when not needed.

AdmiringWorm (changes requested):

automatic/kubernetes-cli/update.ps1
rgl: The restrictions came from the existing package. I just made them more explicit. From what I understood, this package intent was to track only the latest version of the latest three kubernetes releases.

But now it seems you want to track all versions and releases? Even no longer supported versions? So scape all the versions that are available?

AdmiringWorm: Typically the restriction is only in place for the beginning, to not overwhelm Chocolatey Community repository with too many versions at the same time, as well as not needing to fix a lot of package versions if they fail the verifier/validator.

Most likely, this restriction was just forgotten to be removed. However, since it already has the restriction we can leave it here for this PR.

automatic/kubernetes-cli/update.ps1
AdmiringWorm: Although we can leave the 3 limit restriction for now, this if statement should still be removed.

automatic/kubernetes-cli/update.ps1
AdmiringWorm: Just to note, this is more of a safe guard in case the changelog gets incorrectly updated at some point.

rgl: Not sure you can prevent that with such a simple heuristic. For example, I guess this can go both ways. It can be flapping between wrong/right, so, I would not bother.

automatic/kubernetes-cli/update.ps1
rgl: I've removed it in the last push. The output still looks like:

```json
{
  "Streams": {
    "1.22": {
      "Version": "1.22.12",
      "ReleaseNotes": "https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md",
      "URL32": "https://dl.k8s.io/v1.22.12/kubernetes-client-windows-386.tar.gz",
      "URL64": "https://dl.k8s.io/v1.22.12/kubernetes-client-windows-amd64.tar.gz"
    },
    "1.23": {
      "Version": "1.23.9",
      "ReleaseNotes": "https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md",
      "URL32": "https://dl.k8s.io/v1.23.9/kubernetes-client-windows-386.tar.gz",
      "URL64": "https://dl.k8s.io/v1.23.9/kubernetes-client-windows-amd64.tar.gz"
    },
    "1.24": {
      "Version": "1.24.3",
      "ReleaseNotes": "https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md",
      "URL32": "https://dl.k8s.io/v1.24.3/kubernetes-client-windows-386.tar.gz",
      "URL64": "https://dl.k8s.io/v1.24.3/kubernetes-client-windows-amd64.tar.gz"
    }
  }
}
```

Should it contain all the all versions inside a given stream (although the function name is `au_GetLatest`)? Or should it only contain the latest (as its now)?

automatic/kubernetes-cli/update.ps1
AdmiringWorm: Adding a simple if check if to see if the stream have already been added is mainly for the benefit of not aborting the entire collection.

If the changelog at some point will be incorrectly updated with the same major+minor version, then an exception will happen causing nothing to actually be opened, even the streams that was already added/will be added.

So, as I mentioned it is purely a safe guard so the updater won't throw an exception nothing else.

Basically the only thing that need to be added is:

```powershell
if ($streams.ContainsKey($minor_version)) {
  return
}
```

rgl: Done. Please verify.

automatic/kubernetes-cli/update.ps1
AdmiringWorm: Could be that something unrelated prevents additional streams from being added, I'll ignore that for now and come back to it later to check what is causing the issue.

automatic/kubernetes-cli/update.ps1
AdmiringWorm: The changes look good, thank you for adding the `if` check.

AdmiringWorm (approved):
LGTM, I have updated the stream limit to 10 packages. I would normally remove the limit, but it seems we are missing quite a few versions on community.chocolatey.org which exceeds the number of versions that can be in the submitted queue.

rgl:
Any idea why https://community.chocolatey.org/packages/kubernetes-cli still does not have the latest versions?

AdmiringWorm:
Seems that the latest version failed to be pushed to the Community Chocolatey Repository, but AU did not see it as an error. I'll get the stream for that version reset so it will try it again.

rgl:
It has worked. 1.24.3 is now available. Thanks for looking into it!

PR #1993, opened 2022-10-01
Bring over Python 3.x minor version packages from Jack's repository

jack1142:
## Description
This PR moves Python 3.x minor version packages from my repository as this has been suggested during a review of those on the Chocolatey Community Repository.

So far, I've only copied over the packages and updated it according to this repository's guidelines as I haven't yet been informed what I have to do to get it accepted here.

Some open questions:
- When uploading these packages, I was also aiming to create a python3.9 package but as it later turned out, the name is already taken, sort of. There's actually not a single version of the package listed as the only uploaded version was fully rejected by the automatic review process yesterday. Should I be pursuing this package name through some process or is this something that you can take care of?
- Looking at the current python3 Chocolatey package, I imagine I would have to add support for things that this package supports and that my package currently don't such as:
    - package parameters that python3 package supports
    - installation location folder lockdown
    - Maybe something else?
- Should I put contents of `update.ps1` scripts in some common location to avoid duplication or do I not need to concern myself with that?
- The packages are currently in review so it is still possible to update them. Do I need to do anything special so that these packages get pushed with the same version number when you merge this PR or is it something that you would take care of?

## Motivation and Context
It's been suggested to me by @gep13 to move my packages for Python 3.x minor versions in their review:
https://community.chocolatey.org/packages/python3.7/3.7.9#versionhistory
My words on why I think it is necessary to have a package per each minor version of Python can be found there as well.

## How Has this Been Tested?
Tested using packages generated with `$au_Force = $au_WhatIf = $true; ./update.ps1` in the package folder and then tested locally (installation, uninstallation, running some simple code in installed Python's REPL, verifying that created shims work) in chocolatey-test-environment. The AU scripts have also been run in AppVeyor's Windows environment which is where the currently reviewed versions of packages were uploaded from.

## Types of changes

## Checklist:

## Original Location
- [Original Repository](https://github.com/jack1142/choco-auto)
- [No Open Issues](https://github.com/jack1142/choco-auto)
  - This is a brand new package currently in review so there are no open issues.
    - Haven't made one yet, it sort of feels a bit early considering that this has only really been talked about in the Chocolatey package review.
- [x] The [migration guidelines](https://github.com/chocolatey-community/chocolatey-packages/wiki/Package-migration-process) have been followed


jack1142:
Seems like CI is mistakenly treating generated packages as versions of `python3` rather than versions of respective `python3.x` packages which makes it fail due to trying to install a `x.3.x.y.z` version of `python3` package rather than `3.x.y.z` version of `python3.x` package.

AdmiringWorm:
Looking at the code, this is a very good start for this.
While there is nothing in the code itself that is wrong, there is a lot of duplication with more or less the exact same code.

I think this is a very good candidate to be part of streams, instead of having a distinct package for every versions (the id of the package can be changed by setting `PackageName` in the return hashtable).
Would this be something you would be able to change?
If not, we can get this merged into a temporary branch until we have completed that work.

With the naming of the packages though, we should avoid using dots in the package name, and in my opinion it would be better if the package name would be `python36`, `python37`, etc.

Thoughts on this @JPRuskin @pauby

jack1142:
I'll try switching it to streams soon, maybe later today or tomorrow.

I'll keep the package names as is until agreement on that is reaches if that's okay.

AdmiringWorm:
> I'll try switching it to streams soon, maybe later today or tomorrow.
> 
> I'll keep the package names as is until agreement on that is reaches if that's okay.

Sure thing, keep the names as it is for now. When we decide which names to use, and if it is updated to use streams, it will be quite easy to change them later.

pauby:
> With the naming of the packages though, we should avoid using dots in the package name, and in my opinion it would be better if the package name would be python36, python37, etc.

I would agree. A dot should only be used if it matches one of the known package types (`.extension`, `.powershell`, `.portable`, `.install`, etc.).

TheCakeIsNaOH:
> With the naming of the packages though, we should avoid using dots in the package name, and in my opinion it would be better if the package name would be python36, python37, etc.

At the moment, the current exceptions to not having dots in the name are these:
- To add a postfix to signify that a package is a specific type, like `.install`, `.extension`, etc.
- If the software name itself contains a dot, the example given is `paint.net`
- For when the package ID contains a major+minor version. Examples are `netfx-4.8`,`dotnet-3.1-runtime`, and lots of other dotnet packages.

Given that `python3.minorversion` fits the last case, I'm personally fine with having the dot to seperate the major and minor version. 

I generally would prefer to have the dot, because then it differentiates what is the major version part and what is the minor version part. Say for python `v3.10.x`, if the package ID `python310` was used, it is not clear if that is refering to python `v30.0.x`, or `v3.10.x` or even `v310.0.x`. The package title and/or version is often visable at the same time and allow someone to figure out the notation used (like in the CCR webpage), but sometimes may take more work to find (such as a non-verbose list/search on the CLI).

Or for another example, given the rate of dotnet releases, dotnet 11 is only a cople of years away, and once that happens, if there is a package named `dotnet-11-runtime`, it would not be clear if that is dotnet `v11.x` or dotnet `v1.1.x`, assuming the policy would be that major and minor versions in the package id should not be seperated by a dot.

jack1142:
I agree with TheCakeIsNaOH's reasoning and some of these reasons were why I chose to use the dot in the name. Debian-based distros use the same convention which may make it more familiar to some users though there are also distros that don't use dots for minor Python version packages (e.g. Fedora-based).

Anyhow, I've made significant progress towards moving this to streams, there are still some small changes I need to make before it's ready (notably, au is hanging on search & replace when processing a second stream, probably because of LICENSE.txt update, I'll try switching that to full overwrite of the file instead of a regex matching whole file contents) + I need to test it out in chocolatey VM to confirm that it's working as intended but most of the work on the code is done I would say. Moving this to a stream in python3 also means that the package parameters from python3 which I didn't add are now available as well. I did end up using package name without dots after moving it to streams but I will change it back if you agree that the package name with a dot is superior :)

AdmiringWorm:
> For when the package ID contains a major+minor version. Examples are netfx-4.8,dotnet-3.1-runtime, and lots of other dotnet packages.

And neither of those packages is maintained in this repository, we have stricter rules and a few differences when comparing against the Chocolatey Community Repository.
Do note that the `netfx-*` naming is primarily due to historical reasons with other versions (not sure about the dotnet packages, most likely an oversight when the initial version was pushed).

After talking about it a bit with @pauby internally, we have come to the agreement that we do not accept using a `.` in the name for the version as this goes against the naming style referred to in the Chocolatey Documentation (https://docs.chocolatey.org/en-us/create/create-packages#naming-your-package), that there are existing packages that violate this does not change the matter in this repository.

However, as an alternative, we can accept using dashes `-` to separate the major and minor versions, but not the dots.

jack1142:
> as this goes against the naming style referred to in the Chocolatey Documentation

In response to the automated review on the chocolatey community repository, I argued that `3.8` *is* part of the software name but I guess it's a rather weak point. The only other point I made for this name is that the naming style in Chocolatey documentation tells you to use the same package id as popular Linux distributions if they already have a package of it and this is the package id in Debian-based distros. This is just reiterating my points from that review so I don't really expect your decision to change here.

> However, as an alternative, we can accept using dashes `-` to separate the major and minor versions, but not the dots.

I think I would prefer to not use a separator in that case. I know this goes against TheCakeIsNaOH's argument that this is making it harder to figure out the version but I think that it's more important to use a naming scheme that will be more natural to the users. I don't think I've seen any package that would use `-` for version separator *and* there are plenty of distros that use the `python3x` naming scheme already which would definitely make it more natural to those people as well as make it easier to just guess the name in general in comparison to `python3-x`.

---

So unless someone wants to argue really hard for `python3-x` name, I think this means that there's no open question for this PR anymore and it is ready for review.

AdmiringWorm:
> The only other point I made for this name is that the naming style in Chocolatey documentation tells you to use the same package id as popular Linux distributions if they already have a package of it and this is the package id in Debian-based distros.

While it is true that the documentation says to use the same package id as popular Linux distributions, remember that Debian isn't the only popular distros.
Additionally, there are so much inconsistency between linux distros to actually honor this part of the documentation (even among popular distros).
Some distros use dot separated, others doesn't separate and several doesn't even offer major+minor packages [](https://repology.org/project/python/versions).

All in all, the decision won't be changed.
I have noticed you updated the PR, and have it on my TODO list to do another review on this PR on Monday (unfortunately won't have time until then).

> I don't think I've seen any package that would use - for version separator and there are plenty of distros that use the python3x naming scheme already which would definitely make it more natural to those people as well as make it easier to just guess the name in general in comparison to python3-x.

It is true that no other packages I am aware of uses `dashes` as a version separator, that suggestion was mainly given as an option while still honoring the the naming guidelines, and separating major and minor version.
To be honest though, whether a separator is used or not doesn't really matter to me, as long as the package id does not use dots in it.

jack1142:
> All in all, the decision won't be changed.
> I have noticed you updated the PR, and have it on my TODO list to do another review on this PR on Monday (unfortunately won't have time until then).

@AdmiringWorm ping in case you forgot. If you just haven't had time to look at this, ignore me :smile:

AdmiringWorm:
> @AdmiringWorm ping in case you forgot. If you just haven't had time to look at this, ignore me 😄

I still have this PR on my TODO list, but unfortunately haven't had the time yet. I plan to look at this PR tomorrow if nothing else prevents me.

automatic/python3/update.ps1
```diff
@@ -12,52 +14,146 @@ function global:au_SearchReplace {
...
+
+function global:au_AfterUpdate($Package) {
+    $cdata = $Package.NuspecXml.CreateCDataSection($Latest.Copyright)
+    $xml_Copyright = $Package.NuspecXml.GetElementsByTagName('copyright')[0]
+    $xml_Copyright.RemoveAll()
+    $xml_Copyright.AppendChild($cdata) | Out-Null
+    $Package.NuspecXml.package.metadata.licenseUrl = $Latest.LicenseUrl
+    $Package.NuspecXml.package.metadata.title = $Latest.Title
+    $Package.SaveNuspec()
+    $Latest.License | Out-File './legal/LICENSE.txt'
```
AdmiringWorm, line 77: For consistency sake, make use of the helper we already have available.

```suggestion
    Update-Metadata -data @{
        copyright  = $Latest.Copyright
        licenseUrl = $Latest.LicenseUrl
        title      = $Latest.Title
    }
```

automatic/python3/update.ps1
```diff
@@ -12,52 +14,146 @@ function global:au_SearchReplace {
...
 
-    $url32 = $download_page.links | ? { $_.href -match "python-3.+.(exe|msi)$" -and $_.href -notmatch "amd64" } | select -first 1 -expand href
-    $url64 = $download_page.links | ? href -match "python-3.+amd64\.(exe|msi)$" | select -first 1 -expand href
-    if (!$url32 -or !$url64) {
-      Write-Host "Skipping due to missing installer: '$version'"; return
+  $version_re = '3\.(?<minor>\d+)\.(?<micro>\d+)(?:(?<releaselevel>a|b|rc)(?<serial>\d+))?'
+  $re = '^python-(?<version>' + $version_re + ')(?<amd64>-amd64)?\.exe$'
+  # collect URL pairs for all Python versions
+  $all_versions = @{ }
+  foreach ($file in $release_files) {
```
AdmiringWorm, line 87: We rarely use `foreach` loops in this repository and prefer piped calls.

If possible, change the foreach loop to `$release_files | ForEach-Object { }` instead.

automatic/python3/update.ps1
```diff
@@ -12,52 +14,146 @@ function global:au_SearchReplace {
...
+              $all_versions[$version]['64'] = $file.url
+          } else {
+              $all_versions[$version]['86'] = $file.url
+          }
+      }
+  }
+
+  # find latest version of each Python minor (3.x) version
+  $latest_versions = @{ }
+  foreach ($version_data in $all_versions.GetEnumerator()) {
```
AdmiringWorm, line 105: We rarely use `foreach` loops in this repository and prefer piped calls.

If possible, change the foreach loop to `$all_versions | ForEach-Object { }` instead.

automatic/python3/update.ps1
```diff
@@ -12,52 +14,146 @@ function global:au_SearchReplace {
...
+      $known_latest_version = Get-Version '0.0.0'
+    }
+    if ($version -gt $known_latest_version) {
+      $latest_versions[$version.Version.Minor] = $version_data.Name
     }
+  }
+
+  # generate two version streams per version, one for python3 and one for python3x package
+  $streams = @{ }
+  foreach ($version_data in $latest_versions.GetEnumerator()) {
```
AdmiringWorm, line 123: We rarely use `foreach` loops in this repository and prefer piped calls.

If possible, change the foreach loop to `$latest_versions | ForEach-Object { }` instead.

automatic/python3/update.ps1
```diff
@@ -12,52 +14,146 @@ function global:au_SearchReplace {
...
-    $versionTwoPart = $version -replace '([\d]+\.[\d]+).*', "`$1"
+  # download Python documentation archive
+  $webrequest = [System.Net.HttpWebRequest]::Create($Latest.ZipUrl)
+  $response_stream = $webrequest.GetResponse().GetResponseStream()
+  $zip = [IO.Compression.ZipArchive]::new($response_stream)
 
-    if ($streams.$versionTwoPart) { return }
+  # license text for legal/LICENSE.txt
+  $license_entry = $zip.GetEntry("$($Latest.ZipName)/license.txt")
+  $Latest.License = [System.IO.StreamReader]::new($license_entry.Open()).ReadToEnd()
```
AdmiringWorm, line 40: Since you are downloading the license and replacing the existing one, we would want to validate the license to ensure it hasn't been changed.

AdmiringWorm (changes requested):
This is great work, a few things I added a comment on though.

Additionally, the thought is to make the `python3` package a meta package that depends on the `major+minor` versions of the packages.

As such, it would be best to have the streams themself as part of a different package (directory), say for instance move it to a `python3-streams` directory, but leave the python3 package itself as is, or source the `update.ps1` file in the streams directory. (See [`7zip.install`](https://github.com/chocolatey-community/chocolatey-packages/blob/master/automatic/7zip.install/update.ps1) and [`7zip`](https://github.com/chocolatey-community/chocolatey-packages/blob/master/automatic/7zip/update.ps1) as an example.)

automatic/python3/update.ps1
jack1142: It does change a few times a year (based on commit history: 6 in 2022, 6 in 2021, 3 in 2020, 5 in 2019, see: https://github.com/python/cpython/commits/main/Doc/license.rst) so this might be an additional maintenance burden when it is well-established that Python will remain under open source license, including software that's been incorporated into Python.

jack1142:
> Additionally, the thought is to make the `python3` package a meta package that depends on the `major+minor` versions of the packages.
> 
> As such, it would be best to have the streams themself as part of a different package (directory), say for instance move it to a `python3-streams` directory, but leave the python3 package itself as is, or source the `update.ps1` file in the streams directory. (See [`7zip.install`](https://github.com/chocolatey-community/chocolatey-packages/blob/master/automatic/7zip.install/update.ps1?rgh-link-date=2022-11-11T09%3A35%3A37Z) and [`7zip`](https://github.com/chocolatey-community/chocolatey-packages/blob/master/automatic/7zip/update.ps1?rgh-link-date=2022-11-11T09%3A35%3A37Z) as an example.)

Alright, I'll try to work on that tomorrow. One problem related to this is that I'm not sure that `Test-RepoPackage` will work properly with `python3x` packages being generated by a package named `python3-streams`, i.e. see the supported cases in the "How Has This Been Tested?" section in #2015 and we'll probably have to decide how to change the logic to work with those kinds of packages as well.

AdmiringWorm:
> Alright, I'll try to work on that tomorrow. One problem related to this is that I'm not sure that `Test-RepoPackage` will work properly with `python3x` packages being generated by a package named `python3-streams`, i.e. see the supported cases in the "How Has This Been Tested?" section in #2015 and we'll probably have to decide how to change the logic to work with those kinds of packages as well.

While the `Test-RepoPackage` may not work properly with the `python3-streams`, it is a known issue with the helper script as it doesn't really have a knowledge about streams at all.

However, I would say feel free to ignore any failures from that script right now. It has been on my TODO list for a few years now to get it sorted, but never had time to do it unfortunately.

Also, feel free to ignore creating the `python3` meta package, that is something I would like to be pushed after we have at least one version of the `python3x` package approved by moderators. We also need to let people know on that package that there will be changes (for instance, passing in package arguments and install arguments won't work anymore since they will not be passed down from the meta package to the `python3x` packages), and recommend them to directly install the major+minor versions that they want instead.

Bottom line, if you can move to the `python3-streams` directory, and remove the pushing of the `python3` package itself this can be considered complete.

jack1142:
> Alright, I'll try to work on that tomorrow.

Welp, that didn't work out... I should get to this some time in the second half of this week as I'm rather busy in the first half. In the meantime, thanks for the additional clarifications, it seems clear to me what I have to do now.

jack1142:
@AdmiringWorm I think I addressed everything that you requested. Commits are a bit of a mess, I can split the changes out between them better if it were to make your review easier.

I also went ahead and reverted #2015 in this PR because it causes tests for `pkgname-streams` packages to fail entirely *and* it is no longer necessary now that this PR no longer tries to use `python3.x` as the package naming scheme.

AdmiringWorm (approved):
LGTM

AdmiringWorm:
@Jackenmen your changes have been merged, thanks for your contribution 👍. Great work.

Before merging I updated the PR to move the python3 package to a meta package, and temporarily disabled python and python3 updates (until we at least have 3.11 sub version available).
Just to be sure that we don't need to change too many packages once it they gets looked at by moderators, I also limited the amount of packages that will be pushed to the first 2 versions found (at this time that is 3.12 and 3.11). That should be removed once we have the first version approved.

Jackenmen:
Looks like the replacement of python3x was done somewhat incorrectly here, PackageSourceUrl and choco install command on description are incorrect :/ I thought I checked, maybe I didn't account for something.

karig:
> The open question that remains is what the package names should be, current state of this PR has `python311`, without a dot rather than `python3.11`. I personally prefer the variant with the dot.

`python311` (without a dot) is a good convention because the default install directory for this minor version is also `C:\Python311` .

`chocolatey upgrade` overwrites the directory and does not remove it. Therefore, a parallel installation of several minor versions is possible and can only be prevented manually (`cuninst` and `cinst` again instead of `cup`). With own package names for minor versions this becomes more obvious.

Issue #2007, opened 2022-10-12
Fix / update packages that use GitHub for version checking


pauby:
### Current Behaviour

GitHub has changed how it accepts web scraping that many packages are using for determining available versions (and other things). As a consequence, updating of those packages are now breaking and becoming outdated.

### Expected Behaviour

To fix this we need to now use the GitHub API, preferably in a PowerShell way with a module such as [PowerShellForGitHub](https://www.powershellgallery.com/packages/PowerShellForGitHub). And to achieve that, many packages need to be refactored. I've already done some of this work on these packages in my repository (I'm linking these as suggested ways I've done things that I hope helps, not the way _we_ or _it_ should be done):

* [Hugo](https://github.com/pauby/ChocoPackages/blob/master/automatic/hugo/update.ps1)
* [Hugo Extended](https://github.com/pauby/ChocoPackages/blob/master/automatic/hugo-extended/update.ps1)
* [Keepass-Plugin-KeeAgent](https://github.com/pauby/ChocoPackages/blob/master/automatic/keepass-plugin-keeagent/update.ps1)

NOTE: There is a [bug in PowerShellForGitHub](https://github.com/microsoft/PowerShellForGitHub/issues/372) that only returns the first 30 release assets. It's been 'fixed' but not sure when it will be merged and released. This may affect the timing of the work for this, which is why I mention it.

### Affected Packages

I would propose we use this issue to collect all the affected packages here, in a checklist, so we can work our way through them. If you don't have permission to edit this post, please add a comment and somebody can make sure it's added here:

* [x] `aptana-studio`
* [x] `bulk-crap-uninstaller`
* [x] `composer`
* [x] `freecad`
* [x] `ghostscript`
* [x] `ghostscript.app`
* [x] `git` (#1995)
* [x] `git.install` (#1995 
* [x] `git.portable` (#1995)
* [x] `git-lfx`
* [x] `gif-lfs.install`
* [x] `intunewinapputil`
* [x] `k9s`
* [x] `kubelogin` ? (this uses the same URL, but the method is different enough that I suspect it should be okay)
* [x] `kubernetes-kompose`
* [x] `minikube` (#1994)
* [x] `minishift`
* [x] `mpc-hc-clsid2`
* [x] `msys2`
* [x] `notepadplusplus`
* [x] `notepadplusplus.install`
* [x] `notepadplusplus.commandline`
* [x] `openshift-cli`
* [x] `paint.net`
* [x] `poi`
[...]

PR #2013, opened 2022-10-15
(#2007) Adds Get-LatestGitHubRelease

JPRuskin:
## Description
Adds a script to address the updates to GitHub Release parsing. 

This has been kept simple, not addressing that wasn't needed for this issue.

## Motivation and Context
As discussed in #2007, the way we were using for scraping GitHub releases in this repository has changed / become unsupported. This function is now available in AU via Update_All (and within the repository). It allows us to simplify update scripts that were previously scraping GitHub releases.

This PR is a prerequisite for the numerous package updates for that issue.

## How Has this Been Tested?
- `Get-LatestGitHubRelease` has been tested against public releases
- `Get-LatestGitHubRelease` has been tested against private releases with a token
- `Get-LatestGitHubRelease` has been used with a PAT hoping to not get rate limited
- Tested using this in Update.ps1 scripts via Update_All.ps1
- Tested with all affected packages in issue #2007 (though with changes to those scripts)

## Types of changes

## Checklist:

- ~[ ] All files are up to date with the latest [Contributing Guidelines](https://github.com/chocolatey-community/chocolatey-packages/blob/master/CONTRIBUTING.md)~
- ~[ ] The added/modified package passed install/uninstall in the chocolatey test environment.~
- ~[ ] The changes only affect a single package (not including meta package).~

### DIFF
--- a/scripts/au_extensions.psm1
+++ b/scripts/au_extensions.psm1
@@ -5,6 +5,7 @@
 # but the file containing the functions is expected
 # to be named using the same name.
 $funcs = @(
+  'Get-GitHubRelease'
   'Set-DescriptionFromReadme'
   'Update-ChangelogVersion'
   'Update-OnETagChanged'
--- a/scripts/Get-GitHubRelease.ps1
+++ b/scripts/Get-GitHubRelease.ps1
@@ -1,0 +1,45 @@
+﻿function Get-GitHubRelease {
+  <#
+    .SYNOPSIS
+      Gets the latest or a specific release of a given GitHub repository
+
+    .EXAMPLE
+      Get-GitHubRelease cloudflare cloudflared
+  #>
+  [CmdletBinding()]
+  param(
+    # Repository owner
+    [Parameter(Mandatory, Position = 0)]
+    [string]$Owner,
+
+    # Repository name
+    [Parameter(Mandatory, Position = 1)]
+    [string]$Name,
+
+    # The Name of the tag to get the relase for. Will default to the latest release.
+    [string]$TagName,
+
+    # GitHub token, used to reduce rate-limiting or access private repositories (needs repo scope)
+    [string]$Token = "$($env:github_api_key)"
+  )
+  end {
+    $apiUrl = "https://api.github.com/repos/$Owner/$Name/releases/latest"
+
+    if ($TagName) {
+      $apiUrl = "https://api.github.com/repos/$Owner/$Name/releases/tags/$TagName"
+    }
+
+    $Request = @{
+      Uri = $apiUrl
+    }
+
+    if (-not [string]::IsNullOrEmpty($Token)) {
+      $Request.Headers = @{
+        Accept        = 'application/vnd.github+json'
+        Authorization = "Bearer $($Token)"
+      }
+    }
+
+    Invoke-RestMethod @Request
+  }
+}


RedBaron2:
@JPRuskin 
Two questions about your function.
1) Will this require a plain text API key be needed in the package update script?
2) The function only checks the latest tag, but what about a specific tag?

You wouldn't happen to know if this is something newly implemented by GitHub that created this issue?
Thanks for any insight.

JPRuskin:
With regards to your first question, it shouldn't require us adding something in plaintext to each script - it follows examples in other update scripts (and the main Update_All.ps1) assuming there's an environment variable specific in our build - I should definitely double check that, though.

On the second, I made the choice to not complicate the function more than it needed to be for the issue it was addressing (I've got a horrible habit of adding everything and the kitchen sink, sometimes, and I'm trying to be better) - hence the name `Get-LatestGitHubRelease`.  
We could certainly extend it, if we found a requirement we wanted to address, or add another function for `Get-GitHubRelease` with a tag / name parameter.

I had heard that GitHub had made some specific change, but haven't found any handy announcement being made - just saw the issue, and the sum of builds using the previous method that started failing in [the update status](https://gist.github.com/choco-bot/a14b1e5bfaf70839b338eb1ab7f8226f).  
I may be incorrect, but it's probably better practice (and I think cleaner to read and maintain) to use the REST API rather than scraping the HTML - and it's certainly fixed the build issue(s) for [all the packages I've got branches for](https://github.com/JPRuskin/chocolatey-coreteampackages/branches/active).

JPRuskin:
We could add further improvement to this by outputting an object with properties we guarantee rather than the direct return, but that feels a bit more than should be required (assuming breaking changes to the API don't happen frequently, which should at least be less frequently than changes to scraping the page)

scripts/Get-LatestGitHubRelease.ps1
```diff
@@ -0,0 +1,35 @@
...
+      [string]$Owner,
+
+      # Repository name
+      [Parameter(Mandatory, Position=1)]
+      [string]$Name,
+
+      # GitHub token, used to reduce rate-limiting or access private repositories (needs repo scope)
+      [string]$Token = "$($env:github_api_key)"
+  )
+  end {
```
AdmiringWorm, line 22: shouldn't this be the following instead?

```suggestion
  process {
```

scripts/Get-LatestGitHubRelease.ps1
```diff
@@ -0,0 +1,35 @@
...
+      $Request = @{
+          Uri = "https://api.github.com/repos/$Owner/$Name/releases/latest"
```
AdmiringWorm, line 24: Not so sure I like this being hard coded for the latest release only.
Any chance to make it allow a tag being specified by the user, but use `latest` by default?

scripts/au_extensions.psm1
```diff
@@ -5,6 +5,7 @@
 # but the file containing the functions is expected
 # to be named using the same name.
 $funcs = @(
+  'Get-LatestGitHubRelease'
```
AdmiringWorm, line 8: If the tag is made configurable, this should be changed to be named `Get-GitHubRelease` instead.

AdmiringWorm (changes requested):

AdmiringWorm:
> You wouldn't happen to know if this is something newly implemented by GitHub that created this issue?

They made changes to how the Artifacts are showing up on the releases page.
Previously it was embedded into the HTML, but was recently changed to be dynamically injected through javascript.

Since we typically use `Invoke-WebRequest -UseBasicParsing`, javascript is not run in this case and the artifacts never show up.

scripts/Get-LatestGitHubRelease.ps1
JPRuskin: I would almost suggest that we have a separate function for getting a specific release tag - though at the moment, we don't have any use for that functionality in this repository (I think?). 

Would that be suited to a follow up PR / issue, where we have a use for it?

RedBaron2: @JPRuskin
FYI: The FreeCAD package uses the tag _weekly-build_

scripts/au_extensions.psm1
RedBaron2: This so happens to be the same naming as one of the current cmdlet from the [PowerShellForGitHub](https://github.com/microsoft/PowerShellForGitHub) module

RedBaron2:
This is true.
> Since we typically use `Invoke-WebRequest -UseBasicParsing`, javascript is not run in this case and the artifacts never show up.

Changes to the cmdlet `Invoke-WebRequest` which makes the switch `UseBasicParsing` a moot point.

> [-UseBasicParsing](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.2#-usebasicparsing)
> This parameter has been deprecated. Beginning with PowerShell 6.0.0, all Web requests use basic parsing only. This parameter is included for backwards compatibility only and any use of it has **no effect** on the operation of the cmdlet.

AdmiringWorm:
> Changes to the cmdlet Invoke-WebRequest which makes the switch UseBasicParsing a moot point.

This typically requires IE to have been executed at least once (maybe edge on Windows 11), as it uses the IE javascript engine. This is typically not the case with CI providers, as well as it takes longer for web requests to happen (from what I remember).
As such, it is not a solution.

Remember, it is also just deprecated for PowerShell 6.0, while we are using Windows PowerShell 5.0 in the repository and the CI.

AdmiringWorm:
In the interest of time, I have updated this PR with the ability to get releases based on the tag name (defaulting to the latest release if no tag name is specified).

I'll get this merged once the appveyor build has completed.

AdmiringWorm:
@JPRuskin your changes have been merged, thanks for your contribution 👍

pauby:
Thank you @JPRuskin and @AdmiringWorm!

PR #2028, opened 2022-10-23
powershell v7 support

tapika:
Added support for powershell v7 with my own fork:

https://github.com/tapika/swupd

Similar merge was done already with following nuget packages:

https://github.com/jberezanski/ChocolateyPackages/pull/138

### DIFF
--- a/automatic/vcredist140/tools/chocolateyInstall.ps1
+++ b/automatic/vcredist140/tools/chocolateyInstall.ps1
@@ -9,7 +9,11 @@ $validExitCodes = @(0, 1638, 3010)
 $force = $Env:chocolateyPackageParameters -like '*Force*'
 
 Write-Verbose 'Checking Service Pack requirements'
-$os = Get-WmiObject -Class Win32_OperatingSystem
+if ($null -ne (Get-Command -Name Get-CimInstance -ErrorAction SilentlyContinue)) {
+    $os = Get-CimInstance -ClassName Win32_OperatingSystem
+} else {
+    $os = Get-WmiObject -Class Win32_OperatingSystem
+}
 $version = [Version]$os.Version
 if ($version -ge [Version]'6.1' -and $version -lt [Version]'6.2' -and $os.ServicePackMajorVersion -lt 1)
 {


automatic/vcredist140/vcredist140.nuspec
```diff
@@ -3,7 +3,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>vcredist140</id>
-    <version>14.32.31332</version>
+    <version>14.32.31332.2</version>
```
AdmiringWorm, line 6: Do not update the version for automatic packages in this repository.

Please revert this change.

AdmiringWorm (changes requested):

pauby:
I'm trying to understand the purpose of adding PowerShell Core support? Chocolatey CLI uses the Windows PowerShell host. If we are adding this support, then we should add it everywhere and not just a few packages.

automatic/vcredist140/vcredist140.nuspec
tapika: Reverted.

tapika:
> I'm trying to understand the purpose of adding PowerShell Core support? Chocolatey CLI uses the Windows PowerShell host. If we are adding this support, then we should add it everywhere and not just a few packages.

In my own [swupd-fork of choco](https://github.com/tapika/swupd) I have upgraded choco to use .net core 3.1, which in a turn required to use newer powershell SDK - to my best understanding it's v7, and change is somehow independent from powershell host, which is still version 5. (I still don't have full understanding of what I'm doing, but at least want to fix installation failure for now).

Concerning which packages this affect - I have already applied this change to one package (See link above), this is second one, and there is also third one.

After all changes gets applied - I will have `choco install dotnetcore` command working, and after that `choco(swupd)` will be able to install .net core 3.1 (with all dependencies) and will be able to perform installation of `choco + choco gui (swupd)`. (wix/msix technologies can be obsoleted after that)

After bootstrap completes, I guess `choco(swupd)` can either install existing choco packages or define newer file format on it's own or even do both. I see that powershell is indeed one approach to have scripting, but it's also possible to replace it with c# scripting capability. There exists [mechanics](https://github.com/tapika/syncProj/blob/master/CsScript.cs) to compile c# code - but I haven't tested them how they will work with readytorun. (How much more disk space it will require)

But maybe one step back, let's start from this change first?

pauby:
I'm still not clear on why adding PowerShell Core support to a packages repository that is intended for a CLI that uses Windows PowerShell as its host is something that we should do?

Having said that, it does sound like your fork of Chocolatey uses PowerShell Core? And it sounds like you want to have packages from this repository be updated to be managed by your fork using PowerShell Core? If that is the case, my suggestion would be to create your own repository for your fork and update those packages that you need, to PowerShell Core.

If we add PowerShell Core support to one package in here, then we're not being consistent. If we add it to 10 packages, we're still not being consistent. Do we go back and update all of the packages? Do we add contributing guidelines that PowerShell Core support should be added to all packages? What about packages that won't work in PowerShell Core?

I appreciate that this is one PR, but it has implications for consistency and ease of use for maintainers adding new and managing the current packages.

tapika:
> Having said that, it does sound like your fork of Chocolatey uses PowerShell Core? And it sounds like you want to have packages from this repository be updated to be managed by your fork using PowerShell Core? If that is the case, my suggestion would be to create your own repository for your fork and update those packages that you need, to PowerShell Core.

I've quickly googled to identify terminology - https://techgenix.com/powershell-core/.
And yes, you're right - my own fork uses powershell core, and not powershell like normal choco does.

You're right in sense that I could indeed duplicate a lot of choco packages on my own choco server.
This most probably would include following packages , as dependency of `dotnetcore`:

```
chocolatey-compatibility.extension
chocolatey-core.extension
chocolatey-windowsupdate.extension
dotnetcore
dotnetcore-3.1-runtime
dotnetcore-runtime
KB2919355
KB2919442
KB2999226
KB3033929
KB3035131
KB3063858
vcredist140
```

but I think simpler solution could be just patch this if-else and avoid package duplication.

I think powershell core is aiming to be portable to other OS's, including linux, which even current `choco` should support.

> If we add PowerShell Core support to one package in here, then we're not being consistent. If we add it to 10 packages, we're still not being consistent. Do we go back and update all of the packages? Do we add contributing guidelines that PowerShell Core support should be added to all packages? What about packages that won't work in PowerShell Core?

I was also thinking what about consistency indeed. 
In theory we could have some common powershell function, which would be somehow centralized and all other packages would call it. Closest point where that powershell function could reside could be for example:

`chocolatey-windowsupdate.extension` package, `Get-OS` function:

https://github.com/jberezanski/ChocolateyPackages/blob/a78144fc0977cfcdd77399071aa1ced125c74577/chocolatey-windowsupdate.extension/extensions/Install-WindowsUpdate.ps1#L175

Currently that function is kinda private, but could be made as public.

However - soon I've realized that this is not only place and not only use case which requires `Get-CimInstance/Get-WmiObject` function switching - there are other places as well, with different parameters.

For me personally for now it's easier just to patch 3 packages then try to patch all packages which `choco` supports.
Also so far I haven't tested `readytorun + powershell core` in empty virtual machine in full scale (for example on .net core 3.1) - I did test some demo package, but suspect windows might require reboots in a middle of installation of all packages. This is something I need to test still.

I propose so make changes to 3 packages first and then later on reconsider other packages as well. 

My own `choco/swupd`-fork is experimental indeed, but before trying to reapply changes to all packages - maybe makes sense to test that my fork will work in smaller scale.

And in long term perspective - sooner or later `choco/swupd`-fork changes can be merged to `choco` master.

pauby:
> And yes, you're right - my own fork uses powershell core, and not powershell like normal choco does.

This repository exists for Chocolatey CLI packages that work with Windows PowerShell. The decision to use PowerShell Core was one that was taken for your project. U understand _why_ it was taken but the doesn't change that I feel asking for changes to packages based on that decision means that the maintainers in this repository should have additional work to do.

I appreciate the changes you are currently asking _seem_ small, but as I mentioned in my previous comment, for things to be consistent we need to make sure those changes are carried forward, so everybody is on the same page. We are only talking about 3 packages just now but when you want to expand that in future you will be talking about more.

> but I think simpler solution could be just patch this if-else and avoid package duplication.

I do understand that it is simpler for swupd to do that. Given what I mentioned above about consistency, it's not simpler for the maintainers here.

> I was also thinking what about consistency indeed.
> In theory we could have some common powershell function, which would be somehow centralized and all other packages > would call it. Closest point where that powershell function could reside could be for example:

I'm not against that. And it could be a good middle ground. Perhaps 'overloading' `Get-WMIObject` (as an example) where the package code stays the same and isn't confusing for anybody else for example and doesn't add any more work on the maintainers here. That could be something that would help you reach your goal. 

> For me personally for now it's easier just to patch 3 packages then try to patch all packages which choco supports.

Again, just to be clear that it is easier for swupd that this is done. But I have to ask you to look at what you are asking for in terms of the work the maintainers here would need to do for consistency. There is no need for PowerShell Core compatible packages here. Any work done on that would be to make things easier for swupd. It's not required for Chocolatey CLI.

I don't feel this is a PR that should be merged in here. There is no need for Chocolatey CLI to have PowerShell Core support in packages. Chocolatey CLI uses Windows PowerShell. I appreciate that those changes would make those packages consumable inside your fork. However, this repository exists for Chocolatey CLI. The additional changes we would need to make to ensure consistency in rules / guidelines for contributors, on those community maintainers reviewing PR's etc. for something that won't be used for Chocolatey CLI doesn't add up in my head.

If you want to have packages that can be used with swupd, then the best suggestion would be for you to create a repository for them. Or as an alternative, the PowerShell module mentioned above.

tapika:
> Or, as an alternative, the PowerShell module mentioned above.

Let's try this one. But I don't own `chocolatey-windowsupdate.extension` - will discuss with package owner first. Let's keep this PR, but I'll reapply last commit. But `chocolatey-windowsupdate.extension` needs to be released and tested first.

tapika:
Managed to fix this on my own. Added additional compatibility module, see issue https://github.com/tapika/swupd/issues/6.
Abandoning pull request as unnecessary.

PR #2046, opened 2022-11-01
Do not create badly-behaved inkscape shim

mueslo:
Shim is installed as GUI shim despite not being set as such, resulting in broken command line behaviour. As far as I can tell there is no way to explicitly install as a non-GUI shim.

But also switching to non-GUI behaviour is not an acceptable solution, since inkscape.exe is also a GUI program depending on the command line arguments.

Therefore no shim should be generated. If people want inkscape.exe in $PATH they can add it manually and chocolatey should not create a broken/semi-functional inkscape.exe there.

See: https://github.com/chocolatey/home/issues/211

Reverts: #1347

I have not removed the `Uninstall-BinFile` line for cleanup purposes.

### DIFF
--- a/automatic/inkscape/tools/chocolateyInstall.ps1
+++ b/automatic/inkscape/tools/chocolateyInstall.ps1
@@ -49,7 +49,6 @@ Get-ChildItem $toolsPath\*.msi | ForEach-Object { Remove-Item $_ -ea 0; if (Test
 $packageName = $packageArgs.packageName
 $installLocation = Get-AppInstallLocation $packageArgs['softwareName']
 if ($installLocation) {
-  Install-BinFile 'inkscape' $installLocation\bin\inkscape.exe
   Write-Host "$packageName installed to '$installLocation'"
 }
 else { Write-Warning "Can't find $PackageName install location" }


pauby:
> If people want inkscape.exe in $PATH they can add it manually 

If inkscape was usable on the path _before_ this PR, then it should be usable _after_ this PR, in the same manner. That's what users are expecting. Changing that and say 'if you want the same behaviour as you had before, you need to do some manual steps' is not acceptable.

I could be missing something and that's not what is happening here?

> chocolatey should not create a broken/semi-functional inkscape.exe there.

Agreed.

mueslo:
Then the only acceptable solution is to add the true inkscape directory to $PATH and not generate a shim.

RedBaron2:
@mueslo 
Actually it is to create the shim with the proper arguments instead of having it use the defaults. This `exe` is unique in that it doesn't behave like other GUI when the shim uses the defaults.

automatic/inkscape/tools/chocolateyInstall.ps1
```diff
@@ -49,7 +49,6 @@ Get-ChildItem $toolsPath\*.msi | ForEach-Object { Remove-Item $_ -ea 0; if (Test
 $packageName = $packageArgs.packageName
 $installLocation = Get-AppInstallLocation $packageArgs['softwareName']
 if ($installLocation) {
-  Install-BinFile 'inkscape' $installLocation\bin\inkscape.exe
```
AdmiringWorm, line 52: If this will be added, the disabling of the inkscape shim must be behind a package parameters flag that the user passes in.

ie, by default a shim gets created, but a user can decide to disable this generation if they want.

AdmiringWorm (changes requested):

Issue #1904, opened 2022-06-08
Migrating Critical Chocolatey Packages To Chocolatey Community


AdmiringWorm:
### Discussed in https://github.com/chocolatey-community/chocolatey-packages/discussions/1898

<div type='discussions-op-text'>

<sup>Originally posted by **pauby** May 30, 2022</sup>
I have a proposal that I wanted to discuss with the Community here to find the best way to approach migrating critical Chocolatey packages to this repository. 

Chocolatey CLI requires specific packages for it's package sources such as`ruby`, `python` and `webpi`. There are also packages that are required for Chocolatey CLI features such as `intunewinappyutil`. And packages required for Chocolatey products such as `nexus` and `jenkins`. Different maintainers are responsible for these packages.

To ensure Chocolatey has a secure, trusted and well-defined package management process, we have had some discussions internally and wanted to understand what the Community thought of:

* Migrating the packages discussed above to this repository.
* The use of the `CODEOWNERS` file ensures that the maintainers best placed to review package changes can do so, providing a trusted and secure process for those critical packages.
* Use the automation that already exists in this repository to ensure packages are up-to-date.

I see the benefits to the Community as

* Members of the Chocolatey Team would be available to help with management of the repository.
* Increased number of package maintainers, able to work on packages and review / merge PR's.
* The migrated packages would be open-source in this repository, allowing others to get involved in helping to resolve problems and add new features

The two important things that I wanted to ensure from this is:

* A secure, trusted and well-defined process for managing critical Chocolatey packages.
* Increase in the number of Community maintainers working in this repository.

I believe this will provide both of those, and I hope that it will encourage other new and experienced maintainers to become involved too. I would really like to see this repository thrive, and I feel this is a good step on that path.

Please let me know your thoughts.</div>

The full list of packages that will be migrated are shown below:

- ~~[ ] webpicmd~~ (this is no longer going to be migrated as the `webpi` source has been removed by Microsoft).
- [x] #1929
- [x] #2065
- [x] #2175
- [x] #2336
- [x] #1871
- [x] #1861

Additionally, there is a need to create a new package that will be part of this as well for the following software:

- [x] #2179

pauby:
Answering the questions from https://github.com/chocolatey-community/chocolatey-packages/pull/2336#issuecomment-1786991828:

> • Replace azcopy package with content of azcopy10 package (and maybe create an azcopy8 package with old content if people still want to use an old version)
[...]

PR #2065, opened 2022-11-15
(nexus-repository) Adds Nexus-Repository

JPRuskin:
## Description
This adds the `nexus-repository` package to the community packages GitHub repository.

## Motivation and Context
This package migrates a "critical Chocolatey Software package" to this community maintained repository.

Adds to #1904

## How Has this Been Tested?
- Tested update script locally, otherwise package has not been changed from previous push

## Types of changes
- ~[ ] New feature (non-breaking change which adds functionality)~

## Checklist:

## Original Location
- Original Repository: Chocolatey Internal Automatic Packages Repository
- [x] Internal: Please see !14
- [x] The [migration guidelines](https://github.com/chocolatey-community/chocolatey-packages/wiki/Package-migration-process) have been followed


automatic/nexus-repository/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,210 @@
+$ErrorActionPreference = 'Stop'
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$version = '[[VERSION]]'
+$url64 = '[[URL]]'
+$checksum64 = '[[CHECKSUM]]'
+$checksumType64 = 'SHA256'
+$validExitCodes = @(0)
+Import-Module (Join-Path $toolsDir 'supportingfunctions.psm1')
+
+$name = $env:Computername
```
steviecoaster, line 10: This is overly complex here. There's nothing inside Nexus after installation from this package that needs to be this strict. Setting `$hostname = 'localhost'` will be sufficient, and should get around issues with the verifier

steviecoaster (changes requested):

steviecoaster (dismissed):
This looks good! I can't speak to the AU stuff much, but it _appears_ correct when compared with other AU scripts here.

automatic/nexus-repository/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,144 @@
...
+Remove-Item "$ExtractFolder" -Force -Recurse
+
+# Install the Nexus service
+$processArgs = @{
+    ExeToRun       = "$TargetFolder\bin\nexus.exe"
+    Statements     = "/install $servicename"
+    validExitCodes = @(0)
+}
+
+$null = Start-ChocolateyProcessAsAdmin @processArgs
```
AdmiringWorm, line 88: Any reason you need to use `Start-ChocolateyProcessAsAdmin` instead of using `Install-ChocololateyInstallPackage`?

automatic/nexus-repository/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    Start-Sleep -Seconds 5
+
+    $ProcessArgs = @{
+        exeToRun       = "$InstallFolder\bin\nexus.exe"
+        Ssatements     = "/uninstall $ServiceName"
+        validExitCodes = @(0)
+    }
+    Start-ChocolateyProcessAsAdmin @ProcessArgs
```
AdmiringWorm, line 21: Any reason not to use `Uninstall-ChocolateyPackage` instead of starting the process itself?

automatic/nexus-repository/update.ps1
```diff
@@ -0,0 +1,24 @@
+#requires -Modules AU
```
AdmiringWorm, line 1:
```suggestion
Import-Module AU
```

AdmiringWorm (changes requested):
A few things to fixup/respond to. Please also run through formatting on the file you are adding, as I noticed some of them don't honor the editorconfig file in this repository.

automatic/nexus-repository/nexus-repository.nuspec
```diff
@@ -0,0 +1,28 @@
...
+    <projectSourceUrl>https://github.com/sonatype/nexus-public</projectSourceUrl>
+    <docsUrl>http://books.sonatype.com/nexus-book/3.0/reference/index.html</docsUrl>
+    <tags>nexus-repository server freeware cross-platform maven nuget npm java rubygems docker eclipse p2 osgi obr apt yum rpm artifact bower ssl respository sonatype nexus admin</tags>
+    <copyright>Sonatype</copyright>
+    <licenseUrl>http://www.eclipse.org/legal/epl-v10.html</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <iconUrl>https://cdn.jsdelivr.net/gh/git@github.com:chocolatey-community/chocolatey-packages/icons/nexus-repository.png</iconUrl>
+    <releaseNotes>https://help.sonatype.com/repomanager3/release-notes</releaseNotes>
+    <dependencies>
+      <dependency id="chocolatey-core.extension" version="[1.1.0,2.0)" />
```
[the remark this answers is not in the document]
JPRuskin: It seemed sensible, if 2.0 introduced breaking changes we'd want to be conscious of upgrading past there. Not a problem for a while, but something to consider.

That said, I can remove it if that's preferred.

automatic/nexus-repository/tools/chocolateyinstall.ps1
JPRuskin: To my knowledge, this file isn't an installer - it's just an argument running on the unpacked binary that triggers the addition of a service. That is, to be fair, a bit of an odd argument now I write it out, but to my mind it makes sense that this is just starting a process rather than installing a thing - the install of the software happens with the `Install-ChocolateyZipPackage`, above.

automatic/nexus-repository/tools/chocolateyuninstall.ps1
```diff
@@ -0,0 +1,25 @@
...
+
+    if (Get-Service $ServiceName -ErrorAction SilentlyContinue){
+        Stop-Service $ServiceName
+    }
+
+    Start-Sleep -Seconds 5
+
+    $ProcessArgs = @{
+        exeToRun       = "$InstallFolder\bin\nexus.exe"
+        Ssatements     = "/uninstall $ServiceName"
```
JPRuskin, line 18:
```suggestion
        Statements     = "/uninstall $ServiceName"
```

JPRuskin: Well, I guess folk don't uninstall Nexus that often?

automatic/nexus-repository/tools/chocolateyuninstall.ps1
JPRuskin: As above, this isn't strictly an installer/uninstaller so it seemed valid to do it this way. We don't want to pass in chocolateyInstallArguments to this call, so I don't think that function adds anything to what we're doing here.

automatic/nexus-repository/nexus-repository.nuspec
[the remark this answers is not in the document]
AdmiringWorm: If it can be removed, then that would be the best.
Unless we use any of the functions (I didn't notice any) provided by the extension, it shouldn't be needed.

If it is still needed, then yeah, remove the upper limit.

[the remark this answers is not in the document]
JPRuskin: I can't find any of the functions I see in the extension in the files, here. I'll remove the dependency!

AdmiringWorm (changes requested):
Missed an item before, there is also a different question that needs to be answered before merging in these changes.

Do we want to create a fix version?
If we do, then additional changes in needed to make that happen, if we don't need to create one then only the file renaming I mention is needed.

automatic/nexus-repository/tools/supportingfunctions.psm1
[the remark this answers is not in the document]
JPRuskin: No requirement for a fix version, in my eyes. Picked one of the methods in use! We should chat about that at some point, maybe -  there's a few advantages to properly using `Import-Module` and a psm1 for helper functions, and I can't think of any compatibility reasons that it wouldn't work?

AdmiringWorm (approved):
LGTM, I'll get this merged once it has completed its AppVeyor run

PR #2104, opened 2023-01-04
(python,python3) enable back meta packages

Jackenmen:
## Description

Another follow-up to #1993. All `python3x` packages are now uploaded to and approved on Chocolatey Community Repository so these can be enabled back.

## Motivation and Context

This was a temporary measure while we were waiting for all `python3x` packages to get pushed to and approved on the community repository.

## How Has this Been Tested?

Dependency-only package, shouldn't really require any testing aside from the update script. The update script ran in CI.

## Types of changes

## Checklist:


Jackenmen:
I'm unsure if this failure is going to be a problem after this gets merged - currently, the packages are forced to be updated and as such, they think there's an update to push for a version that doesn't exist on the community repository.

AdmiringWorm:
The CI failure is not a problem in this case (assuming we can verify the behavior manually).
Extra care must be taken when pushing up fix releases, though (nothing new).

Jackenmen:
Okay, I tested this out and ended up finding and fixing four issues:
- `python` package used one of `python3x` packages instead of `python3` package as dependency
- SearchReplace in the `python3` package missed `]` in the URL markdown after replacing
- `python` package referred to `python3` package as the package where you can specify package parameters rather than `python3x` package
- dependency version in python3 and python packages didn't wrap version in `[]` to ensure exact match

One thing I have not yet solved is an issue with README.md -> `<description>` - in my testing it would appear that it happens too early - README.md gets updated but `<description>` doesn't (but it does get updated with previous stream's README.md when running with a second stream as the replacement does happen, just too soon).

Jackenmen:
> One thing I have not yet solved is an issue with README.md -> `<description>` - in my testing it would appear that it happens too early - README.md gets updated but `<description>` doesn't (but it does get updated with previous stream's README.md when running with a second stream as the replacement does happen, just too soon).

Resolved by using `-NoReadme` and running `Set-DescriptionFromReadme` manually from `au_AfterUpdate` instead. The issue was that the default behavior is to run `Set-DescriptionFromReadme` right after `au_BeforeUpdate` which is before `au_SearchReplace` is run.

This should be ready for review now.

Jackenmen:
I guess it's worth noting that unlike in meta-packages, in the `python3-streams` case we only update references to `python3x` in nuspec while the README just says `python3x` (aside from the README's header which does get updated) which should explain why this was not an issue there.

Jackenmen:
@AdmiringWorm hi! Any chance you have some spare time to review this and get it in? If I did everything right (I sure hope I did, it seems to work fine during my testing), this should be the last piece to this series of PRs for the Python chocolatey packages.

AdmiringWorm (approved):
LGTM

AdmiringWorm:
The AppVeyor failures can be ignored, this is a common problem with the python packages (and meta-packages in general)

PR #2130, opened 2023-01-24
(winpcap) Made compatible with autohotkey version 2

slycordinator:
## Description
Changed Modified automatic/winpcap/tools/chocolateyInstall.ps1:
	* Made packageArgs variable to use file from upstream URL as in existing choco package
	* Changed to run autohotkey uninstall script if the winpcap is already installed (installer crashes if installed over itself w/ 'choco install winpcap --force')

Modified automatic/winpcap/tools/winpcapInstall.ahk:
	* Changed to work with autohotkey version 2
	* Moved repeated actions into functions (like selecting the executable from windows title and hitting enter in menus)
	* Shortened final if/then/else block into a switch/case statement

## Motivation and Context
the winpcap package uses the autohotkey.portable package, which is currently in version 2, which the existing package is incompatible with. Without this change, the package cannot install or even be uninstalled if it were already installed.

* fixes https://github.com/chocolatey-community/chocolatey-packages/issues/2106

## How Has this Been Tested?
I downloaded the official package from chocolatey and then replaced the files in question in the nupkg file.
I tested installing, uninstalling, and reinstalling the package (the latter with 'choco -s . install winpcap --force')

## Types of changes

## Checklist:


automatic/winpcap/tools/chocolateyInstall.ps1
```diff
@@ -1,8 +1,27 @@
 $ErrorActionPreference = 'Stop'
 
+$packageArgs = @{
+	packageName  = 'WinPcap'
+	fileFullPath = "$(Get-PackageCacheLocation)\WinPcapInstall.exe"
+	url          = 'https://www.winpcap.org/install/bin/WinPcap_4_1_3.exe'
```
pauby, line 6: Can we look to embed the binary in the package? This binary is no longer updated, so it will stop it disappearing from the web and the package being unusable.

pauby (commented):
It looks like, on the surface, that the `chocolateyUninstall.ps1` may be broken?

Was this package tested in the Chocolatey Test Environment?

pauby:
I think we should go ahead and remove this package from being automatically updated. It's been abandoned.

slycordinator:
If you notice, the package here in the repo was changed to be embedded a
few years ago, but the changes were never pushed to chocolatey. I reverted
that because it seemed best to have it be similar to what's in chocolatey.

Perhaps it's better to have it be embedded

automatic/winpcap/tools/chocolateyInstall.ps1
```diff
@@ -1,8 +1,27 @@
 $ErrorActionPreference = 'Stop'
 
+$packageArgs = @{
+	packageName  = 'WinPcap'
+	fileFullPath = "$(Get-PackageCacheLocation)\WinPcapInstall.exe"
```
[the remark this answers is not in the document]
slycordinator: I've changed it to not do this.

I've modified the files in the branch to embed the installer, so the above no longer is necessary.

automatic/winpcap/tools/chocolateyInstall.ps1
slycordinator: I've changed it to embed the installer.

slycordinator: Though, I didn't include the file itself in the PR/commits, since for other packages that embed it, they were not included in this repo either.

slycordinator:
> It looks like, on the surface, that the `chocolateyUninstall.ps1` may be broken?
> 
> Was this package tested in the Chocolatey Test Environment?

I didn't test it using the Chocolatey Test Environment that is shared in this repo. It wasn't clear in the above checklist that it's the testing environment that's referenced and the PC I've been working on this on doesn't meet the requirements.

I erred on using the uninstaller powershell script as-is, because it's as was committed here multiple years ago and worked perfectly uninstalling the package on my computer.

Though, it's again perhaps of note that the package in chocolately itself uses a different uninstall powershell script, as the changes present here on this repo never got pushed.

In the choco package, the chocolateyUninstall.ps1 file looks like:
`Write-Output "Running Autohotkey uninstaller"
$toolsPath = Split-Path $MyInvocation.MyCommand.Definition
$ahkScript = "$toolsPath\winpcapInstall.ahk"
AutoHotkey $ahkScript $packageArgs.fileFullPath`

slycordinator:
I was able to test it in the testing environment and it worked fine.

I created the package from my repo, downloaded the upstream installer,
placed it in the package, and in the test environment it worked perfectly.

I tested it there to:
1) install perfectly when not installed already
2) install perfectly when forcing to install for a reinstall
3) uninstall perfectly

Though, one caveat is that it can't be uninstalled using "choco.exe install
-x winpcap", because that removes all dependencies before doing the
uninstall (and it needs AHK to remain installed).

pauby (changes requested):
This is looking good! Thanks for doing this.

One final thing, can you remove the changes you've made to the VERIFICATION.txt file as those changes are made by the update.ps1 file.

slycordinator:
> This is looking good! Thanks for doing this.
> 
> One final thing, can you remove the changes you've made to the VERIFICATION.txt file as those changes are made by the update.ps1 file.

Done

slycordinator:
This still says that "changes" remain that were requested, but the only change was the request to revert VERIFICATION.TXT back to the way it was originally, which was done in March. Is there more that you would like?

pauby:
This works for me in the Chocolatey Test Environment. A couple of things, though:

1. Can you update the version of `autohotkey.portable` to 2.0.0 as a minimum?
2. Can you move the whole package to the `deprecated\packages` folder?
3. Can you [deprecate this package](https://docs.chocolatey.org/en-us/community-repository/maintainers/deprecate-a-chocolatey-package#the-solution) as it is no longer actively maintained?

pauby (approved):
I've updated this PR in line with this comment. Note that I haven't deprecated it, rather added `(Final Version - EOL)` to the title.

PR #2164, opened 2023-03-02
(#2134) (transifex-cli) Add new transifex-cli package

### LINKED ISSUE #2134: Add Transifex CLI as a package
### Discussed in https://github.com/chocolatey-community/chocolatey-packages/discussions/2133

<div type='discussions-op-text'>

<sup>Originally posted by **pauby** January 31, 2023</sup>
Can we add the [Transifex CLI](https://github.com/transifex/cli) in here? This is the new 'Transifex Client' that we currently have in this repository, so it makes sense to add it.

The Transifex Client has been deprecated in favour of this, so there is work to do there too.</div>

vexx32:
## Description

- Added automatically-updatable transifex-cli embedded package for the replacement CLI tool from transifex-client
- Added @chocolatey-community/chocolatey-team-maintainers to the CODEOWNERS for this and the other transifex package (already listed on both packages' owner fields as well).

## Motivation and Context

fixes #2134 

~I haven't done the package deprecation for the older package yet as I'm not that familiar with how we typically do those and whether it should be a separate PR or not.~

## How Has this Been Tested?

- Ran the added `update.ps1` file, verified that the new `1.6.5` version was detected and a new package was generated
- Verified the changes to VERIFICATION.txt and the nuspec file applied correctly, links and checksums were updated
- Installed the package locally, verified that choco generated appropriate shims automatically
- Reran install in the test env, didn't see any issues there either

N/A

## Types of changes

- [x] New package to replace older existing package (transifex-client)

## Checklist:


pauby (changes requested):
Can you follow the format / order of the update.ps1 file that we use in the repository? See the vim package as an example.

Can you also confirm how this is tested because I can't see how the update.ps1 file would work?

automatic/transifex-cli/update.ps1
```diff
@@ -0,0 +1,55 @@
...
+function global:au_SearchReplace {
+  @{
+    '.\legal\VERIFICATION.txt'        = @{
+      '(?i)(^\s*location on\:?\s*)\<.*\>' = "`$1<$($Latest.ReleaseUrl)>"
+      '(?i)(\s*64\-Bit Software.*)\<.*\>' = "`$1<$($Latest.URL64)>"
+      '(?i)(\s*32\-Bit Software.*)\<.*\>' = "`$1<$($Latest.URL32)>"
+      '(?i)(^\s*checksum32\:).*'          = "`$1 $($Latest.Checksum32)"
+      '(?i)(^\s*checksum64\:).*'          = "`$1 $($Latest.Checksum64)"
+      '(?i)(^\s*checksum\s*type\:).*'     = "`$1 $($Latest.ChecksumType64)"
+    }
```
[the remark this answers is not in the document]
vexx32: ah, hmm, that's a good point. I was relying on AU there but it may have the wrong checksums... Hmm. I'll update this, then.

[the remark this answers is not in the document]
vexx32: Also, fun fact, the checksums that they provide in their GH releases are just for the zip files, not the contained executables. OK, I guess...

vexx32:
@pauby I've updated things here according to your comments. For testing I ran `update.ps1` and verified that:

1. The files that needed to be updated (nuspec, VERIFICATION.txt) are updated
2. The files in the package that is generated are the updated copies of the files
3. The executables in the package match the checksums that AU put in the VERIFICATION.txt

automatic/transifex-cli/update.ps1
```diff
@@ -0,0 +1,62 @@
...
+
+    $zip = "$PSScriptRoot\tools\$file"
+    Expand-Archive -Path $zip -Destination $folder
+    $checksum = Get-ChildItem -Path $folder -Recurse -Filter 'tx.exe' | Get-FileHash -Algorithm SHA256
+
+    $Latest."Checksum${architecture}" = $checksum.Hash
+
+    Get-ChildItem -Path $folder -File -Recurse -Exclude "*.exe" | Remove-Item -Force
+    Remove-Item $zip -Force
+  }
```
AdmiringWorm, line 48: This feels a bit wrong to me, I kinda would like to have seen the archive themself being embedded in the package, and install have an `install` script that extracted the file from it instead.

Thoughts @pauby

pauby: I would have preferred that too because we are shipping what the release is and it's much easier for people to see that. It was just my opinion and nothing to back it up beyond my experience.

If you are of the same opinion though, let's ship what is released and do the extraction work in the `chocolateyInstall.ps1`.

automatic/transifex-cli/Readme.md
```diff
@@ -0,0 +1,11 @@
+# [<img src="https://cdn.jsdelivr.net/gh/chocolatey-community/chocolatey-packages@4c2d7c7377b05a3d0a2dd35cfc89ce0c77625051/icons/transifex-client.png" height="48" width="48" /> Transifex Client](https://chocolatey.org/packages/transifex-client)
+
+The Transifex CLI enables you to manage your translations within a project without the need of an elaborate UI system.
+
+You can use the command line tool to create new resources, map locale files to translations, and synchronize your Transifex project with your local repository.
+Translators and localization managers can use it to handle large volumes of translation files.
+The Transifex CLI can help to enable continuous integration workflows and can be run from CI servers like Jenkins and Bamboo.
+
```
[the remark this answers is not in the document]
pauby: As other changes are being requested, adding it shouldn't be much more than a cut and paste from the software site.

If it smore than that, I'd go with what Kim has suggested and we can update it after the fact.

automatic/transifex-cli/update.ps1
AdmiringWorm: I agree, so we should extract the files in the `chocolateyInstall.ps1` script.

While there may be some rare cases we want to extract the files before creating the package, I don't believe this package is one of the cases where this is needed.

automatic/transifex-cli/legal/LICENSE.txt
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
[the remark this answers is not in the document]
AdmiringWorm: If we are going forward with including where the license is from in this file, we'll need to remove it from the `Verification.txt` file, as the only reason it is included there is that we never included the location here in the license file in the past.

AdmiringWorm:
> I haven't done the package deprecation for the older package yet as I'm not that familiar with how we typically do those and whether it should be a separate PR or not.

It is quite rare that we have added a new package and deprecated an older package in the past (I honestly don't remember the last time). However, as we are adding a new package that supersedes the old one it should be fine to include the deprecation of the old package in this PR.

We still need to do a manual update of the deprecated package once this version is approved on Chocolatey Community Repository, but if we do the deprecation at the same time, we ensure that a new version of the deprecated package wouldn't be picked up even if there would have been a new release.

I am not to fuzzed if you include the deprecation in this PR or just open a new PR for the deprecation though.

In general, deprecating the package should be easy.

1. Move the existing package to `deprecated\packages\transifex-client`
2. Remove all files except the `nuspec`
3. Update the title to have the `[Deprecated]` prefix
4. Remove the existing description, and replace it with a mention to the new package as well as the reason the package is deprecated
5. Add a dependency on the new package
6. Change the `<files>` element to be a `<files />` self closing element instead
7. Remove the icon URL element
8. Update the version to be a fix version (ie, if the current version is `0.18.0`, it should be replaced with `0.18.0.20230307`) (Use the date you add the commit for the change)
9. Look over the normal deprecation documentation in case I missed something (https://docs.chocolatey.org/en-us/community-repository/maintainers/deprecate-a-chocolatey-package)

Once everything is done, and the new package is approved well need to just push the package using a new empty commit with the text `[PUSH transifex-client]`. (Note, unlike the what the deprecation docs say, we normally don't unlist old versions).

automatic/transifex-cli/Readme.md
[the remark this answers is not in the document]
vexx32: Yeah I'd love to, unfortunately the doc I've linked seems to be _all_ they really list about it. There are features mentioned throughout, but really it's just usage/reference doc and not much else.

Unsure what exactly we'd put in here tbh.

automatic/transifex-cli/update.ps1
vexx32: That's fair, I'll look to get this updated for that instead.

automatic/transifex-cli/Readme.md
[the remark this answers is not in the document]
vexx32: Opened https://github.com/chocolatey-community/chocolatey-packages/discussions/2167

automatic/transifex-cli/update.ps1
vexx32: I've changed this around to just download the zip files and have the install script figure out which one to extract based on architecture. Tested the install seems to work, shimgen generates a shim for the extracted tx.exe as normal.

automatic/transifex-cli/legal/LICENSE.txt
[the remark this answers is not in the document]
AdmiringWorm: After a quick chat with @pauby regarding including `From` in the license text file, and instead have this information in the verification.txt file instead.

However, @pauby correct me if I got some of that wrong or misunderstood.

vexx32:
@pauby @AdmiringWorm I've addressed the remaining comments (I think) and have added an additional commit to deprecate the original package.

deprecated/packages/transifex-client/transifex-client.nuspec
```diff
@@ -18,18 +18,11 @@
     <tags>transifex-client transifex translation cli cross-platform foss</tags>
```
AdmiringWorm, line 18: I didn't mention it (mainly because I forgot, and we don't have any documentation of it).

But would it make sense if we replace all of the tags to just be `deprecated`?

Thoughs @pauby @vexx32

deprecated/packages/transifex-client/transifex-client.nuspec
```diff
@@ -18,18 +18,11 @@
...
-
-You can use the command line tool to create new resources, map locale files to translations, and synchronize your Transifex project with your local repository. Translators and localization managers can use it to handle large volumes of translation files. The Transifex Command-line Tool can help to enable continuous integration workflows and can be run from CI servers like Jenkins and Bamboo.
-
-# Notes
-- This package uses the Python 3.x version of the transifex-client
-- This package only provides the 64-bit executable due to a number of positives on VirusTotal.
-  The Software Authors believe them to be false positives, but I as the package maintainer am not convinced.
-]]></description>
+    <description><![CDATA[This package is deprecated as the upstream maintainer has deprecated the Python verson of the Transifex client along with the API endpoints it used.
+    Please use the replacement transifex-cli package going forward.]]></description>
```
AdmiringWorm, line 22: Just a suggestion based on what we have done for a previous package we deprecated.

```suggestion
    <description><![CDATA[### New package name: **[transifex-cli](https://community.chocolatey.org/packages/transifex-cli)**
    
--

`transifex-client` has been deprecated as the upstream maintainer has deprecated the Python version of the Transifex client along with the API endpoints it used.

It exists only for backward compatibility to provide a transitional dummy package for existing users and scripts which may still use this package ID.]]></description>
```

deprecated/packages/transifex-client/transifex-client.nuspec
```diff
@@ -18,18 +18,11 @@
...
     <releaseNotes>https://github.com/transifex/transifex-client/releases/tag/0.14.4</releaseNotes>
```
AdmiringWorm, line 23: Suggestion based on what we have done when previously deprecating a package.

```suggestion
    <releaseNotes>#### New users

* Please don't install this package. Install **transifex-cli** instead.

#### Existing users - Updating

* Updating this package will automatically install the most recent version of the **transifex-cli** chocolatey package.
* Please update any links, dependencies, or scripts that reference this package to the new package name **transifex-cli**

#### Existing users - Post-update Cleanup (optional)

`choco uninstall transifex-cli --whatif`</releaseNotes>
```

automatic/transifex-cli/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,21 @@
...
+  packageName = 'transifex-cli'
+  destination = $toolsDir
+  file        = if ((Get-OSArchitectureWidth 64) -and $env:chocolateyForceX86 -ne $true) {
+    Write-Host "Installing 64 bit version"
+    $filePath64
+  }
+  else {
+    Write-Host "Installing 32 bit version"
+    $filePath32
+  }
```
AdmiringWorm, line 18: Since the earliest version of Chocolatey CLI we support in this repository right now is `0.12.1` there is no need for this conditional anymore. The need for such a conditional was needed in versions older than `0.10.4`

```suggestion
  file        = "$toolsDir\tx-windows-386.zip"
  file64      = "$toolsDir\tx-windows-amd64.zip"
```

This will require an update/adding to the regex used in `update.ps1` to ensure we always replace the files with the correct archive names.

See [selenium-gecko-driver](https://github.com/chocolatey-community/chocolatey-packages/blob/a91e4270569b0af6fb6a8b37ec2751de6fa9deef/automatic/selenium-gecko-driver/update.ps1#L18-L19) For the preferred regex to use.

automatic/transifex-cli/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,21 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$toolsDir = $(Split-Path -Parent $MyInvocation.MyCommand.Definition)
+
+$filePath32 = "$toolsDir\tx-windows-386.zip"
+$filePath64 = "$toolsDir\tx-windows-amd64.zip"
+
+$packageArgs = @{
+  packageName = 'transifex-cli'
```
AdmiringWorm, line 9:
```suggestion
  packageName = $env:ChocolateyPackageName
```

automatic/transifex-cli/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+## Notes
+
+- This package uses the Go version of the Transifex CLI
```
AdmiringWorm, line 11: @pauby do you have an example of the notes you usually prefer to have added in this section to direct the user to where to open issues regarding bugs in the package?

deprecated/packages/transifex-client/transifex-client.nuspec
```diff
@@ -3,10 +3,10 @@
...
     <id>transifex-client</id>
-    <version>0.14.4</version>
+    <version>0.14.4.20230307</version>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/transifex-client</packageSourceUrl>
     <owners>chocolatey-community,AdmiringWorm</owners>
-    <title>Transifex Command-Line Tool</title>
+    <title>[Deprecated] Transifex Command-Line Tool</title>
     <authors>The Transifex Team</authors>
     <projectUrl>https://www.transifex.com/</projectUrl>
     <iconUrl>https://cdn.jsdelivr.net/gh/chocolatey-community/chocolatey-packages@4c2d7c7377b05a3d0a2dd35cfc89ce0c77625051/icons/transifex-client.png</iconUrl>
```
AdmiringWorm, line 12: The icon URL needs to be removed.

automatic/transifex-cli/Readme.md
```diff
@@ -0,0 +1,11 @@
+# [<img src="https://cdn.jsdelivr.net/gh/chocolatey-community/chocolatey-packages@4c2d7c7377b05a3d0a2dd35cfc89ce0c77625051/icons/transifex-client.png" height="48" width="48" /> Transifex Client](https://chocolatey.org/packages/transifex-client)
```
AdmiringWorm, line 1:
```suggestion
# [<img src="https://cdn.jsdelivr.net/gh/chocolatey-community/chocolatey-packages@4c2d7c7377b05a3d0a2dd35cfc89ce0c77625051/icons/transifex-cli.png" height="48" width="48" /> Transifex CLI](https://chocolatey.org/packages/transifex-cli)
```

Please also rename the existing icon for `transifex-client` to `transifex-cli` (assuming it is still the same icon being used by the software, if it isn't, then remove the icon and extract the new icon from the software using `becyicongrabber` or a similar program).

automatic/transifex-cli/transifex-cli.nuspec
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>transifex-cli</id>
+    <version>1.6.4</version>
```
AdmiringWorm, line 6:
```suggestion
    <version>0.0.0</version>
```

don't set the latest version in the nuspec, this will actually prevent any automatic update from happening.

AdmiringWorm (changes requested):

deprecated/packages/transifex-client/transifex-client.nuspec
vexx32: No idea tbh. I guess it kind of makes sense that we'd want to reduce the discoverability of the package? Not sure.

automatic/transifex-cli/tools/chocolateyInstall.ps1
vexx32: The archive names don't change from release to release from what I can tell?

automatic/transifex-cli/Readme.md
vexx32: The new version of the CLI doesn't actually have any icon as far as I can tell, so I just used the old one for now 🤷‍♀️

vexx32: Also, if I rename this file, won't the link be invalid since it contains a commit hash?

deprecated/packages/transifex-client/transifex-client.nuspec
AdmiringWorm: I am honestly not sure if this should be as part of the release notes, or the description. I am not too fuzzed where it is added, but either in the description or in the release notes.

automatic/transifex-cli/Readme.md
AdmiringWorm: That is exactly why we use commit hashes, so it won't be invalid. It would be invalid if we didn't use commit hashes.

automatic/transifex-cli/tools/chocolateyInstall.ps1
vexx32: Updated

deprecated/packages/transifex-client/transifex-client.nuspec
vexx32: Done, thanks!

automatic/transifex-cli/Readme.md
vexx32: I believe I've updated this one and the link seems to work, so we should be good.

automatic/transifex-cli/transifex-cli.nuspec
vexx32: Done, thanks!

deprecated/packages/transifex-client/transifex-client.nuspec
vexx32: Updated, thanks!

deprecated/packages/transifex-client/transifex-client.nuspec
vexx32: Updated this one too

deprecated/packages/transifex-client/transifex-client.nuspec
vexx32: After thinking for a bit I think this makes sense, I've updated this too.

pauby (approved):

pauby:
I tested the build, install and uninstall of the package in the Chocolatey Test Environment and it all worked.

Issue #1881, opened 2022-05-09
(filezilla) File links are invalid


pauby:
### Discussed in https://github.com/chocolatey-community/chocolatey-packages/discussions/1880

<div type='discussions-op-text'>

<sup>Originally posted by **uwemolnar** May  9, 2022</sup>
As requested by the maintainer of filezilla: the installer-links for filezilla seem to be outdated. Can you please check and fix them:
https://github.com/chocolatey-community/chocolatey-packages/blob/master/automatic/filezilla/legal/VERIFICATION.txt

Would be great - thanks!</div>

pauby:
@AdmiringWorm As the author seems to be going out of their way to make this difficult to package (assuming I understand this correctly) if this is a package we want to stop maintaining here? I'm happy to raise an RFM request for it.

pauby:
Submitted PR #2166 to fix this.

pauby:
Sure.

michael-franck:
Submitted PR #2316 to fix this.

PR #2166, opened 2023-03-04
(filezilla)(#1881) Fix verification URLS

pauby:
## Description
Added the download page to the VERIFICATION.txt file.

## Motivation and Context
The download links for Filezilla use a unique value that is timed. Putting them into the VERIFCATION.txt file means that after the time has run out, the links don't work. Using the download page links instead.

## How Has this Been Tested?
This is a simple text file change so hasn't been tested.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/filezilla/update.ps1
+++ b/automatic/filezilla/update.ps1
@@ -9,8 +9,6 @@ function global:au_SearchReplace {
         }
 
         ".\legal\VERIFICATION.txt" = @{
-          "(?i)(\s+x32:).*"            = "`${1} $($Latest.URL32)"
-          "(?i)(\s+x64:).*"            = "`${1} $($Latest.URL64)"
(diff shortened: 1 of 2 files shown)


automatic/filezilla/update.ps1
```diff
@@ -9,8 +9,6 @@ function global:au_SearchReplace {
         }
 
         ".\legal\VERIFICATION.txt" = @{
-          "(?i)(\s+x32:).*"            = "`${1} $($Latest.URL32)"
-          "(?i)(\s+x64:).*"            = "`${1} $($Latest.URL64)"
```
AdmiringWorm, line 13: Instead of removing these replacements, can you swap out the replacement to use replace any URL with the URL specified in the `$releases` variable?

This way, if we change this variable in the future we will also have the new download location.

AdmiringWorm (changes requested):

PR #2236, opened 2023-06-06
(selenium.powershell) Adds selenium.powershell

### LINKED ISSUE #2179: Create Selenium PowerShell Module package
This Is split off from the main issue of to its own separate issue.

As the PowerShell module is something we use ourself, we need to create and own a package for this module.

JPRuskin:
## Description
This new package contains the Selenium module.

It is updated from the PowerShell gallery releases, and there are not currently plans to run a prerelease stream.

Please note that there is some outstanding conversation about the correct category tag to use, and that the update script is _interesting_ in that it adds verification for each and every file within the module. This doesn't fit nicely with any of our examples, as this is the first PowerShell module to be added to this repository.

We should (probably) resolve the conversation around tags before merging this package, so the PR has been created as a draft.

## Motivation and Context
As part of #1904, it was decided to migrate packages we depend on to the Chocolatey Community Chocolatey Packages Git repository. This package was included as wanted, but wasn't actually a pre-existing package.

Fixes #2179.

## How Has this Been Tested?
- Pack package / run Update.ps1 / Update_all.ps1 selenium.powershell
- Tested installation of package without arguments
- Tested uninstallation of package
- Tested install of package with arguments (both individually and together)
- Observed expected results
- Tested uninstall with both arguments

## Types of changes

## Checklist:


JPRuskin:
As folk in the meeting agreed on the `addon` tag, this is ready for review.

pauby:
If you try and uninstall the package, while the module is loaded, you get this error:

```
ERROR: The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: Cannot remove item C:\Program Files\WindowsPowerShell\Modules\Selenium\3.0.1.20230824\assemblies\WebDriver.dll: Access to the path 'WebDriver.dll' is denied.
selenium.powershell uninstall not successful.
```

pauby:
For something like this, I'd be looking to find out what is locking the process and taking actions from there. There is no one size fits all for a lot of this stuff. It could be something as easy as catching it, outputting a warning that the module could not be unloaded and that they will need to remove it manually may be enough. Or it could be it's always opened under a process called 'blah' and that you can simply terminate that, and it will be good.

I think it's important that the package uninstalls correctly. I feel a maintainer should put effort into removing the software too. But sometimes it's not possible, and the focus should be telling the user under these sorts of circumstances. That's the least bad user experience, _in my opinion_.

pauby:
@JPRuskin Can you provide an update on this?

pauby:
This has had no updates in some time now. I'm going to go ahead and close this, but we can reopen it if necessary.

PR #2246, opened 2023-06-25
(autohotkey.install)(#1710) Add AutoHotkey v2 with v1 compatibility

### LINKED ISSUE #1710: (autohotkey) Create virtual, v1 and v2 packages
AHK v2 is dramatically different from AHK v1. Given the importance of this package for unattended installations, I propose the following:

1. Current ID `autohotkey` should become virtual package
2. Rename current autohotkey to autohotkey1
3. Create new package, autohotkey2
4. Via streams, make autohotkey to contain all releases from both versions.

Example for this methodology is done once by me in package PostgreSql and seems to work fine as one can choose major version but also choose not to care and entire thing is easily discoverable.

Alternatively, we could leave autohotkey as it is, and create set of new packages with different names: ahk, ahk1, ahk2 and deprecate autohotkey to target ahk1. Seems more clean and I like shorter name.

Also, this would have to handle portable variants. So its some work, but it could be done as single package with 5 streams.

If done, this would make #1663 and its PR obsolete.

Thoughts ?

pauby:
## Description
Adds AutoHotkey v2 with v1 compatibility.

## Motivation and Context
AutoHotkey v1.1 has been deprecated and v2 has been out for some time. There is concern around upgrading to v2 and v1 scripts failing. There are a lot of v1 script still on CCR.

I don't feel this is a breaking change (as functionality has not changed) so have not ticked that below.

Closes #1710 

## How Has this Been Tested?
It was tested in the Chocolatey Test Environment with the [Logitech Presentation](https://community.chocolatey.org/packages/logitech-presentation) package that has an AutohotKey v1 uninstall script. 

## Types of changes

## Checklist:

### DIFF
--- a/automatic/autohotkey.install/tools/chocolateyInstall.ps1
+++ b/automatic/autohotkey.install/tools/chocolateyInstall.ps1
@@ -1,29 +1,53 @@
 ﻿$ErrorActionPreference = 'Stop'
 
 $toolsDir = Split-Path $MyInvocation.MyCommand.Definition
+$installerv1File = 'AutoHotkey_1.1.36.02_setup.exe'
+$installerv2File = 'AutoHotkey_2.0.11_setup.exe'
 
 $pp = Get-PackageParameters
 if (!$pp.DefaultVer){
   $pp.DefaultVer = if ((Get-OSArchitectureWidth 64) -and ($Env:chocolateyForceX86 -ne 'true')) { 'U64' } else { 'U32' }
 }
 
-$packageArgs = @{
+$packageArgsv2 = @{
   packageName    = 'autohotkey.install'
   fileType       = 'exe'
-  file           = Get-Item "$toolsDir\*.exe"
-  silentArgs     = "/S /$($pp.DefaultVer)"
+  file           = Join-Path -Path $toolsDir -ChildPath $installerv2File
+  silentArgs     = "/silent"
   softwareName   = 'AutoHotkey*'
   validExitCodes = @(0, 1223)
 }
-Install-ChocolateyInstallPackage @packageArgs
-Remove-Item $toolsDir\*.exe
 
-$installLocation = Get-AppInstallLocation $packageArgs.softwareName
-$packageName = $packageArgs.softwareName
-if ($installLocation)  {
-    $installName = 'AutoHotkey'
-    Write-Host "$packageName installed to '$installLocation'"
-    Register-Application "$installLocation\$installName.exe"
-    Write-Host "$packageName registered as $installName"
+Install-ChocolateyInstallPackage @packageArgsv2
+
+$installLocation = Get-AppInstallLocation $packageArgsv2.softwareName
+$packageName = $packageArgsv2.softwareName
+$installName = 'AutoHotkey'
+$exePath = Join-Path -Path $installLocation -ChildPath "v2\$installName.exe"
+if ($installLocation) {
+  Write-Host "$packageName installed to '$installLocation'"
+  Register-Application -ExePath $exePath
+  Write-Host "$packageName registered as $installName"
 }
-else { Write-Warning "Can't find $packageName install location" }
+else {
+  Write-Warning "Can't find $packageName install location"
+}
+
+$packageArgsv1 = @{
+  packageName    = 'autohotkey.install'
+  fileType       = 'exe'
+  file           = Join-Path -Path $toolsDir -ChildPath $installerv1File
+  silentArgs     = "/S /$($pp.DefaultVer) /install"
+  softwareName   = 'AutoHotkey*'
+  validExitCodes = @(0, 1223)
+}
+
+Install-ChocolateyInstallPackage @packageArgsv1
+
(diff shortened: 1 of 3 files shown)


pauby:
@AdmiringWorm I have set this to draft just now as I have not updated the `update.ps1` as yet. My thoughts are to not update the package automatically if a new v1 version is released but stick strictly to the v2 numbering. It is possible, but unlikely, that a v1 version is released as it has been deprecated.

pauby:
This should be ready for review.

pauby:
I'm unsure why AppVeyor is failing.

The package was tested on the Chocolatey Test Environment. I can also see both `packageName` fields have string values (so no chance of an empty variable).

So I'm confused.

automatic/autohotkey.install/tools/chocolateyInstall.ps1
```diff
@@ -10,12 +12,44 @@ if (!$pp.DefaultVer){
...
-  file           = Get-Item "$toolsDir\*.exe"
-  silentArgs     = "/S /$($pp.DefaultVer)"
+  file           = Join-Path -Path $toolsDir -ChildPath $installerv2File
+  silentArgs     = "/silent"
   softwareName   = 'AutoHotkey*'
   validExitCodes = @(0, 1223)
 }
-Install-ChocolateyInstallPackage @packageArgs
+
+Install-ChocolateyInstallPackage @packageArgsv2
```
AdmiringWorm, line 21: This one is the cause of the failure.

You are attempting to use the variable `packageArgsv2`, however it is still called `packageArgs` above.

pauby: I've pushed a fixed version.

AdmiringWorm (dismissed):
While I still don't like the idea of mixing v1 and v2 into the same package, I understand the reason for it. As such I'll approve this PR now.

@pauby as there are merge conflicts, could you rebase the PR to fix those, then we can get this PR merged in?

pauby:
@AdmiringWorm this has been fixed up and rebased.

automatic/autohotkey.install/tools/chocolateyInstall.ps1
```diff
@@ -1,21 +1,55 @@
...
+  Write-Host "$packageName installed to '$installLocation'"
+  Register-Application -ExePath $exePath
+  Write-Host "$packageName registered as $installName"
+}
+else {
+  Write-Warning "Can't find $packageName install location"
+}
+
+$packageArgsv1 = @{
+  packageName    = 'autohotkey.install'
```
AdmiringWorm, line 37: Looking at the output on AppVeyor, do wme maybe want to change the packageName here so it is identifying that it is installing v1 of autohotkey?

automatic/autohotkey.install/tools/chocolateyInstall.ps1
```diff
@@ -1,21 +1,55 @@
...
+
+# This is run to ensure that version 2.x is shown in Programs and Features and registered with the system.
+# See https://www.autohotkey.com/docs/v2/Program.htm#install_v1
+# The /silent switch was found by looking at the ux\install.ahk script
+$ahkInstall = Join-Path -Path $installLocation -ChildPath 'ux\install.ahk'
+Start-ChocolateyProcessAsAdmin -ExeToRun $exePath -Statements """$ahkInstall"" /silent"
+
 Remove-Item $toolsDir\*.exe
 
 $installLocation = Get-AppInstallLocation $packageArgs.softwareName
```
AdmiringWorm, line 55: This line, and the line below uses the wrong argument name.

These should probably use `$packageArgsv1` instead (can't make a suggestion due to the below line is out of range).

AdmiringWorm (changes requested):

automatic/autohotkey.install/tools/chocolateyInstall.ps1
pauby: This actually comes from not updating your branch with upstream changes from a computer that doesn't have them, then updating with the small change, force pushing and overwriting all of your changes.

I think it's called being an idiot.

But I panicked and fixed it. It's now passing AppVeyor.

AdmiringWorm (approved):
LGTM, there is a need to fix some URLs as well before we attempt to push this. I'll get that updated when I merge this PR

asheroto:
FYI, v2 will not execute all v1 scripts. After upgrading to v2, it broke two of my AutoHotkey scripts... so some breaking changes, not completely backwards compatible. Although from what I can tell not major. But just a heads up, a straight upgrade to v2 may cause some users scripts to break.

I mentioned some of the details in #2378 with workaround.

virginviolet:
The autohotkey package is updated to v2.0.18 on GitHub, but appears to not have been successfully pushed to the community repo.

https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/autohotkey
https://community.chocolatey.org/packages/autohotkey

asheroto:
The process also needs to end before running, otherwise you get this...

image

PR #2263, opened 2023-07-28
(spotify) Remove Windows Store version

pauby:
## Description
Detects and removes the Windows Store version of Spotify if it's already installed and the `--force` option is used. Warn otherwise.

## Motivation and Context
If the Windows Store version is installed, the package will not install.

## How Has this Been Tested?
Run in Windows 10 with the Microsoft Store version of Spotify installed.

Run without the `--force` option.

image

Run with the `--force` option:

image

## Types of changes

## Checklist:

### DIFF
--- a/automatic/spotify/tools/ChocolateyInstall.ps1
+++ b/automatic/spotify/tools/ChocolateyInstall.ps1
@@ -12,6 +12,24 @@ $arguments          = @{
     validExitCodes  = @(0, 1641, 3010)
 }
 
+# check we can use Get-AppxPackage
+if (Get-Command 'Get-AppxPackage' -ErrorAction SilentlyContinue) {
+    # there is likely going to be two packages returned for x86 and x64.
+    # we don't care about the architecture, just the version and they will both be the same.
+    $allAppxPackages = Get-AppxPackage
+    $installedAppXPackage = @($allAppxPackages | Where-Object -Property Name -eq 'SpotifyAB.SpotifyMusic')
+    if ($installedAppXPackage.Count -gt 0) {
+        if ($env:ChocolateyForce) {
+            #when you remove a package, you don't remove it per architecture. You just remove it for all architectures.
+            Write-Warning 'Attempting to remove Spotify installed from the Microsoft Store.'
+            Remove-AppxPackage -Package $installedAppXPackage[0].PackageFullName
+        }
+        else {
+            throw "Cannot install the Spotify package because the Microsft Store version is installed. Please uninstall this manually or add the '--force' option to the command line."
+        }
+    }
+}
+
 # Download the installer
 $arguments['file'] = Get-ChocolateyWebFile @arguments
 
@@ -24,10 +42,10 @@ schtasks.exe /Delete /TN $arguments['packageName'] /F
 # Wait for Spotify to start, then kill it
 $done = $false
 do {
-  if (Get-Process Spotify -ErrorAction SilentlyContinue) {
-    Stop-Process -name Spotify
-    $done = $true
-  }
+    if (Get-Process Spotify -ErrorAction SilentlyContinue) {
+        Stop-Process -name Spotify
+        $done = $true
+    }
 
-  Start-Sleep -s 10
+    Start-Sleep -s 10
 } until ($done)


pauby:
AppVeyor is failing for to a time out after running the scheduled task. I didn't changed this part of the code so I'm going to assume it's normal?

automatic/spotify/tools/ChocolateyInstall.ps1
```diff
@@ -12,6 +12,24 @@ $arguments          = @{
...
+if (Get-Command 'Get-AppxPackage' -ErrorAction SilentlyContinue) {
+    # there is likely going to be two packages returned for x86 and x64.
+    # we don't care about the architecture, just the version and they will both be the same.
+    $allAppxPackages = Get-AppxPackage
+    $installedAppXPackage = @($allAppxPackages | Where-Object -Property Name -eq 'SpotifyAB.SpotifyMusic')
+    if ($installedAppXPackage.Count -gt 0) {
+        if ($env:ChocolateyForce) {
+            #when you remove a package, you don't remove it per architecture. You just remove it for all architectures.
+            Write-Warning 'Attempting to remove Spotify installed from the Microsoft Store.'
+            Remove-AppxPackage -Package $installedAppXPackage[0].PackageFullName
```
AdmiringWorm, line 25: IMO, we should not remove the Microsoft Store application in this package.

It is still two different applications, which goes against having one software for each package.

IMO, I think we should throw an exception instead, no matter what, possibly with details on how the user can uninstall the application manually or add a new package that only has the purpose of removing the Store version of Spotify (that we possibly can depend on in this package).

automatic/spotify/tools/ChocolateyInstall.ps1
AdmiringWorm, line 1: Please fix the indentation used on the lines changed to follow the editor config file.
The editor config file specifies that indentations should only be two spaces.

AdmiringWorm:
> AppVeyor is failing for to a time out after running the scheduled task. I didn't changed this part of the code so I'm going to assume it's normal?

I'm honestly not sure. It could be that appveyor doesn't allow scheduled tasks to be created...

automatic/spotify/tools/ChocolateyInstall.ps1
pauby: A package parameter? Warn them but if they add the package parameter it also uninstalls. Seems like the best of both worlds.

automatic/spotify/tools/ChocolateyInstall.ps1
pauby: The indentation was updated to match what the rest of the file has. As it was inconsistent I used what it started with.

I'll update the file to match 2 spaces.

automatic/spotify/tools/ChocolateyInstall.ps1
AdmiringWorm: hmm. Add it behind a package parameter 🤔.

Yeah, could work as well. Let's go with that instead. I still do think a separate package could be beneficial though, but we can go with the parameter for now.

corbob:
> AppVeyor is failing for to a time out after running the scheduled task. I didn't changed this part of the code so I'm going to assume it's normal?

For what it's worth, I attempted to install the version of this package in CCR, and it seemed to do that same thing for me, as soon as I manually uninstalled the Spotify store app, then it didn't timeout and installed without issue. Depending on what AppVeyor is using, this could be the same thing (I would have thought it'd be on a Server OS without the Store, but maybe not 🤷 )

AdmiringWorm:
> > AppVeyor is failing for to a time out after running the scheduled task. I didn't changed this part of the code so I'm going to assume it's normal?
> 
> For what it's worth, I attempted to install the version of this package in CCR, and it seemed to do that same thing for me, as soon as I manually uninstalled the Spotify store app, then it didn't timeout and installed without issue. Depending on what AppVeyor is using, this could be the same thing (I would have thought it'd be on a Server OS without the Store, but maybe not 🤷 )

It is reporting that it is running on `Microsoft Windows Server 2016 Datacenter` (https://ci.appveyor.com/project/chocolateycommunity/chocolatey-packages/builds/48351047#L134). So it shouldn't be a problem with the Windows Store. It may be possible that Spotify does not run on a Server, or it only supports the absolute latest (I honestly do not know).

PR #2264, opened 2023-07-28
(vscode) Updated version check API call

andtong:
## Description
Updated version check API call. API was updated to require 40 characters for version input.

- vscode.install
- vscode-insiders.install

## Motivation and Context
Version check URL is no longer working stopping the automatic package upgrades.

## How Has this Been Tested?
Ran update.ps1 and choco pack locally. Tested package in a sandboxed environment.

## Types of changes
* [x]  Bug fix (non-breaking change which fixes an issue)
* [ ]  New feature (non-breaking change which adds functionality)
* [ ]  Breaking change (fix or feature that would cause existing functionality to change)
* [ ]  Migrated package (a package has been migrated from another repository)

## Checklist:
* [x]  My code follows the code style of this repository.
* [x]  My change requires a change to documentation (this usually means the notes in the description of a package).
* [ ]  I have updated the documentation accordingly (this usually means the notes in the description of a package).
* [ ]  I have updated the package description and it is less than 4000 characters.
* [x]  All files are up to date with the latest [Contributing Guidelines](https://github.com/chocolatey-community/chocolatey-packages/blob/master/CONTRIBUTING.md)
* [x]  The added/modified package passed install/uninstall in the chocolatey test environment.
* [ ]  The changes only affect a single package (not including meta package).

### DIFF
--- a/automatic/vscode-insiders.install/update.ps1
+++ b/automatic/vscode-insiders.install/update.ps1
@@ -1,8 +1,8 @@
 ﻿import-module au
 import-module "$PSScriptRoot\..\..\extensions\chocolatey-core.extension\extensions\chocolatey-core.psm1"
 
-$releases32 = 'https://update.code.visualstudio.com/api/update/win32/insider/VERSION'
-$releases64 = 'https://update.code.visualstudio.com/api/update/win32-x64/insider/VERSION'
+$releases32 = 'https://update.code.visualstudio.com/api/update/win32/insider/0000000000000000000000000000000000000000'
+$releases64 = 'https://update.code.visualstudio.com/api/update/win32-x64/insider/0000000000000000000000000000000000000000'
 
 if ($MyInvocation.InvocationName -ne '.') {
   function global:au_BeforeUpdate {
--- a/automatic/vscode.install/update.ps1
+++ b/automatic/vscode.install/update.ps1
@@ -1,8 +1,8 @@
 import-module au
 import-module "$PSScriptRoot\..\..\extensions\chocolatey-core.extension\extensions\chocolatey-core.psm1"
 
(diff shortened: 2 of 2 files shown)


pauby:
@andtong please go back and complete the PR template fully. We will be unable to consider this otherwise.

There.

andtong:
Let me know if there's anything else needed

pauby:
> API was updated to require 40 characters for version input.

@andtong Can you provide a source that confirms this?

Roemer:
The endpoint is to return the version of a specific githash (which is 40 characters). So I guess they now just added validation for that. As vscode is crucial for many, the package should in my opinion be fixed asap, even if the fix is not 100% verified but works.

pauby:
@Roemer The reason I asked is that if this is an official API that is documented there is less chance of these sorts of changes being made. 

If this is an undocumented API, they can change it at will, and we'll be here again in 2, 3, 5 or so weeks. I'd much rather we used the [GitHub releases](https://github.com/chocolatey-community/chocolatey-packages/issues/2269#issuecomment-1668220507) for this if it is viable (caveat I don't know if they are published on the same cadence).

Roemer:
In my opinion, it would still be better to use the undocumented api as this is alway up to date and works for vscode and vscode-insiders and fix it, if it breaks. The github release might be out of sync with the real version (the api provides) and also it does not work for vscode-insiders. Anyway, I opened a PR for a change to github releases here: https://github.com/chocolatey-community/chocolatey-packages/pull/2276
It's almost two weeks since we cannot update anymore and I think we should get anything working soon.

chimmmpie:
@pauby Any chance on merging this as is. Than do the github releases afterwards?

pauby:
#2276 has been submitted to use GitHub release for `vscode` so that should be removed from this PR.

PR #2296, opened 2023-09-07
(freecad) Increases Maintainability of Update Helper

JPRuskin:
## Description
Increases legibility / maintainability of the update helper function for FreeCAD, which was hard to grok.

This should also "fix" a portable asset not being available failing the build.

## Motivation and Context
This is a bit of a rewrite of the helper function that I'd done previously for the GitHub Release changes, plus a change to ensure that the build doesn't fail when the portable artifact isn't published as the portable zip hasn't been pushed to the release in a few versions.

I am unsure if the portable artifact is being published going forward (can't find any context / intent around the change), and we should address that separately if >22.x doesn't appear.

## How Has this Been Tested?
- `update_all.ps1 -Name FreeCAD` run with a non-existant portable release, should still "work" for other streams
- Import the helpers with `. .\update_helpers.ps1` and then test the output of the new and old versions of `Get-FreeCAD`. They should be identical (roughly speaking).
- Tested installing the generated packages in a test environment
    - Note: Uninstall doesn't appear to be fully silent, but not due to any changes I have made. I may need to address this as part of this PR, but as this PR was to fix the build I haven't yet.

## Types of changes

Note: I do not count this as a breaking change, despite the Get-FreeCAD signature completely changing, as it's only used here (internally, without any exposure to actual users) and is essentially a private function.

## Checklist:
- ~[ ] The changes only affect a single package (not including meta package).~ Affects three streams of package / two actual package IDs.

### DIFF
--- a/automatic/freecad/freecad.nuspec
+++ b/automatic/freecad/freecad.nuspec
@@ -51,6 +51,7 @@ Example: `choco install freecad --params "/InstallDir:'C:\FreeCAD' /NoShortcut"`
     <tags>freecad 3d cad modeling engineering admin</tags>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/freecad</packageSourceUrl>
     <dependencies>
+      <dependency id="chocolatey" version="1.0.0" />
       <dependency id="chocolatey-core.extension" version="1.3.3" />
     </dependencies>
   </metadata>
--- a/automatic/freecad/update.ps1
+++ b/automatic/freecad/update.ps1
@@ -22,6 +22,7 @@ function global:au_SearchReplace {
       "(?i)(^\s*\<id\>).*(\<\/id\>)"                     = "`${1}$($Latest.PackageName)`${2}"
       "(?i)(^\s*\<title\>).*(\<\/title\>)"               = "`${1}$($Latest.Title)`${2}"
       "(?i)(^\s*\<releaseNotes\>).*(\<\/releaseNotes\>)" = "`${1}$($Latest.ReleaseNotes)`${2}"
+      '(?i)(^\s*\<dependency id="chocolatey" version=").*(" \/\>)' = "`${1}$(if ($Latest.FileType -eq '7z') {'2.2.2'} else {'1.0.0'})`${2}"
     }
     ".\tools\chocolateyUninstall.ps1" = @{
       "(?i)(^\s*packageName\s*=\s*)'.*'"  = "`$1'$($Latest.PackageName)'"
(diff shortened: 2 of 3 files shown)


JPRuskin:
image

I am unsure quite what's happening with this, as it works on my box. @AdmiringWorm have you seen anything like this before?

If it were another environment, I would assume the zip was being modified by some form of security feature, but AppVeyor hasn't had that issue (to my knowledge) before.

automatic/freecad/freecad.nuspec
```diff
@@ -51,6 +51,7 @@ Example: `choco install freecad --params "/InstallDir:'C:\FreeCAD' /NoShortcut"`
     <tags>freecad 3d cad modeling engineering admin</tags>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/freecad</packageSourceUrl>
     <dependencies>
+      <dependency id="chocolatey" version="1.0.0" />
```
AdmiringWorm, line 54: This is the wrong version. It is the `2.x` version that has the correct algorithm.

AppVeyor upgraded to `2.2.2` automatically during the freecad installation (may be a bug in CLI).

Please update it to be (I think) `2.2.2`.

AdmiringWorm (changes requested):

automatic/freecad/freecad.nuspec
JPRuskin: So, this is using 1.0.0 for the Freecad-non7z packages, and 2.2.2 for the 7z package that depends on that. I'm unsure if it would be better to remove the dependency entirely for the install packages, but 1.0 seemed a reasonable compromise there - I assumed we didn't want to have a latest-version dependency where it's not actually needed?

automatic/freecad/update.ps1
```diff
@@ -25,6 +22,7 @@ function global:au_SearchReplace {
       "(?i)(^\s*\<id\>).*(\<\/id\>)"                     = "`${1}$($Latest.PackageName)`${2}"
       "(?i)(^\s*\<title\>).*(\<\/title\>)"               = "`${1}$($Latest.Title)`${2}"
       "(?i)(^\s*\<releaseNotes\>).*(\<\/releaseNotes\>)" = "`${1}$($Latest.ReleaseNotes)`${2}"
+      '(?i)(^\s*\<dependency id="chocolatey" version=").*(" \/\>)' = "`${1}$(if ($Latest.FileType -eq '7z') {'2.2.2'} else {'1.0.0'})`${2}"
```
JPRuskin, line 25: @AdmiringWorm This is (probably) the reason AppVeyor upgraded during the run, as the 7z version would require 2.2.2.

automatic/freecad/freecad.nuspec
AdmiringWorm: We do not want to have a dependency on Chocolatey CLI if we do not need to have one.

So in the end for this package, only the version that needs the dependency should add it, while the other versions should have no dependency on the CLI.

automatic/freecad/update.ps1
AdmiringWorm: Ah, I completely missed that

PR #2302, opened 2023-09-07
(youtube-dl) Fixes Update.ps1 and adds Nightly stream

JPRuskin:
## Description
This commit both switches to using the GitHub releases for sourcing update information, but also adds nightly builds to the streams to build.

## Motivation and Context
The build was broken due to the update-target being a site that has been taken down. This moves the target to the GitHub release page, and adds a nightly versions to the repository when found (with a `-nightly` addition to the version to ensure they track as pre-release).

## How Has this Been Tested?
- Ran `upgrade_all.ps1 -Name youtube-dl -force youtube-dl` before and after changes were made

## Types of changes

## Checklist:

### DIFF
--- a/automatic/youtube-dl/update.ps1
+++ b/automatic/youtube-dl/update.ps1
@@ -1,8 +1,5 @@
-import-module au
-import-module "$env:ChocolateyInstall\helpers\chocolateyInstaller.psm1"
-
-$domain   = 'https://youtube-dl.org'
-$releases = "$domain/"
+Import-Module AU
+Import-Module "$env:ChocolateyInstall\helpers\chocolateyInstaller.psm1"
 
 function global:au_BeforeUpdate {
   if (Test-Path "$PSScriptRoot\tools") {
@@ -18,7 +15,7 @@ function global:au_BeforeUpdate {
 function global:au_SearchReplace {
   @{
     ".\legal\VERIFICATION.txt" = @{
-      "(?i)(listed on\s*)\<.*\>" = "`${1}<$releases>"
+      "(?i)(listed on\s*)\<.*\>" = "`${1}<$($Latest.ReleaseUrl)>"
       "(?i)(1\..+)\<.*\>"          = "`${1}<$($Latest.URL32)>"
       "(?i)(checksum type:).*"   = "`${1} $($Latest.ChecksumType32)"
       "(?i)(checksum:).*"       = "`${1} $($Latest.Checksum32)"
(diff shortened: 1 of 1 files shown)


automatic/youtube-dl/update.ps1
```diff
@@ -11,14 +8,14 @@ function global:au_BeforeUpdate {
     New-Item -ItemType Directory "$PSScriptRoot\tools"
   }
   $Latest.FileName = Get-WebFileName $Latest.URL32 'youtube-dl.exe'
-  $filePath = "$PSScriptRoot\tools\$($Latest.FileName)"
-  Get-WebFile $Latest.URL32 $filePath
+  $FilePath = "$PSScriptRoot\tools\$($Latest.FileName)"
+  Get-WebFile $Latest.URL32 $FilePath
```
pauby, line 12: Is there a reason to PascalCase the variable names (I don't see them as function parameters)? They should be camelCase.

automatic/youtube-dl/update.ps1
```diff
@@ -27,22 +24,37 @@ function global:au_SearchReplace {
...
+    if ($Repository.EndsWith('nightly')) {
+      $Version += '-nightly'
+    }
+
+    $Checksum = if ($ChecksumFile = $LatestRelease.assets.Where{$_.name -eq 'SHA2-512SUMS'}) {
+      $Checksum = (-join [char[]](Invoke-WebRequest -Uri $ChecksumFile.browser_download_url -UseBasicParsing).content).Split("`n") -match "^(?<checksum>.+)(?=\s+youtube-dl\.exe)"
+      $Checksum.Split(' ')[0]
+    } else {
+      Get-RemoteChecksum -Algorithm "sha512" $Url
+    }
```
pauby, line 42: Could this be simplified. Have an assignment with an if statement assigning to the results of a where is overly complicated. 

I _don't think_ we do cuddled elses.

pauby (changes requested):

automatic/youtube-dl/update.ps1
JPRuskin:
```suggestion
    $Checksum = if ($ChecksumFile = $LatestRelease.assets.Where{$_.name -eq 'SHA2-512SUMS'}) {
      ((-join [char[]](Invoke-WebRequest -Uri $ChecksumFile.browser_download_url -UseBasicParsing).content).Split("`n") -match "^(?<checksum>.+)(?=\s+youtube-dl\.exe)").Split(' ')[0]
    } else {
      Get-RemoteChecksum -Algorithm "sha512" $Url
    }
```

automatic/youtube-dl/update.ps1
JPRuskin: Do we have guidance that suggests that? Our code, I think, doesn't reflect it.

In the original version of this file, we seem to have at least three different types of variable casing 😅:
- `$Latest`
- `$domain`
- `$download_page`
- `$filename` (half point, arguable)

I know I spoke about a coding standards document for PowerShell code way back when - I don't think it ever happened. I've always been a fan of / used by default the [PoshCode PowerShell Practice and Style guide](https://poshcode.gitbook.io/powershell-practice-and-style/style-guide/code-layout-and-formatting) which says:

> PowerShell uses PascalCase for all public identifiers: module names, function or cmdlet names, class, enum, and attribute names, public fields or properties, global variables and constants, etc. [...]
> If you wish, you may use camelCase for variables within your functions (or modules) to distinguish private variables from parameters, but this is a matter of taste.

C# certainly uses camel case for private variables more frequently. To me, the above indicates that it's the second-choice in PowerShell.

pauby: There is lots in that guide that criticises itself and then trips up over itself then criticises itself again. It doesn't ever make statements, just guidelines, recommendations and personal opinions. I really wish it was more definitive. Using it however, it does suggest:

> If you wish, you may use camelCase for variables within your functions (or modules) to distinguish private variables from parameters, but this is a matter of taste. Shared variables should be distinguished by using their scope name, such as $Script:PSBoundParameters or $Global:DebugPreference. If you are using camelCase for a variable that starts with a two-letter acronym (where both letters are capitalized), both letters should be set to lowercase (such as adComputer).

`$Latest` is a global variable, so using the style guide again:

> PowerShell uses PascalCase for all public identifiers: module names, function or cmdlet names, class, enum, and attribute names, public fields or properties, global variables and constants, etc.

`$domain` and `$filename` are camelCase (there just is no second word to Case). 

If the change was made for consistency then I'd agree. If you changed all of the variables to PascalCase then I could see your reasoning. but I don't see the need for this change at all as all it's doing is renaming a variable from camelCase to PascalCase and nothing more.

automatic/youtube-dl/update.ps1
pauby: We don't use cuddled else's as far as I am aware but I'll defer to @AdmiringWorm on that.

With regards to the [complexity of the code](https://github.com/chocolatey-community/chocolatey-packages/blob/master/CONTRIBUTING.md#117-clean-code):

> Remember, code is written for humans, not for computers. Otherwise, we'd all write assembly. So it's better to be able to reason about code than it is to take shortcuts that make code harder to decipher.

I know that complexity is subjective, but this needs to be simplified in my opinion. Breaking it out into separate steps would seem to be a good start rather than doing everything in one line. With comments.

AdmiringWorm: Throughout different changes, there have been statements that were both cuddled and not cuddled, as well as on its own line and not.
It looks to be more or less an equal amount of these (this mix is due to, historically, the look of the code did not matter, just that it works).

Personally, I would prefer else statements starting on their line (but I am not tied to that).

Also, I agree with the complexity. It makes reading what you are doing hard and needs to be simplified.

automatic/youtube-dl/update.ps1
JPRuskin: > [...] $filename are camelCase (there just is no second word to Case)

Impossible to tell really, but yes, I gave it a half point. :) File name and user name always get me as fun compound words that I'm never sure if I prefer split and (arguably) work either way as a definition. Yay, naming/words.

If we were more consistent, had written guidelines, or relevant formatter rules in place this would be easier! :) We've both quoted "it's a matter of taste" to each other, now, and the chunk about public identifiers - though I had them the other way around.

Of course, the downside of the many-scoped, joyfully fun layout of AU would likely mean that any automated rule would completely fail to parse the difference between global variables (that haven't been prefixed with `$global:`, which is, in my opinion, the better way to show scope) and local variables, but... there's only so much fun we can have, I suppose.

No matter. Sorted. Or resorted, or unsorted, or something. :)

automatic/youtube-dl/update.ps1
JPRuskin: Broken it out into many lines with comments in my latest push.

I'm happy to argue my case for us adopting cuddled-elses and no-new-line as our overall preference for a few reasons:
- It reduces the potential for contributors accidentally breaking chained statements by inserting new lines between an `if {}` and an `else {}` - to me it offers a very clear linkage.
- I love being able to paste chunks of code into terminals without modification when I'm debugging scripts, which doesn't happen reliably if the `if {}` has evaluated already (and then it makes a complete hash of the else-scriptblock)

Overly Simplistic Example:
image

If I wasn't on holiday for it, I'd say we could talk about it at the upcoming maintainers meeting. If we do so, perhaps we can look to add a rule to the formatting PSD1 and make a formatting commit to standardise once and for all.

pauby (approved):

pauby:
I've tested this and it works fine.

Thanks for your work @JPRuskin!

Issue #2187, opened 2023-03-25
(multiple) Fix packages failing normal updates and automated checkes


AdmiringWorm:
## Expected Behavior

All packages are expected to be able to automatically update, and once they are pushed to Chocolatey Community Repository they are expected to successfully pass all of the automated checks.

## Current Behavior

At the time of writing we have multiple packages that are failing our automatic updater, as well as there are multiple packages that are failing Chocolatey Community Repository automatic checks (validator, verifier and package scanner).

## Possible Solution

That is not known at this time, we need to look at trying to fix all failing packages. If they can not be fixed we may need to look at removing these packages.

## Steps to Reproduce (for bugs)

N/A

## Context

This issue is meant as tracking failing packages, which have their own issues (in the case of failing automated checks).

See the following list of failing updater packages, and links to other issues that is failing automated checks.

```[tasklist]
### Packages failing automated updates
- [ ] autohotkey
- [ ] autohotkey.install
- [ ] autohotkey.portable
- [ ] #2158
- [ ] encfs4win
- [ ] librecad
- [ ] maxthon.commandline
- [ ] nomacs
- [ ] nomacs.portable
- [ ] pencil
- [ ] prey
- [ ] pspp
- [ ] umlet
- [ ] wixtoolset
- [ ] wps-office-free
### Packages failing automated checks
- [ ] #2182 
- [ ] #2171 
- [ ] #2121 
```

## Your Environment

N/A

nykBM:
Please add [Ghostscript (Install) 10.01.0](https://community.chocolatey.org/packages/Ghostscript.app/10.01.0) and [Ghostscript 10.01.0](https://community.chocolatey.org/packages/Ghostscript/10.01.0) to the list.  I do not know if this is the same issue but they are failing automatic verification with the new versions.

AdmiringWorm:
@nykBM Thank you, I have added those to the failing automated checks section.

PR #2317, opened 2023-10-01
(pencil) Fix update script and installation

Windos:
## Description

This PR adjusts the pencil update script so that it works correctly, and also updates the chocolateyInstall.ps1 script and VERIFICATION.txt to reflect these changes.

This PR also expands a number of aliases in the pencil update script.

## Motivation and Context

At the moment the update for this package fails, due to the assumption that there is a separate 32-bit and 64-bit installers for Windows. This is no longer the case and there is only a single installer for Windows now.

This failure has been noted in #2187.

## How Has this Been Tested?

On a vanilla Windows 10 (fully patched) test VM I verified that running the pencil packages' update.ps1 script fails with the same error as seen when run via this [repository's automation](https://gist.github.com/choco-bot/a14b1e5bfaf70839b338eb1ab7f8226f#pencil).

I retested after making these changes and verified that the update.ps1 script succeeded.

The resulting nupkg successfully installed the pencil application on my test VM, and also in my local copy of Chocolatey test environment.

N/A

## Types of changes

## Checklist:


automatic/pencil/tools/chocolateyInstall.ps1
```diff
@@ -5,8 +5,7 @@ $toolsPath = Split-Path $MyInvocation.MyCommand.Definition
 $packageArgs = @{
   packageName    = $Env:ChocolateyPackageName
   fileType       = 'exe'
-  file           = Get-Item $toolsPath\*-i386.exe
-  file64         = Get-Item $toolsPath\*_64.exe
+  file           = Get-Item $toolsPath\Pencil-*.exe
```
pauby, line 8: We avoid doing this now and all for the file name to be explicitly used.

pauby (changes requested):
Thanks for picking this up.

automatic/pencil/tools/chocolateyInstall.ps1
Windos: I've fixed that up

pauby (approved):

pauby:
@Windos awesome work! Thanks for your contribution!

PR #2326, opened 2023-10-10
(musescore) Cleanup update.ps1 links

umol-unibas:
## Description
The changed hard-coded link is not necessary in the old format and the way it is worded suggests that the download location is sufficient. There is already a valid download-link generated and pushed to the VERIFICATION.txt, so there is no need for a second one.

## Motivation and Context
In the VERIFICATION.txt were two msi-links included. As the first one is not necessary and suggests that the download location is sufficient, it would be great if it could be changed, because of some automations we are using.

## How Has this Been Tested?
This change does not harm or interrupt the update-module and its process at all as the link is only there for informational purpose.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/musescore/update.ps1
+++ b/automatic/musescore/update.ps1
@@ -1,4 +1,4 @@
-﻿﻿Import-Module AU
+﻿Import-Module AU
 
 $releases = 'https://musescore.org/en/download/musescore.msi'
 
@@ -7,7 +7,6 @@ function global:au_BeforeUpdate { Get-RemoteFiles -Purge -NoSuffix }
 function global:au_SearchReplace {
   @{
     ".\legal\VERIFICATION.txt"      = @{
-      "(?i)(^\s*location on\:?\s*)\<.*\>" = "`${1}<$releases>"
       "(?i)(\s*1\..+)\<.*\>"              = "`${1}<$($Latest.URL32)>"
       "(?i)(^\s*checksum\s*type\:).*"     = "`${1} $($Latest.ChecksumType32)"
       "(?i)(^\s*checksum(32)?\:).*"       = "`${1} $($Latest.Checksum32)"
--- a/automatic/musescore/update.ps1
+++ b/automatic/musescore/update.ps1
@@ -1,4 +1,4 @@
-﻿Import-Module AU
+﻿﻿Import-Module AU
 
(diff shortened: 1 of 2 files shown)


automatic/musescore/update.ps1
```diff
@@ -1,6 +1,6 @@
 ﻿Import-Module AU
 
-$releases = 'https://musescore.org/en/download/musescore.msi'
+$releases = 'https://musescore.org/en/download'
```
AdmiringWorm, line 3: This change is invalid, without additional work done in the `au_GetLatest` handler.

You are effectively breaking the current automatic updater by changing this URL without making any other changes.

Can you please make the necessary changes to make the updater work as well?

AdmiringWorm (changes requested):

automatic/musescore/update.ps1
umol-unibas: Sorry for the confusion! I mixed things up here. For sure i did not intend to break the update-script. I corrected it now, on basis of the [disscussion with @pauby](https://github.com/orgs/chocolatey-community/discussions/2325). Would be great to have it merged. Sorry for not replying sooner - i was on vacation ;)

pauby:
This latest change is being made in the wrong place. If we are going to hard code the link we should just update the VERIFICATION.txt file directly and remove the updating of 'location on' that is done here. There isn't any point in automatically updating a file with the same text each time if it's not going to change. And if it does change then we need to update the update script anyway so it's no more hassle updating the VERIFICATION.txt file directly.

@umol-unibas Does that make sense?

umol-unibas:
You are right @pauby. There is no need in updating the update.ps1 as it works fine - i'm not sure why i changed that. According to your message and our [discussion ](https://github.com/orgs/chocolatey-community/discussions/2325) i reverted the update-script and changed the VERIFICATION.txt. That was my idea was in the first place for this PR. 

Reasons for the changes in the VERIFICATION.txt: the hard-coded link ("https://musescore.org/en/download/musescore.msi ") is 
- not necessary as it is a duplication
- it could be puzzling because of the other link
- it has no meaning

AdmiringWorm (approved):
LGTM

umol-unibas:
Thanks!

PR #2357, opened 2023-11-19
(libreoffice-still) LibreOffice-Still 7.5.8.2

zWhdmB5T:
(libreoffice-still) LibreOffice-Still 7.5.8.2


brogers5:
This was functionally duplicated by a88561de6f473777d8c5ae001fee64775b3c9668 and should no longer be necessary to review/merge. I'd recommend closing this PR.

AdmiringWorm:
To note, even if the automatic commit didn't happen we wouldn't be able to accept this PR as it would only cause updates to **NOT** happen.

This is an automatic package that is completely dependent on when the developers of LibreOffice decide to give the updates (it uses the same update URL as when checking manually in LibreOffice).

As such, I will be closing this issue now. We do appreciate that you took the time and effort to attempt to get it updated.

zWhdmB5T:
Sorry for any inconvenience (I myself learned a lot in this case…). Resolution in corresponding issue as stated. Thanks a lot!

PR #2366, opened 2023-12-07
(datagrip) Strip `.0` suffix from installation directory

9numbernine9:
## Description

When determining the installation directory for Datagrip, remove any `.0` suffix that might be present in the version number. Jetbrains doesn't use this part of the release version number when installing, so this causes Chocolatey to create an empty directory that never gets used or cleaned up.

## Motivation and Context
Fixes: https://github.com/chocolatey-community/chocolatey-packages/issues/2257

## How Has this Been Tested?
`choco install -s . datagrip` with these changes; observe that spurious `.0` directory is no longer created.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/datagrip/tools/ChocolateyInstall.ps1
+++ b/automatic/datagrip/tools/ChocolateyInstall.ps1
@@ -6,7 +6,9 @@ $toolsDir     = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
 $programFiles = (${env:ProgramFiles(x86)}, ${env:ProgramFiles} -ne $null)[0]
 $pp = Get-PackageParameters
 
-$installDir = "$programFiles\JetBrains\DataGrip $env:ChocolateyPackageVersion"
+# If the release version ends in '.0', remove this since JetBrains don't
+# use this part of the version number when creating the installation directory.
+$installDir = "$programFiles\JetBrains\DataGrip $($env:ChocolateyPackageVersion -replace '.0$', '')"
 if ($pp.InstallDir) {
     $installDir = $pp.InstallDir
 }


automatic/datagrip/tools/ChocolateyInstall.ps1
```diff
@@ -6,7 +6,9 @@ $toolsDir     = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
 $programFiles = (${env:ProgramFiles(x86)}, ${env:ProgramFiles} -ne $null)[0]
 $pp = Get-PackageParameters
 
-$installDir = "$programFiles\JetBrains\DataGrip $env:ChocolateyPackageVersion"
+# If the release version ends in '.0', remove this since JetBrains don't
+# use this part of the version number when creating the installation directory.
+$installDir = "$programFiles\JetBrains\DataGrip $($env:ChocolateyPackageVersion -replace '.0$', '')"
```
pauby, line 11: We're creating the directory further down on line 19. Can we not just remove that line and let the installer install wherever it thinks it should? Why do we need to create the directory?

9numbernine9: Well... originally I had assumed (possibly incorrectly) that the `New-Item -ItemType Directory` step in the original install script was required, but based on my own testing maybe it's not needed anymore? My quick searching seems to imply that it was part of a workaround for this older bug in the Jetbrains installer ( https://youtrack.jetbrains.com/issue/IDEA-202935 ) but it seems that since then the installer bug has been fixed, so this directory creation step might just be vestigial!

I can change the PR to remove the unnecessary cruft if you'd prefer?

AdmiringWorm: The directory creation was indeed just because of a bug in the installer that would cause any custom directories used to fail the installation.

If this is now fixed, and works as expected, then I agree with @pauby that removing the manual creation of the directory instead of replacing it here.

To note though as well. IMO this wouldn't be the correct approach anyway, as if the directory does not respect the 3rd part of a version number if it is zero, we should not use the `ChocolateyPackageVersion` environment variable, but instead a hard-coded value we automatically insert when creating the package.

PR #2374, opened 2023-12-17
VERIFICATION.txt

Arietriss:
Updated link to latest version, and checksum of official installer of latest version (7.2.0)


pauby:
Thanks for the PR.

We don't update that file directly. It is instead updated by the update script.

As we can't accept this PR, I'll go ahead and close it.

PR #2393, opened 2024-01-05
(ruby*) Correct package verification failures

### LINKED ISSUE #2260: (ruby)(ruby.install) Packages rejected after failing verification
## Current Behavior
Both the `ruby` and `ruby.install` packages are currently failing package verification with the last 4 versions having been rejected.

image

This needs to be investigated and the issues resolved.

chadlwilson:
## Description
As discussed with @AdmiringWorm in https://github.com/chocolatey-community/chocolatey-packages/issues/2260#issuecomment-1878142315 - attempts to workaround the failing `ProjectSourceUrl` checks in automation for all Ruby packages on the assumption that the old URLs are now being blocked on `git.ruby-lang.org` for some reason (`The ProjectSourceUrl element in the nuspec file should be a valid Url`).

Also corrects silent install for RubyInstaller packages.

## Motivation and Context

- Fixes #2260 

All new Ruby versions are stuck in validation right now due to `ProjectSourceUrl` (also affects portable install).

_Edited:_
There are also historical issues going back to `3.2.0` with RubyInstaller due to the installer now adding a `/allusers` vs `/currentuser` option which must be chosen even for silent installs, which this addresses by defaulting to `/allusers`, as is the norm for Chocolatey installs.

## Types of changes

## Checklist:

### DIFF
--- a/automatic/ruby.install/ruby.install.nuspec
+++ b/automatic/ruby.install/ruby.install.nuspec
@@ -26,7 +26,7 @@ Example: `choco install ruby --package-parameters="'/NoPath  ""/InstallDir:C:\yo
     <tags>ruby language programming development dynamic cross-platform foss</tags>
     <licenseUrl>https://www.ruby-lang.org/en/about/license.txt</licenseUrl>
     <projectUrl>https://www.ruby-lang.org</projectUrl>
-    <projectSourceUrl>https://git.ruby-lang.org/ruby.git/tree</projectSourceUrl>
+    <projectSourceUrl>https://github.com/ruby/ruby</projectSourceUrl>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/ruby.install</packageSourceUrl>
     <bugTrackerUrl>https://bugs.ruby-lang.org/projects/ruby-master/issues</bugTrackerUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
--- a/automatic/ruby.install/tools/chocolateyInstall.ps1
+++ b/automatic/ruby.install/tools/chocolateyInstall.ps1
@@ -16,7 +16,7 @@ $packageArgs = @{
   fileType       = 'exe'
   file           = "$toolsPath\rubyinstaller-3.3.0-1-x86.exe"
   file64         = "$toolsPath\rubyinstaller-3.3.0-1-x64.exe"
-  silentArgs     = '/verysilent /dir="{0}" /tasks="{1}"' -f $installDir, ($tasks -join ',')
+  silentArgs     = '/verysilent /allusers /dir="{0}" /tasks="{1}"' -f $installDir, ($tasks -join ',')
   validExitCodes = @(0)
   softwareName   = 'ruby *'
 }
--- a/automatic/ruby.portable/ruby.portable.nuspec
+++ b/automatic/ruby.portable/ruby.portable.nuspec
@@ -18,7 +18,7 @@ Ruby is a dynamic, open source programming language focusing on simplicity and p
     <tags>ruby language programming development dynamic cross-platform foss portable</tags>
     <licenseUrl>https://www.ruby-lang.org/en/about/license.txt</licenseUrl>
     <projectUrl>https://www.ruby-lang.org</projectUrl>
-    <projectSourceUrl>https://git.ruby-lang.org/ruby.git/tree</projectSourceUrl>
+    <projectSourceUrl>https://github.com/ruby/ruby</projectSourceUrl>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/ruby.portable</packageSourceUrl>
     <bugTrackerUrl>https://bugs.ruby-lang.org/projects/ruby-master/issues</bugTrackerUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
--- a/automatic/ruby/ruby.nuspec
+++ b/automatic/ruby/ruby.nuspec
@@ -27,7 +27,7 @@ Example: `choco install ruby --package-parameters="'/NoPath  ""/InstallDir:C:\yo
     <tags>ruby language programming development dynamic cross-platform foss</tags>
     <licenseUrl>https://www.ruby-lang.org/en/about/license.txt</licenseUrl>
     <projectUrl>https://www.ruby-lang.org</projectUrl>
-    <projectSourceUrl>https://git.ruby-lang.org/ruby.git/tree</projectSourceUrl>
+    <projectSourceUrl>https://github.com/ruby/ruby</projectSourceUrl>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/ruby</packageSourceUrl>
     <bugTrackerUrl>https://bugs.ruby-lang.org/projects/ruby-master/issues</bugTrackerUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>


chadlwilson:
OK, the `ProjectSourceUrl` errors are gone, but now the `ruby.install` package is failing for the same reasons that the earlier versions were failing through the year (from memory)

```
ruby.install v3.3.0.1
ruby.install package files install completed. Performing other installation steps.
Ruby is going to be installed in 'C:\tools\ruby33'
Installing 64-bit ruby.install...
Chocolatey timed out waiting for the command to finish. The timeout
 specified (or the default value) was '600' seconds. Perhaps try a
 higher `--execution-timeout`? See `choco -h` for details
```

Can probably split `ruby.portable` out to unblock that, as that one is OK, which I can do tomorrow.

In terms of figuring out why the rubyinstaller executables are now getting stuck (believe it started with `3.2.0` and continued with `3.1.4`, `2.7.8` etc, so likely a change in rubyinstaller or its dependencies (msys2?) between `2022-11-24` and `2022-12-25`.

My guess would be due to options added for per-user or all-users installation from https://github.com/oneclick/rubyinstaller2/blob/d808ebc8c15ca0a1724456d5761a53889f5688cb/CHANGELOG-3.2.md?plain=1#L26-L30

https://github.com/oneclick/rubyinstaller2/wiki/FAQ#user-content-silent-install

I've tried to add `/allusers` to validate unblocking this, but probably need to think about the appropriate default (most other packages seem to use /allusers?) and whether the package needs an option added.

AdmiringWorm:
> I've tried to add /allusers to validate unblocking this, but probably need to think about the appropriate default (most other packages seem to use /allusers but I think things like a subsequent ridk install 2 3 might not work without the write permissions you only get with a /currentuser install?) and whether the package needs an option added.

Hmm, I know too little about Ruby to say for sure. From what you are saying, it does indeed sound like using `/currentuser` could be the better option. The problem is though that generally (and preferably) softwares installed through Chocolatey CLI should be systemwide, not per user (so usually `/allusers` is appropriate).

I'll think about this, and come back to this PR on Monday, but would love to hear the thoughts from other people as well (especially those working with Ruby daily).

chadlwilson:
Ok, cheers. I'm not a ruby-on-windows expert either :)

I'll try such a machine wide install empirically and play around with ridk to see how it behaves. If one has to do ridk installs or msys2 updates via admin (post choco install) I dont think that's unreasonable, and the changelog implies they have done some magic to allow gem installs by the current user even with machine wide installs, so maybe it's OK.

I have no idea how the upgrade path will behave either though. Sigh.

Probably should also check what the behaviour of the older releases was as well.

chadlwilson:
OK, after testing, I think this will work just fine with `/allusers` install change above as the most sensible default.

- Prior versions of the package fail with a permissions error anyway when trying to `gem install xx` as a normal user after being installed as admin (with the normal chocolatey administrator recommendation) anyway.
- The new versions will allow users to install gems as current user with an /allusers install, due to the magic RubyInstaller folks have implemented to install gems into a local path
  ```
  PS C:\Users\chad> gem install hello-world
  Fetching hello-world-1.2.0.gem
  Using rubygems directory: C:/Users/chad/.local/share/gem/ruby/3.3.0
  Successfully installed hello-world-1.2.0
  ```
- Using ridk to install msys etc works fine (via interactive installer), even if done as a normal user after using Chocolatey to install `ruby.install` for "all users".
- You can then install gems that require native compiles as well with the MSYS2 DevKit which I've tested.
- I'm not 100% sure what will happen when _upgrading_ an existing install, but this package puts new point releases (`3.1.x` -> `3.2.x`) of Ruby into new folders anyway so that's a question limited to patches release upgrades (e.g `3.1.3` -> `3.1.4`). The RubyInstaller folks don't mention anything specific in the release notes that I've seen, so I'll make the naive assumption that it just works.
- I have NOT personally tested the variants where you ignore Chocolatey warnings and try and install/upgrade as a regular user. I'd assume if folks had previously installed via Choco for current user, they will need to do via admin going forward.

**Conclusion**
Despite [the `/currentuser` recommendation from the RubyInstaller folks](https://github.com/oneclick/rubyinstaller2/wiki/FAQ#user-content-install-mode), it sounds like `/allusers` is still the better default approach for Chocolatey users and I believe will work fine - and even better than the previous release due to the [support they have added for current user gem usage on machine wide installs](https://github.com/oneclick/rubyinstaller2/blob/d808ebc8c15ca0a1724456d5761a53889f5688cb/CHANGELOG-3.2.md?plain=1#L31) in these newer installers. 
  > * Enable automatic private gem installation in user's home on a machine wide ruby setup, if the user doesn't have write access.

Given there have been no new Ruby packages published to Chocolatey for the last 1 year and not that many issues/comments raised, perhaps we can conclude that there aren't too many highly motivated folks worried about doing so via Chocolatey, so having **some** packages working is better than nothing? :-)

Let me know if you want me to split this PR up, and your opinion of the path forward.

AdmiringWorm (approved):
LGTM

AdmiringWorm:
@chadlwilson your changes have been merged. Thanks for your contribution 👍

chadlwilson:
@AdmiringWorm thanks! assuming the latest 3.3 versions go through, is it possible somehow to trigger/produce some of the missing versions for ruby and ruby.install over the past year, using the current package source?

Im thinking ideally we’d have the latest versions in each major Ruby release, so 3.2.2 / 3.1.4 / 3.0.6 (and maybe 2.7.8 as the final security release when it became EOL).

If there is hassle involved and I were to pick just one, it’d be 3.2.2 - since due to the issues here there has been no Chocolatey non-portable package in the 3.2.x line and 3.3 is very new, missing a lot of native package support right now (until the community produce new library versions with windows native code confirmed against 3.3).

chadlwilson:
If you could give me a pointer to what I need to do to trigger a rebuild of https://community.chocolatey.org/packages/ruby.install/3.3.0.1 .

Does it need a bump to the nuspecs to 3.3.0.2 or can it be re-run on the latest package source somehow?

AdmiringWorm:
> Im thinking ideally we’d have the latest versions in each major Ruby release, so 3.2.2 / 3.1.4 / 3.0.6 (and maybe 2.7.8 as the final security release when it became EOL).

It is on my TODO list to initiate a new push for these, but that may be some time before that happens (need to unreject any existing packages, then trigger a new build for every minor version separately. Need available time before I do that).

> If you could give me a pointer to what I need to do to trigger a rebuild of [community.chocolatey.org/packages/ruby.install/3.3.0.1](https://community.chocolatey.org/packages/ruby.install/3.3.0.1) that’d also be good.

I know you struck this out, but I thought I'd at least provide some information here anyway.

Unfortunately, currently, only people with write permissions (or can merge PR's) can trigger a new rebuild of a package in this repository.

For us to trigger a forced rebuild, with need to add a new commit with a specific commit message to make that happen (this is detailed in the [Provide a Fix Version](https://github.com/chocolatey-community/chocolatey-packages/blob/master/COMMITTERS.md#providing-a-fixed-version) in our Committers guideline). This additionally must be the last commit submitted to the repository.

chadlwilson:
I've tried to make sense of what I could do to help this along, however given the ruby ones seem to use streams and 4-part version numbers already it seems perhaps a bit more complicated.

On the plus side, Ruby `3.2.3` was released a few days ago so when RubyInstaller updates I think the automation will 'just work' for that one as a new release - which will at least close the major gap in the releases (since there is no 3.2.x release on Chocolatey). For the earlier  3.1/3.0/2.7 streams Chocolatey will be missing the most recent security patch releases, but not as big of a gap.

chadlwilson:
The `3.2.3` releases have come through the automation across `ruby`/`ruby.install`/`ruby.portable` subsequent to the fix here, so think we are in pretty good shape.

It's possible the team are prepping a `3.1.5` release based on the branch still getting backports and changes, so with any luck we wont have to worry about that feature stream either.

chadlwilson:
We now have new builds off the `3.1` and `3.0` streams, so this is all looking good.

`2.7.x` is EOL so not expecting that to be changed to provide `2.7.8` or `2.7.9` Chocolatey packages.

PR #2395, opened 2024-01-09
(cmake) Fix AU version detection

### LINKED ISSUE #2394: (cmake) outdated
### Checklist

- [X] I have checked the moderation queue that no new version has been submitted (only visible when logged in), or the submitted package version has failed the automated checks.

Please navigate to the following [link](https://community.chocolatey.org/packages?q=&moderatorQueue=true&moderationStatus=all-statuses&prerelease=false&sortOrder=package-download-count) to view the moderation queue.

- [X] I have verified that the new version is a stable release.
- [X] I have looked to see if there are any Bug reports that prevents a new version being submitted and that no existing Outdated Reports have been created.
- [X] I have verified the new version has a Windows binary (EXE or MSI installer, Zip or other archive, etc).

### New Software Version

3.28.1

### Download location

https://github.com/Kitware/CMake/releases/download/v3.28.1/cmake-3.28.1-windows-x86_64.msi

### Package Page

https://community.chocolatey.org/packages/cmake#versionhistory

TheCakeIsNaOH:
## Description

The cmake download page was changed to have both old prerelease versions still available for download after the stable minor version is released. This causes the AU script to error, as the streams variable already has the key, so the stable version is not added. This adds a check to fix that.

## Motivation and Context

Fixes #2394
As the `.install` and `.portable` packages reference this AU script, this change fixes both of them as well.

## How Has this Been Tested?

AU script ran on Windows 10. Package tested in Chocolatey test environment. 

## Types of changes

## Checklist:

### DIFF
--- a/automatic/cmake/update.ps1
+++ b/automatic/cmake/update.ps1
@@ -31,13 +31,27 @@ function global:au_GetLatest {
     $url32_portable = $allZips | Where-Object { $_ -match "$version-$re32" }
     $url64_portable = $allZips | Where-Object { $_ -match "$version-$re64" }
 
-    $streams.Add($versionTwopart, @{
-        Version = $version
-        URL32_i = [uri]$_
-        URL64_i = [uri]$url64
-        URL32_p = [uri]$url32_portable
-        URL64_p = [uri]$url64_portable
-      })
+    if ($streams.ContainsKey($versionTwoPart)) {
+        $previousKeyVersion = Get-Version $streams[$versionTwoPart].Version
+        $currentKeyVersion = Get-Version $version
+        if ($currentKeyVersion -gt $previousKeyVersion) {
+            $streams[$versionTwopart] = @{
+                Version = $version
+                URL32_i = [uri]$_
+                URL64_i = [uri]$url64
+                URL32_p = [uri]$url32_portable
+                URL64_p = [uri]$url64_portable
+            }
+        }
+    } else {
+        $streams.Add($versionTwopart, @{
+            Version = $version
+            URL32_i = [uri]$_
+            URL64_i = [uri]$url64
+            URL32_p = [uri]$url32_portable
+            URL64_p = [uri]$url64_portable
+          })
+    }
   }
 
   return @{ Streams = $streams }


MatteoBrentegani (commented):
Do you know when this feature will be merged?

automatic/cmake/update.ps1
```diff
@@ -31,13 +31,27 @@ function global:au_GetLatest {
...
-    $streams.Add($versionTwopart, @{
-        Version = $version
-        URL32_i = [uri]$_
-        URL64_i = [uri]$url64
-        URL32_p = [uri]$url32_portable
-        URL64_p = [uri]$url64_portable
-      })
+    if ($streams.ContainsKey($versionTwoPart)) {
+        # Prefer release over prerelease if both are available for download
+        if ($streams.$versionTwoPart.Version.contains("-") -and !$version.contains("-")) {
```
AdmiringWorm, line 36: Normally, we just ignore the entry if the stream already contains the key.

If we want to have some comparison, this should be based on whether the version is higher or not.

This can be achieved by using the `Get-Version` helper that AU provides.

AdmiringWorm (changes requested):

automatic/cmake/update.ps1
```diff
@@ -31,13 +31,27 @@ function global:au_GetLatest {
...
+            $streams.Remove($versionTwopart)
+            $streams.Add($versionTwopart, @{
```
AdmiringWorm, line 38: You do not need to remove the existing version first.

Instead do

```suggestion
            $streams[$versionTwopart] = @{
```

automatic/cmake/update.ps1
TheCakeIsNaOH: Fixed up.

automatic/cmake/update.ps1
TheCakeIsNaOH: Fixed up.

automatic/cmake/update.ps1
```diff
@@ -31,13 +31,27 @@ function global:au_GetLatest {
...
 
-    $streams.Add($versionTwopart, @{
-        Version = $version
-        URL32_i = [uri]$_
-        URL64_i = [uri]$url64
-        URL32_p = [uri]$url32_portable
-        URL64_p = [uri]$url64_portable
-      })
+    if ($streams.ContainsKey($versionTwoPart)) {
+        $previousKeyVersion = Get-Version $streams[$versionTwoPart].Version
```
AdmiringWorm, line 35: Not requiring a change, as it is fine.

But to note, you can store a version acquired form `Get-Version` from the hashtable (this will work with AU). If that is done you do not need to transform the version again every time.

AdmiringWorm (approved):
LGTM, just one note for future work.

PR #2460, opened 2024-04-22
(selenium.powershell) Adds selenium.powershell

### LINKED ISSUE #2179: Create Selenium PowerShell Module package

corbob:
## Description
This new package contains the Selenium module.

It is updated from the PowerShell gallery releases, and there are not currently plans to run a prerelease stream.

Please note that there is some outstanding conversation about the correct category tag to use, and that the update script is _interesting_ in that it adds verification for each and every file within the module. This doesn't fit nicely with any of our examples, as this is the first PowerShell module to be added to this repository.

## Motivation and Context
As part of #1904, it was decided to migrate packages we depend on to the Chocolatey Community Chocolatey Packages Git repository. This package was included as wanted, but wasn't actually a pre-existing package.

Fixes #2179.

## How Has this Been Tested?
- Pack package / run Update.ps1 / Update_all.ps1 selenium.powershell
- Tested installation of package without arguments
- Tested uninstallation of package
- Tested install of package with arguments (both individually and together)
- Observed expected results
- Tested uninstall with both arguments
- Tested uninstall with the module being loaded in a PowerShell session
- Observed that module files are not removed and warning message advises what is needed

## Types of changes

## Checklist:


automatic/selenium.powershell/tools/chocolateyBeforeModify.ps1
```diff
@@ -0,0 +1,8 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$ModuleName = "Selenium"
+
+# The module shouldn't be present in the Chocolatey session, but...
+if (Get-Module $ModuleName) {
+  Remove-Module $ModuleName -Force -ErrorAction SilentlyContinue
+}
```
AdmiringWorm, line 8: The module shouldn't be removed as part of a before modify script.
Instead, extract it out to a helper file that is called from both the `chocolateyInstall` and `chocolateyUninstall` script.

automatic/selenium.powershell/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,51 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$toolsDir = Split-Path -Parent $MyInvocation.MyCommand.Definition
+
+$ModuleName = "Selenium"
+$ModuleVersion = $env:ChocolateyPackageVersion
+
+$SourcePath = Join-Path $toolsDir "$($ModuleName).zip"
+$SavedPaths = Join-Path $ToolsDir 'installedpaths'
```
AdmiringWorm, line 9: Please keep the case consistent, as it is defined with a lowercase `t`, please use that same casing everywhere.

```suggestion
$SourcePath = Join-Path $toolsDir "$($ModuleName).zip"
$SavedPaths = Join-Path $toolsDir 'installedpaths'
```

automatic/selenium.powershell/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,51 @@
+﻿$ErrorActionPreference = 'Stop'
+
+$toolsDir = Split-Path -Parent $MyInvocation.MyCommand.Definition
+
+$ModuleName = "Selenium"
```
AdmiringWorm, line 5: Using an uppercase for the first letter goes against the practices that have already been present in this repository for years.

Please update the variables used to be consistent with existing code style everywhere.

automatic/selenium.powershell/tools/chocolateyInstall.ps1
```diff
@@ -0,0 +1,51 @@
...
+if (-not $DestinationPath) {
+  $DestinationPath = Join-Path $env:ProgramFiles "WindowsPowerShell\Modules"
+}
+
+$UnzipArgs = @{
+  FileFullPath = $SourcePath
+  PackageName  = $env:ChocolateyPackageName
+}
+
+$InstalledPaths = foreach ($Path in $DestinationPath) {
```
AdmiringWorm, line 30: Any reason why you are using a `foreach` loop, instead of doing piping and the `ForEach-Object` cmdlet?

automatic/selenium.powershell/update.ps1
```diff
@@ -0,0 +1,74 @@
...
+function global:au_SearchReplace {
+  @{
+    ".\legal\VERIFICATION.txt"      = @{
+      # Replace the module version on the Save-Module line
+      "(?i)(^\s*\-\s*Save\-Module.*\-Version\s+).*" = "`${1}$($Latest.ModuleVersion)"
+
+      # Attempt to remove all existing checksums
+      "(?i)(^\s*file\s*name\:).*"         = ""
+      "(?i)(^\s*checksum\s*type\:).*"     = ""
+      "(?i)(^\s*checksum(32)?\:).*"       = ""
```
AdmiringWorm, line 15: These, or at least some of these are wrong. The AppVeyor build is failing due to not being able to replace this regex values.

They also should not be empty, please fill out this section.

automatic/selenium.powershell/update.ps1
```diff
@@ -0,0 +1,74 @@
...
+      "(?i)(^\s*3. The checksums should match the following:)" = "`${1}`n`n$(
+        @($Latest.ModuleChecksums.ForEach{
+          @(
+            "  file name: $($_.Path -replace "^(?<BasePath>.+)\\Selenium\\$($Latest.ModuleVersion)\\")"
+            "  checksum type: $($_.Algorithm)"
+            "  checksum: $($_.Hash)"
+            "`n"
+          ) -join "`n"
+        }) -join ''
+      )"
```
AdmiringWorm, line 30: This would be better to be handled in a `global:au_AfterUpdate` instead of inline here.

The `au_SearchReplace` cmdlet should be as simple as possible, not include nested logic like this.

automatic/selenium.powershell/README.md
```diff
@@ -0,0 +1,20 @@
+# Selenium PowerShell Module
+
+The Selenium PowerShell module allows you to automate browser interaction using the [Selenium API](https://selenium.dev/). You can navigate to pages, find elements, click buttons, enter text and even take screenshots. 
+
+- Wraps the C# WebDriver for Selenium
+- Easily execute web-based tests
+- Works well with Pester
```
AdmiringWorm, line 7: These sounds like they should be in a `Features` section, see the Contribution guidelines: https://github.com/chocolatey-community/chocolatey-packages/blob/master/CONTRIBUTING.md#description

AdmiringWorm (changes requested):
A few things to be fixed up, please also keep things consistent with existing code style in the repository.

automatic/selenium.powershell/update.ps1
JPRuskin: To be clear, as in the comment "Attempt to remove all existing checksums", this is removing the lines. The emptiness is a feature.

corbob: The failure was definitely because of the lack of existing checksums. I added one of them back for the PR, and it is working locally for me (I'm not entirely sure why it worked before for me... It's possible I might not have had a fully clean working directory, or was working in pwsh)

automatic/selenium.powershell/tools/chocolateyBeforeModify.ps1
corbob: I've just taken this out completely as there shouldn't be a scenario where it's ever loaded in the Chocolatey PowerShell session, nor will removing it actually acomplish anything for this scenario as it would still have the DLLs locked.

corbob:
I've updated the code based on the review comments @AdmiringWorm.

automatic/selenium.powershell/tools/chocolateyInstall.ps1
corbob: Mostly because I picked it up from @JPRuskin's previous work. I've updated it to using `ForEach-Object` to be more consistent with how PowerShell would normally write the assignment.

automatic/selenium.powershell/legal/VERIFICATION.txt
```diff
@@ -0,0 +1,17 @@
...
+1. Save the module from the PowerShell Gallery:
+  - Save-Module -Name Selenium -Version 3.0.1
+2. Get the checksum for each file using one of the following methods:
+  - Using powershell function 'Get-FileHash'
+  - Use chocolatey utility 'checksum.exe'
+3. The checksums should match the following:
+
+  file name: LICENSE
+  checksum type: SHA256
+  checksum: BAEB4A5B8A2F782E21C9F508EEBC728ADBA739F7B7FB8BA3FF07747A4C9FE296
```
AdmiringWorm, line 17: This indicates that the checksum is for the License file, is that correct?
We typically do not include that checksum for the license.

The location of the embedded license is usually included differently as well, typically at the bottom of the file with something like the following
```
File 'LICENSE.txt' is obtained from <https://github.com/git-for-windows/git/blob/703601d6780c32d33dadf19b2b367f2f79e1e34c/COPYING>
```

automatic/selenium.powershell/update.ps1
AdmiringWorm: Then it sounds like a different approach is needed, something to consider at least.

Maybe keep a seperate file as the template of this, then during update that file is read in, any additional information is appended to the file and then it writes/overwrites the VERIFICATION.txt file, that should be done in the AfterUpdate method though in that case, and then remove this regexes.

AdmiringWorm (changes requested):

automatic/selenium.powershell/legal/VERIFICATION.txt
JPRuskin: I think in this case it's because that version of the license file is included in the output saved artifact you get from step 1, which is unusual-ish. It is one of the files that could be checked (like the README.md).

image

It seems valid to include that - otherwise we're removing stuff from the artifact you get from the publisher, or not mentioning it in the verification, which might raise questions if noticed.

automatic/selenium.powershell/legal/VERIFICATION.txt
```diff
@@ -0,0 +1,17 @@
+VERIFICATION
+Verification is intended to assist the Chocolatey moderators and community
+in verifying that this package's contents are trustworthy.
+
+The module has been downloaded from <https://www.powershellgallery.com/packages/Selenium>
+and can be verified by following these steps:
+
+1. Save the module from the PowerShell Gallery:
+  - Save-Module -Name Selenium -Version 3.0.1
```
JPRuskin, line 9:
```suggestion
  - Save-Module -Name Selenium -MaximumVersion 3.0.1
```

Just noticed that -Version isn't actually a valid argument in the PS5/7 sessions I have open.

automatic/selenium.powershell/legal/VERIFICATION.txt
JPRuskin: Technically I guess we're shipping that file twice in the package. I don't think that's problematic?

AdmiringWorm: In which case, that license file should probably be compared against the license we embed in the legal folder instead then.

automatic/selenium.powershell/legal/VERIFICATION.txt
corbob: Good catch. It's also missing the required `-Path`. I've updated both of these.

automatic/selenium.powershell/legal/VERIFICATION.txt
corbob: > I thought it was to the license file we had there, but if it isn't then the verification file is missing the location where that license was downloaded from and needs to be added.

I believe this is step 1 in the verification file: it's downloaded with the `Save-Module` command. Would it be sufficient to remove the `LICENSE.txt` from the `legal` directory and just use the one from the `Save-Module` command?

automatic/selenium.powershell/update.ps1
corbob: I've moved this logic into the AfterUpdate.

automatic/selenium.powershell/legal/VERIFICATION.txt
JPRuskin: I think we still need that license file for CCR moderation purposes, unless we mess with the zip file before pushing the package, but that's just a surface level thought - I'm still uneasy messing with the artifact.

AdmiringWorm (changes requested):

automatic/selenium.powershell/legal/LICENSE.txt
```diff
@@ -0,0 +1,23 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+File 'LICENSE.txt' is obtained from <https://github.com/adamdriscoll/selenium-powershell/blob/478369b92437f7bfba14ddc64109ee07ee2ecca2/LICENSE>
```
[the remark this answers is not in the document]
corbob: I misread your previous comment on the VERIFICATION.txt file and interpretted it to mean adding that to the end of the LICENSE.txt file. I've corrected this and added it to the VERIFICATION.txt file.

automatic/selenium.powershell/legal/VERIFICATION.txt
```diff
@@ -0,0 +1,118 @@
+VERIFICATION
+Verification is intended to assist the Chocolatey moderators and community
+in verifying that this package's contents are trustworthy.
+
+The module has been downloaded from <https://www.powershellgallery.com/packages/Selenium>
```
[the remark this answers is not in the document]
corbob: I've clarified this in the file.

PR #2479, opened 2024-05-28
`(mumble)` Outdated, fix getting auto updates and add a required dependency

suprovsky:
## Description
This should fix checking Mumble version available on GitHub, resulting in properly working automatic updates, as well as a required dependency, which is not installed on Windows Server by default.

## Motivation and Context
The version available in Chocolatey is 1.4.287 where the latest stable is 1.5.634. Automatic Chocolatey updates does not work because `mumble.info` does not work as it's been moved to `www.mumble.info` it seems. Updated script uses fetching version from GitHub which is more reliable.
I've added `vcredist140` as a dependency because without it Mumble won't work on Windows Server (which was a testing environment system).

## How Has this Been Tested?
I've run `update.ps1` on my local machine (Windows 11 amd64) to see if `Version` and `URL32` is set properly. Also used a Chocolatey Test Environment to see if the package installs.

## Types of changes

## Checklist:

- [X] The added/modified package passed install/uninstall in the Chocolatey Test Environment(https://github.com/chocolatey-community/chocolatey-test-environment/). _Note that we don't support the use of any other environment_.


daerSeebaer (commented):
Looks good to me, much cleaner than before.

automatic/mumble/update.ps1
```diff
@@ -19,27 +19,21 @@ function global:au_SearchReplace {
...
+  $url32 = $LatestRelease.assets | Where-Object {$_.name.StartsWith("mumble_client")} | Where-Object {$_.name.EndsWith(".msi")} | Select-Object -ExpandProperty browser_download_url
+  $releaseNotes = $download_page.Links | ? href -match "www.mumble.info\/blog" | Select-Object -first 1 -expand href
   if (!$releaseNotes) {
     $releaseNotes = "$releases/tag/$version32"
   }
 
   @{
     URL32        = $url32
-    Version      = Get-Version $version32
+    Version      = $LatestRelease.tag_name.TrimStart("v.")  # Tags have a "v." prefix, not a typo
```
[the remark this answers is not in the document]
suprovsky: I've removed the redundant dot.

automatic/mumble/update.ps1
```diff
@@ -19,27 +19,21 @@ function global:au_SearchReplace {
...
-
-  $verRe = '\/'
-  $version32 = $url32 -split "$verRe" | select -last 1 -skip 1
-
-  $releaseNotes = $download_page.Links | ? href -match "mumble.info\/blog" | select -first 1 -expand href
-
+  $url32 = $LatestRelease.assets | Where-Object {$_.name.StartsWith("mumble_client")} | Where-Object {$_.name.EndsWith(".msi")} | Select-Object -ExpandProperty browser_download_url
+  $releaseNotes = $download_page.Links | ? href -match "www.mumble.info\/blog" | Select-Object -first 1 -expand href
   if (!$releaseNotes) {
     $releaseNotes = "$releases/tag/$version32"
```
TheCakeIsNaOH, line 31: It looks like the `$version32` variable definition was removed, but it is still used here.

TheCakeIsNaOH (changes requested):

automatic/mumble/update.ps1
suprovsky: alright, i'll remove it

TheCakeIsNaOH: In the case that the release notes is null/empty, it looks like there is not handling for it anymore. Instead of removing that if statement entirely, the code should be updated to still link to that page; so the `$version32` would still be needed unless you used a different way to get the fallback release notes url.

suprovsky: Ok, I've used `$LatestRelease.body` as a fallback option.

TheCakeIsNaOH (approved):
LGTM

PR #2502, opened 2024-07-04
(kubernetes-cli) Fix Update.ps1 (Prerelease version broke all)

JPRuskin:
## Description

Changing to `continue` rather than `return` when a URL isn't matched fixes an issue where a pre-release version broke our regex.

Also takes advantage of the REST API to get the changelogs, which improves performance from ~164s to ~86s on my machine.

We may consider releasing prerelease streams at some point, but we don't currently do so - otherwise, we could fix up the regex to respect full semver versions.

## Motivation and Context

GetLatest was failing to return anything. On inspection, it seems that a latest -alpha release was causing get latest to return immediately instead of listing the rest of the available streams.

## How Has this Been Tested?

- Before changes:
    image
- With just the fix to `continue`:
    image
- Including the change to using the REST API:
    image

## Types of changes

## Checklist:
- [ ] The added/modified package passed install/uninstall in the Chocolatey Test Environment(https://github.com/chocolatey-community/chocolatey-test-environment/). _Note that we don't support the use of any other environment_.

### DIFF
--- a/automatic/kubernetes-cli/update.ps1
+++ b/automatic/kubernetes-cli/update.ps1
@@ -3,7 +3,10 @@ param($IncludeStream, [switch] $Force)
 
 Import-Module Chocolatey-AU
 
-$changelogs = 'https://raw.githubusercontent.com/kubernetes/kubernetes/master/CHANGELOG/README.md'
+$changelogRepository = @{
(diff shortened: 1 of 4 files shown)


AdmiringWorm (approved):
LGTM

PR #2647, opened 2025-03-21
(calibre) switch AU script to use Github for version

### LINKED ISSUE #2646: (calibre) Outdated
### Checklist

- [x] I have checked the moderation queue that no new version has been submitted (only visible when logged in), or the submitted package version has failed the automated checks.

Please navigate to the following [link](https://community.chocolatey.org/packages?q=&moderatorQueue=true&moderationStatus=all-statuses&prerelease=false&sortOrder=package-download-count) to view the moderation queue.

- [x] I have verified that the new version is a stable release.
- [x] I have looked to see if there are any Bug reports that prevents a new version being submitted and that no existing Outdated Reports have been created.
- [x] I have verified the new version has a Windows binary (EXE or MSI installer, Zip or other archive, etc).

### New Software Version

8.0.1

### Download location

https://download.calibre-ebook.com/8.0.1/calibre-64bit-8.0.1.msi

### Package Page

https://community.chocolatey.org/packages/calibre

TheCakeIsNaOH:
## Description
 
Updates AU script to scrape Github releases page to get the latest version number.

## Motivation and Context

Update script outdated. 
Does not use the Github download URL, as Github binaries removed from older releases.

Fixes #2646 

## How Has this Been Tested?

Win11 and test env

## Types of changes

## Checklist:

- [x] The added/modified package passed install/uninstall in the [Chocolatey Test Environment](https://github.com/chocolatey-community/chocolatey-test-environment/). _Note that we don't support the use of any other environment_.

### DIFF
--- a/automatic/calibre/update.ps1
+++ b/automatic/calibre/update.ps1
@@ -20,7 +20,7 @@ function global:au_SearchReplace {
 
 function global:au_GetLatest {
     $githubLatestRelease = "https://github.com/kovidgoyal/calibre/releases/latest"
-    $download_page = Invoke-WebRequest -Uri $githubLatestRelease
+    $download_page = Invoke-WebRequest -Uri $githubLatestRelease -UseBasicParsing
 
     $versionHyperlink = $download_page.BaseResponse.ResponseUri
 
--- a/automatic/calibre/update.ps1
+++ b/automatic/calibre/update.ps1
@@ -1,8 +1,6 @@
 ﻿Import-Module Chocolatey-AU
 Import-Module "$PSScriptRoot\..\..\scripts\au_extensions.psm1"
 
-$releases = 'https://download.calibre-ebook.com/7.html'
-
 function global:au_BeforeUpdate {
   $Latest.Checksum64 = Get-RemoteChecksum $Latest.URL64
 }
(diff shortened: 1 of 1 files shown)


pauby:
The [download page](https://calibre-ebook.com/download_windows) has the latest version number displayed. Could this not be parsed to retrieve that and the other URL's determined from it?

The latest version can be download from https://calibre-ebook.com/download_windows. I appreciate that is not one that a specific version can be downloaded from.

Do we just update the code every version, or have we looked at refactoring this?

If it can be refactored it would save it being updated for every major version, defeating the goal of having full automation, and save us having to document the update requirement, assuming that it hasn't been already.

TheCakeIsNaOH:
>  Could this not be parsed to retrieve that and the other URL's determined from it?

Yes, however it looks like a minor change in the page would break that parsing, which is why this was not used last time I look at it. I'd personally much rather have to update the script per major version than have to fix the parsing or deal with a incorrect version getting pushed if the page changes.

> Just a drive-by comment (and I really appreciate all your efforts to keep this up to date!): the maintainer also releases on GitHub (https://github.com/kovidgoyal/calibre/releases/), that's the "Alternate download location" from https://calibre-ebook.com/download_windows.

My understanding is that those binaries are removed once a new version comes out. However, it looks like releases are reliably getting published now, so perhaps that could be used for getting the latest version.

TheCakeIsNaOH:
I've pushed an update.

RogueScholar:
It seems like the ideal source would actually be <https://calibre-ebook.com/signatures/>, where we'd not only be able to monitor for new files according to a seemingly well-established rubric, but also grab the signature files for use in verifying the fidelity of the downloaded files. Debian's `uscan` seems to manage this type of need robustly; maybe we should look into porting the relevant bits from Perl to PowerShell, or does Chocolatey already have a similar capability?

flyingstar16:
> My understanding is that those binaries are removed once a new version comes out. However, it looks like releases are reliably getting published now, so perhaps that could be used for getting the latest version.

Ah awesome, thanks! I was going to suggest to use the JSON endpoint instead of doing URL parsing (https://api.github.com/repos/kovidgoyal/calibre/releases?per_page=1 - got it from [here](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28)) but it still requires parsing either `tag_name` or `name` (and those are not particularly stable :( )

Unless @RogueScholar's suggestion is better :D

automatic/calibre/update.ps1
```diff
@@ -21,12 +19,12 @@ function global:au_SearchReplace {
 }
 
 function global:au_GetLatest {
-    $download_page = Invoke-WebRequest -Uri $releases
+    $githubLatestRelease = "https://github.com/kovidgoyal/calibre/releases/latest"
+    $download_page = Invoke-WebRequest -Uri $githubLatestRelease
 
-    $versionHyperlink = $download_page.links | Select-Object -First 1
-    if ($versionHyperlink.Title -notmatch 'Release (7[\d\.]+)' ) { throw "Calibre version 7.x not found on $releases" }
+    $versionHyperlink = $download_page.BaseResponse.ResponseUri
```
Jaykul, line 25: Does chocolatey always run these in _Windows PowerShell_ 5.x?

PR #2757, opened 2025-12-08
(sysinternals) Remove file checksum checks to avoid breakage

### LINKED ISSUE #2751: (sysinternals) sysinternals v2025.11.17 checksum mismatch
### Checklist

- [x] I have verified that this is happening in the latest available version of the package.

### Chocolatey Version

2.5.1

### Chocolatey License

None

### Package Version

v2025.11.17

### Current Behaviour

Checksum mismatch, will not install:

### Expected Behaviour

_No response_

### Steps To Reproduce

Try to install/upgrade to latest sysinternals package via chocolatey.

### Environment

```markdown
- Operating System:
- PowerShell Version:
- Shell:
```

### Chocolatey Log

```shell
Relevant bit:
`ERROR: Checksum for 'C:\Users\username\AppData\Local\Temp\chocolatey\sysinternals\2025.11.17\SysinternalsSuite.zip' did not meet 'f544fd336bfce496870060c8bc56fd7f084e8f696e51676497fa0330310d0a0a' for checksum type 'sha256'. Consider passing the actual checksums through with --checksum --checksum64 once you validate the checksums are appropriate. A less secure option is to pass --ignore-checksums if necessary.
The upgrade of sysinternals was NOT successful.`
```

### Anything else?

_No response_

forderud:
Fixes #2751. The SHA256 file checks here are problematic, since they're based on URLs that point to the latest version of `SysinternalsSuite` & `SysinternalsSuite-Nano`. This means that the script will break every time a new `SysinternalsSuite` version is released.

Proposing to remove the SHA256 file checks to avoid problems like #2751 moving forward. A better solution would be to instead download `SysinternalsSuite` from version-specific URLs. However, I've been unable to find any such URLs on https://learn.microsoft.com/nb-no/sysinternals/downloads/

### DIFF
--- a/automatic/sysinternals/tools/chocolateyInstall.ps1
+++ b/automatic/sysinternals/tools/chocolateyInstall.ps1
@@ -9,8 +9,6 @@ Write-Host "Sysinternals Suite is going to be installed in '$installDir'"
 $packageArgs = @{
   packageName    = 'sysinternals'
   url            = 'https://download.sysinternals.com/files/SysinternalsSuite.zip'
-  checksum       = 'ba441238eb68b33aa7b9abf141ee9bc5393dfce2e174135095b898840b0b6c58'
-  checksumType   = 'sha256'
   unzipLocation  = $installDir
 }
 Install-ChocolateyZipPackage @packageArgs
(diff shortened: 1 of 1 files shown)


AdmiringWorm:
As already mentioned, we cannot accept removal of the checksum, and the package can not be embedded due to licensing issues.

While there may be work that can be possible to do if versioned URLs are created in the future, we are not at that stage right now.

As such, I will be closing this PR now, if you want to keep the conversation going somewhere, please open up a Discussion instead.

PR #2803, opened 2026-06-29
(googlechrome) derive version from downloaded MSI artifact

mu88:
## Description
Update `automatic/googlechrome` AU flow to download the MSI first, extract `ProductVersion` from MSI metadata, and compute checksums from the downloaded artifacts.

## Motivation and Context
The Chrome enterprise MSI URL is rolling and can drift from API-reported version data. Verifying version from the actual downloaded MSI avoids publishing metadata based on a different artifact than the one hashed.

## How Has this Been Tested?
Local AU test runs were executed in the fork.
A fail-fast mismatch path was observed with live data, and a controlled aligned run verified that nuspec and chocolateyInstall values are updated consistently.

N/A

## Types of changes

## Checklist:
- [ ] The added/modified package passed install/uninstall in the [Chocolatey Test Environment](https://github.com/chocolatey-community/chocolatey-test-environment/). _Note that we don't support the use of any other environment_.

### DIFF
--- a/automatic/googlechrome/update.ps1
+++ b/automatic/googlechrome/update.ps1
@@ -4,9 +4,106 @@ import-module "$PSScriptRoot\..\..\scripts\au_extensions.psm1"
 $releases = "https://versionhistory.googleapis.com/v1/chrome/platforms/win/channels/stable/versions"
 $paddedUnderVersion = '57.0.2988'
 
+function Get-MsiProductVersion {
+  param(
+    [Parameter(Mandatory = $true)]
+    [string]$MsiPath
+  )
+
+  $installer = $null
+  $database = $null
+  $view = $null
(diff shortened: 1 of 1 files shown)


mu88:
Good catch on the failed AppVeyor run — this is expected with the current fail-fast check and actually reproduces the user-facing drift problem we are trying to guard against.

The run failed because VersionHistory reported `150.0.7871.47` while the rolling MSI URL still served `149.0.7827.201`. That mismatch is exactly the scenario that causes end-user checksum failures.

In other words: the CI failure here is signal, not noise.

If maintainers prefer fewer false-red CI runs, a follow-up can switch the version source to a fully rolled release filter (for example `fraction=1`) or downgrade mismatch handling from hard-fail to warning.

PR #2811, opened 2026-07-31
jubler: update to 10.0.0

teras:
Updates the `jubler` package from 7.0.0/8.0.0 to the current release, **10.0.0**.

The automatic updater had been stuck because `au_GetLatest` scraped `http://www.jubler.org/download.html`, which is no longer maintained. This PR repoints it at the project's GitHub releases page so future versions are picked up automatically.

Changes:
- `update.ps1`: fetch the latest installer from `https://github.com/teras/Jubler/releases/latest` (matching `Jubler-*-x64.exe`); parse the version from the asset name; `-ChecksumFor 64`.
- `chocolateyinstall.ps1`: switch from embedded binaries to download-at-install (`Install-ChocolateyPackage`, `url64bit` + `checksum64`). Jubler 10 ships a single self-contained x64 installer.
- Silent install/uninstall arguments updated for the current Inno Setup based installer (`/VERYSILENT /SUPPRESSMSGBOXES /NORESTART`).
- Dropped the obsolete `javaruntime`, `smplayer` and `chocolatey-core.extension` dependencies — the installer now bundles its own Java runtime and uses VLC internally.
- Refreshed metadata: `projectUrl`, `licenseUrl`, copyright year, release-notes links, and description/readme (FFMPEG/mplayer references replaced by VLC).
- `VERIFICATION.txt` and `Changelog.md` updated accordingly.

### DIFF
--- a/automatic/jubler/jubler.nuspec
+++ b/automatic/jubler/jubler.nuspec
@@ -29,7 +29,7 @@ Jubler is a tool to edit text-based subtitles. It can be used as an authoring so
 * Translating mode (parent & child editors) is supported
 * Graphical preview of subtitles using VLC. Current frame, waveform preview and waveform listening is supported.
 * Graphical display of subtitles, which can be moved and resized.
-* Test and play the subtitles file using an integrated video player (VLC). While in playing mode the user is able to freely edit the subtitles (and inform the player of this change), add a new subtitle in real time or synchonize subtitles with the movie.
+* Test and play the subtitles file using an integrated video player (VLC). While in playing mode the user is able to freely edit the subtitles (and inform the player of this change), add a new subtitle in real time or synchronize subtitles with the movie.
 * Mark subtitles with different colors, either when editing or real time when playing the video.
 * Spell checking, with support for dictionary selection.
 * Easy installation for Mac, Linux & Windows platforms and a generic installer for all other platforms.
--- a/automatic/jubler/jubler.nuspec
+++ b/automatic/jubler/jubler.nuspec
@@ -3,15 +3,15 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>jubler</id>
-    <version>8.0.0</version>
+    <version>10.0.0</version>
     <packageSourceUrl>https://github.com/chocolatey-community/chocolatey-packages/tree/master/automatic/jubler</packageSourceUrl>
     <owners>chocolatey-community, AdmiringWorm</owners>
     <title>Jubler Subtitle Editor</title>
     <authors>Panayotis Katsaloulis</authors>
-    <projectUrl>http://www.jubler.org/index.html</projectUrl>
+    <projectUrl>https://github.com/teras/Jubler</projectUrl>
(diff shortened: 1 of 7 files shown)


flcdrg (dismissed):

flcdrg:
Build is failing with 'invalid version'

teras:
True, sorry about that. I updated it using GitHup API instead. I hope it works now as expected.

flcdrg (dismissed):

automatic/jubler/jubler.nuspec
```diff
@@ -3,15 +3,15 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
   <metadata>
     <id>jubler</id>
-    <version>8.0.0</version>
+    <version>10.0.0</version>
```
AdmiringWorm: Updating the version number manually will prevent the automatic updater from updating the package, causing only versions relased after `10.0.0` to be pushed.

automatic/jubler/jubler.nuspec
```diff
@@ -48,14 +48,9 @@ Jubler is a tool to edit text-based subtitles. It can be used as an authoring so
 
 ]]></description>
     <releaseNotes>
-[Software Changelog](http://www.jubler.org/changelog.html)
-[Package Changelog](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/jubler/Changelog.md)
+[Software Changelog](https://github.com/teras/Jubler/releases)
```
AdmiringWorm, line 51: The changelog is still being published on the jubler.org site.

As such, we should keep the release note URLs for the software to that page (only changing the protocol used).

AdmiringWorm (changes requested):
Just a couple of items, otherwise this looks good to me.

automatic/jubler/jubler.nuspec
teras: > Updating the version number manually will prevent the automatic updater from updating the package, causing only versions relased after `10.0.0` to be pushed.

I am sorry, what is supposed to be the correct value? 
8.0.0 is a really old version, should it stay like this? I am confused.

automatic/jubler/jubler.nuspec
teras: > The changelog is still being published on the jubler.org site.
> 
> As such, we should keep the release note URLs for the software to that page (only changing the protocol used).

Should I keep the original changelog in my site? Indeed this is the full story. I thought you needed the latest changes, in a more structured way.

automatic/jubler/jubler.nuspec
flcdrg: The update.ps1 script should update the version number when it runs during the scheduled build on the build server. If you change the version here, then the script will think there is nothing to do and not create the new 10.0.0 package or publish it.

automatic/jubler/jubler.nuspec
flcdrg: Given you're the software publisher and that is the source for the application, I think you are correct and the GitHub releases is a better URL

automatic/jubler/jubler.nuspec
teras: so the trick is to keep an older version, i.e. the 8.0.0 one so the script should know that it needs to run it again? OK I'll revert this line

teras: I updated the PR

Issue #2771, opened 2026-02-04
GoogleChrome package v145 installs v144.


heylanwe64:
### Checklist

- [x] I have verified that this is the correct repository, and the package is maintained by the chocolatey-community user.
- [x] I have verified that this is happening in the latest available version of the package.

### Chocolatey Version

2.5.1

### Chocolatey License

None

### Package Version

145.0.7632.26

### Current Behaviour

When installing latest package version Google Chrome which is 145.0.7632.26 currently  the Event Viewer claims it is actually 144.0.7559.110. This is causing all sorts of version mismatch (ex. ChromeDriver).

### Steps To Reproduce

1. Run choco upgrade googlechrome chromedriver -y -f
2. Check version of google chrome in windows

googlechrome version did not upgrade. It breaks CI flows.

### Environment

```markdown
- Operating System: W11
- PowerShell Version: 5.1.26100.7627
- Shell: choco upgrade googlechrome chromedriver -y -f
```

### Chocolatey Log

```shell
Full log in the attachment.
```

### Anything else?

[chocolatey.log](https://github.com/user-attachments/files/25066435/chocolatey.log)

AdmiringWorm:
I am taking this based on my memory, this is nothing new and is something I have seen happen from time to time with several Google Chrome releases (upstream).

Basically what it looks like for me is that they sometimes (or quite often, not sure) aren't updating the version information for the main executable. What still is happening though, is that the first time google chrome is launched, something on their own backend changes the version numbers (but without installing any new version).

PR #2817, opened 2026-08-21
Check that the downloaded MSI version matches the one from the URL

### LINKED ISSUE #2816: (GoogleChrome) 152 early stable release packaged
### Checklist

- [x] I have verified that this is happening in the latest available version of the package.

### Chocolatey Version

2.7.4

### Chocolatey License

Business

### Package Version

152.0.7977.42

### Current Behaviour

152 appears to be an "early stable" release and not available to everyone - https://chromereleases.googleblog.com/2026/08/early-stable-update-for-desktop_02031437787.html

So when we internalized this package we ended up with a 151.X version

### Expected Behaviour

Version is a fully released version so it matches the download

### Steps To Reproduce

Internalize the GoogleChrome package

### Environment

```markdown
- Operating System:
- PowerShell Version:
- Shell:
```

### Chocolatey Log

```shell
N/A
```

### Anything else?

_No response_

opoplawski:
Authored-By: Codex

## Description
Check that the version of the downloaded MSI matches the version in the URL.

## Motivation and Context
Often the package is released with mis-matched versions.  Should fix #2816 #2771 and possibly others

## How Has this Been Tested?
Not at all.  Not sure how to test.

## Types of changes

## Checklist:
- [ ] The added/modified package passed install/uninstall in the [Chocolatey Test Environment](https://github.com/chocolatey-community/chocolatey-test-environment/). _Note that we don't support the use of any other environment_.

### DIFF
--- a/automatic/googlechrome/update.ps1
+++ b/automatic/googlechrome/update.ps1
@@ -2,7 +2,16 @@
 import-module "$PSScriptRoot\..\..\scripts\au_extensions.psm1"
 
 $releases = "https://versionhistory.googleapis.com/v1/chrome/platforms/win/channels/stable/versions"
-$paddedUnderVersion = '57.0.2988'
+
+function Get-MsiVersion([string]$path) {
+  $installer = New-Object -ComObject WindowsInstaller.Installer
+  $database = $installer.OpenDatabase($path, 0)
+  $view = $database.OpenView("SELECT `Value` FROM Property WHERE Property = 'ProductVersion'")
+  $view.Execute()
+  $record = $view.Fetch()
+  if (!$record) { throw "Unable to read ProductVersion from $path" }
+  $record.StringData(1)
+}
 
 function global:au_BeforeUpdate {
   $Latest.Checksum32 = Get-RemoteChecksum $Latest.URL32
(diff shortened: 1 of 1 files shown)


mastadook:
MSI version validation: Package=GoogleChrome PackageVersion=152.0.7977.54 MSI=tools/files/googlechromestandaloneenterprise64.msi ProductVersion=151.0.7922.170 CommentsVersion=151.0.7922.170 MatchedVersion= Match=False

opoplawski:
Fixed some issues with the original request.  It at least seems to run now on my machine.  But as of now it appears that both the api and the msi resolve to 152.0.7977.65 so it's hard to say if it would actually catch a discrepancy

opoplawski:
Well, the basic check works:
```
PS C:\chocolatey-packages> C:\chocolatey-packages\update_all.ps1 -Name GoogleChrome
Updating 1 automatic packages at 2026-08-26 12:03:04
Push is disabled
NoCheckChocoVersion is disabled
   [1/1] googlechrome ERROR:
     au_GetLatest failed
     Downloaded Google Chrome MSI version  152.0.7977.65 does not match API version 153.0.8010.12 (15.67s)
```
But this will simply prevent the package from getting updated except for on those few times that the two versions match.  Maybe we should just get the version number from the MSI and disregard the API list?

corbob:
I could be misremembering things, but I seem to recall, once upon a time the MSI version for Google Chrome was NEVER the same as the version that actually installed. Is that no longer the case? If the MSI version can be relied upon, then we could probably just use that version instead of the API version.

opoplawski:
I have no historical knowledge here.  But at least for the last few versions that I've looked at it seems to match.  There may be a leading space in the version returned by the Get-MSIVersion function.  It also seems to need to be cast to a string for the package Version argument to accept it.  There may be better ways to get the version from the MSI, this is just what codex came up with.

